PR #2, opened 2015-05-12
(maint) Simplify test configuration

MikaelSmith:
Remove unused `development` Gems, and unnecessary variables for `rake
test` setup in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 ---
 language: ruby
-bundler_args: --without development system_tests
+bundler_args: --without system_tests
 before_install: rm Gemfile.lock || true
 rvm:
   - 1.8.7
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,4 +8,4 @@ rvm:
   - 2.0.0
 script: bundle exec rake test
 env:
-  - PUPPET_VERSION="3.8.0" STRICT_VARIABLES=yes FUTURE_PARSER=yes STRINGIFY_FACTS=false
```
branan: why the death of `STRINGIFY_FACTS=false`? Aren't we depending on that being the case before upgrade?

MikaelSmith: It doesn't affect spec tests, because we stub all the facts. I'm also not sure the `STRINGIFY_FACTS` variable actually did anything.

PR #3, opened 2015-05-12
(PUP-4557) Extend unit testing for 100% coverage

MikaelSmith:


spec/classes/agent_upgrade_prepare_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    it { is_expected.to contain_file('/etc/puppetlabs/puppet/ssl').with({
+      'ensure'  => 'directory',
+      'source'  => '/dev/null/ssl',
+      'backup'  => 'false',
+      'recurse' => 'true',
+    }) }
+
+    ['agent', 'main', 'master'].each do |section|
+      ['allow_variables_with_dashes',
```
branan: oh god

MikaelSmith: yeah... maybe not worth it.

spec/classes/agent_upgrade_prepare_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      :osfamily => 'RedHat',
+      :puppet_ssldir   => '/dev/null/ssl',
+      :puppet_config   => '/dev/null/puppet.conf',
+      :mcollective_configured => false,
+    }}
+
+    [true, false].each do |mcollective_configured|
+      context "with mcollective_configured = #{mcollective_configured}" do
+        let(:facts) {{
+          :osfamily => 'RedHat',
```
branan: These facts are set in the parent context. You shouldn't have to re-set them in this one?

Unless rspec-puppet breaks normal rspec semantics :)

spec/classes/agent_upgrade_prepare_spec.rb
branan: I'm not sure it is, TBH

It will be worth keeping around checks for augeas changes that rely on actual logic, but this doesn't seem useful to me.

spec/classes/agent_upgrade_prepare_spec.rb
MikaelSmith: I'm overriding the `:mcollective_configured` fact; I didn't see a great way to do that while inheriting the rest of `facts`.

spec/classes/agent_upgrade_prepare_spec.rb
MikaelSmith: The main use is to verify that we maintain full code coverage for other changes.

MikaelSmith: There might be a way to rework this so we don't duplicate the list of settings that are changed in the spec.

spec/classes/agent_upgrade_prepare_spec.rb
branan: `let(:facts) { facts.merge({...})}`?

Not sure that's much clearer, though

spec/classes/agent_upgrade_prepare_spec.rb
MikaelSmith: I don't see a way to write this without enumerating the settings.

branan:
This conflicts with the other spec pull. I probably should've merged this one first, since it's bigger. Oh well.

This is ready as soon as it gets a rebase, I think

MikaelSmith:
Rebased.

PR #4, opened 2015-05-12
(PUP-4567) Update acceptance infrastructure

MikaelSmith:
Update acceptance tests and nodesets to run basic testing.

Added running multiple test scenarios. Tests can be run with SHA=b94d3b4f2db8de95f151b98a2c6054daddbb0d4d bundle exec rspec spec/acceptance, which downloads a vagrant VM (uses default.yml) and runs it. Specifying a SHA should allow tests to be run in Jenkins as well.

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -3,22 +3,97 @@ require 'spec_helper_acceptance'
 describe 'agent_upgrade class' do
 
   context 'default parameters' do
-    # Using puppet_apply as a helper
+    before(:all) { setup_puppet }
+    after (:all) { teardown_puppet }
+
     it 'should work idempotently with no errors' do
       pp = <<-EOS
       class { 'agent_upgrade': }
       EOS
 
-      # Run it twice and test for idempotency
+      # TODO: Run it twice and test for idempotency; requires ability to change
+      #       Beaker config to AIO mid-run.
       apply_manifest(pp, :catch_failures => true)
-      apply_manifest(pp, :catch_changes  => true)
+      #apply_manifest(pp, :catch_changes  => true)
     end
 
-    describe package('agent_upgrade') do
+    describe package('puppet-agent') do
       it { is_expected.to be_installed }
     end
 
-    describe service('agent_upgrade') do
+    describe service('puppet') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+
+    describe service('mcollective') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  end
+
+  context 'no services enabled on install' do
+    before(:all) { setup_puppet }
+    after (:all) { teardown_puppet }
+
+    it 'should work idempotently with no errors' do
+      pp = <<-EOS
+      class { 'agent_upgrade': service_names => [] }
+      EOS
+
+      # TODO: Run it twice and test for idempotency; requires ability to change
+      #       Beaker config to AIO mid-run.
+      apply_manifest(pp, :catch_failures => true)
+      #apply_manifest(pp, :catch_changes  => true)
+    end
+
+    describe package('puppet-agent') do
+      it { is_expected.to be_installed }
+    end
+
+    describe service('puppet') do
+      it { is_expected.to_not be_enabled }
+      it { is_expected.to_not be_running }
+    end
+
+    describe service('mcollective') do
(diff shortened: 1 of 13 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -4,7 +4,19 @@
...
-    if host.is_pe?
+    if ENV['SHA']
+      install_puppetlabs_release_repo(host)
+      install_puppetlabs_dev_repo(host, 'puppet', ENV['SHA'])
+
+      #install_packages_from_local_dev_repo(host, 'puppet')
+      if host['platform'] =~ /debian|ubuntu|cumulus/
+        on host, 'apt-get install -y puppet'
+      elsif host['platform'] =~ /fedora|el|centos/
+        on host, 'yum install -y puppet'
```
branan: At some point do we want to install/configure MCO so that we can verify migration is correct there?

MikaelSmith: Yes. I'm still working out how to have multiple tests, because I need to create a new VM for each one.

MikaelSmith:
Added running multiple test scenarios. Tests can be run with `SHA=b94d3b4f2db8de95f151b98a2c6054daddbb0d4d bundle exec rspec spec/acceptance`, which downloads a vagrant VM (uses default.yml) and runs it. Specifying a SHA should allow tests to be run in Jenkins as well.

MikaelSmith:
The master/agent test is talking longer. Feel free to review and merge this, and I can add the extra test when it's ready.

spec/acceptance/class_spec.rb
```diff
@@ -3,22 +3,97 @@
 describe 'agent_upgrade class' do
 
   context 'default parameters' do
-    # Using puppet_apply as a helper
+    before(:all) { setup_puppet }
+    after (:all) { teardown_puppet }
+
     it 'should work idempotently with no errors' do
```
branan: Should we just mark this test as pending?

spec/acceptance/class_spec.rb
```diff
@@ -3,22 +3,97 @@
...
+      it { is_expected.to_not be_running }
+    end
+  end
+
+  context 'with mcollective configured' do
+    before(:all) { setup_puppet }
+    after (:all) { teardown_puppet }
+
+    it 'should work idempotently with no errors' do
+      pp = <<-EOS
```
branan: This should probably be moved to the `before(:all)` setup step, just for the sake of organization

spec/acceptance/class_spec.rb
MikaelSmith: I'd have to add an extra test that would be removed when it was no longer pending then. I think updating it is covered sufficiently by [PUP-4606](https://tickets.puppetlabs.com/browse/PUP-4606).

spec/acceptance/class_spec.rb
MikaelSmith: I'm not sure why. It's only needed for this case.

The spec/acceptance model has some funny behavior, like applying a change in one clause and then verifying the effect in others (within a context).

spec/acceptance/class_spec.rb
```diff
@@ -3,22 +3,97 @@
...
+      file { '/etc/mcollective': ensure => directory }
+      file { '/etc/mcollective/server.cfg':
+        ensure  => file,
+        notify  => Class['agent_upgrade'],
+        content => '#{File.read(File.join(FIXTURE_DIR, 'server.cfg'))}'
```
MikaelSmith: Hmm, forgot to add server.cfg

spec/spec_helper_acceptance.rb
```diff
@@ -1,30 +1,128 @@
...
-    puppet_module_install(:source => proj_root, :module_name => 'agent_upgrade')
+    step "Clear SSL on all hosts"
     hosts.each do |host|
-      on host, puppet('module', 'install', 'puppetlabs-stdlib'), { :acceptable_exit_codes => [0,1] }
+      stop_firewall_on host
+      ssldir = on(host, puppet('agent --configprint ssldir')).stdout.chomp
+      on(host, "rm -rf '#{ssldir}'")
+    end
+
+    step "Master: Start Puppet Master" do
```
MikaelSmith: Some of this master/agent co-ordination would be awesome in puppet_helpers (whether that's in Beaker or split into a Beaker helper module).

MikaelSmith:
Merge-able?

PR #5, opened 2015-05-13
(PUP-4559) Extend spec test to Puppet 3.7 and 4.0

MikaelSmith:
Extend spec testing to cover error case when using Puppet prior to 3.8,
and make no changes for Puppet 4.0+.

### DIFF
--- a/spec/classes/agent_upgrade_spec.rb
+++ b/spec/classes/agent_upgrade_spec.rb
@@ -14,26 +14,30 @@ describe 'agent_upgrade' do
               })
             end
 
-            [{}, {:service_names => []}].each do |params|
-              context "agent_upgrade class without any parameters" do
-                let(:params) { params }
+            if Puppet.version < "3.8.0"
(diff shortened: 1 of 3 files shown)


spec/classes/agent_upgrade_spec.rb
```diff
@@ -14,26 +14,33 @@
...
             end
 
-            [{}, {:service_names => []}].each do |params|
-              context "agent_upgrade class without any parameters" do
-                let(:params) { params }
+            if Puppet.version < "3.8.0"
+              it { expect { is_expected.to contain_package('agent_upgrade') }.to raise_error(Puppet::Error, /upgrading requires Puppet 3.8/) }
+            elsif Puppet.version >= "4.0.0"
+              # TODO: How to test this warning and lack of changes?
+              it { is_expected.to contain_package('agent_upgrade') }
```
MikaelSmith: Not sure how I want to handle this. It might be better to have Puppet 4+ behavior match 3.8, with an additional warning. Right now the module skips everything for Puppet 4+.

PR #10, opened 2015-05-20
(maint) Warn when using module in Puppet 4+

MikaelSmith:
In Puppet 4+, this module will continue to manage files related to the
upgrade. The main impact of this is reseting Puppet and Marionette
Collective configuration to defaults, which would be undesirable if
other changes are made.

Add a warning when running under Puppet 4+, reminding admins to remove
the module.


joshcooper:
Talked with @MikaelSmith about making the module effectively a noop when running with puppet 4

igalic:
:+1: for noop

PR #11, opened 2015-05-21
(PUP-4560) Add mcollective support

MikaelSmith:
Adds migrating mcollective configuration - client.cfg and server.cfg -
to the new expected locations; and updating logdir, libdir, and
plugin.yaml to account for new locations.


MikaelSmith:
I'm re-working this into multiple commits to make it a little easier to understand.

MikaelSmith:
Relies on #12, need to rebase after that's merged.

manifests/prepare.pp
```diff
@@ -51,14 +53,87 @@
...
+    }
+
+    if $::mco_server_settings {
+      $mco_server_libdir = $::mco_server_settings['libdir']
+      if $mco_server_libdir {
+        ini_setting { 'server/libdir':
+          section => '',
+          setting => 'libdir',
+          path    => $mco_server,
+          value   => "${mco_server_libdir}:/opt/puppetlabs/mcollective/plugins",
```
richardc: Forgive the driveby, and I may be looking too narrowly, but won't this repeatedly extend libdir?  On the first run it'll grab a libdir from a server.cfg, and write this extended value, then on the next run it'll grab the extended value and extend it again, and then on the next run the fact will grab it again, and then extend it again, etc.

Also libdir is allowed to be specified multiple times in a mcollective configuration file, as the server.cfg isn't really an ini file.  This is legit for a server.cfg: https://github.com/puppetlabs/marionette-collective/blob/master/ext/aio/common/server.cfg.dist#L4-L8

MikaelSmith: The feedback is great, I'll look into it more tomorrow.

The file resource will ensure it syncs to the original, then updates. I can't think of a scenario where repeatedly running this would result in extra paths.

Michael

richardc: The resolution of the $mco_server_settings fact parses a config file based on the value of the $mco_server_config fact, which is noted as having this behaviour:

```
# The mco_*_config facts will return the location of mcollective config (or nil), prefering PE over FOSS
```

Which would grab the AIO settings again, as PE and AIO paths for server.cfg are the same - `/etc/puppetlabs/mcollective/server.cfg`

That said, looking closer the implementation of $mco_server_config is actually going to pick FOSS over the PE/AIO path when both files exist, so I guess it might work out when migrating a FOSS install.

MikaelSmith: Good point on PE, I'll have to think that over a bit more.

MikaelSmith: I'll update this to only append if `libdir` doesn't already contain the new path. Same with `plugin.yaml`.

MikaelSmith: Update to the last commit should resolve this.

joshcooper:
@MikaelSmith needs a rebase

MikaelSmith:
Working on the rebase, it's a big one.

MikaelSmith:
Rebased and fixed.

joshcooper:
Sorry I just merged the other PR about puppet version 4.0, could you rebase again, sorry about that

manifests/prepare.pp
```diff
@@ -52,14 +54,103 @@
...
+    $mco_client = $::agent_upgrade::params::mco_client
+    file { $mco_client:
+      ensure  => file,
+      source  => $::mco_client_config,
+    }
+
+    if $::mco_client_settings {
+      $mco_client_libdir = $::mco_client_settings['libdir']
+      if $mco_client_libdir {
+        $mco_client_libdir_array = split($mco_client_libdir, ':')
```
joshcooper: I think we need to take in account windows vs posix when splitting the libdir, as windows agents will use semicolon as the separator.

manifests/prepare.pp
```diff
@@ -52,14 +54,103 @@
...
+            path    => $mco_client,
+            value   => "${mco_client_libdir}:$::agent_upgrade::params::mco_libdir",
+            require => File[$mco_client],
+          }
+        }
+      }
+
+      $mco_client_plugins = $::mco_client_settings['plugin.yaml']
+      if $mco_client_plugins {
+        $mco_client_plugins_array = split($mco_client_plugins, ':')
```
joshcooper: same platform splitting issue

MikaelSmith:
Should be ready now.

manifests/prepare.pp
```diff
@@ -52,14 +54,103 @@
...
+      $mco_server_libdir = $::mco_server_settings['libdir']
+      if $mco_server_libdir {
+        $mco_server_libdir_array = split($mco_server_libdir, $::agent_upgrade::params::path_separator)
+        # Only add the new path if it's not already in libdir
+        if [] == $mco_server_libdir_array.filter |$x| { $x == $::agent_upgrade::params::mco_libdir } {
+          ini_setting { 'server/libdir':
+            section => '',
+            setting => 'libdir',
+            path    => $mco_server,
+            value   => "${mco_server_libdir}:$::agent_upgrade::params::mco_libdir",
```
joshcooper: This should use `$::agent_upgrade::params::path_separator` instead of hard-coded colon, but can wait until we add windows support.

manifests/prepare.pp
```diff
@@ -52,14 +54,103 @@
...
+      ensure  => file,
+      source  => $::mco_server_config,
+    }
+
+    if $::mco_server_settings {
+      $mco_server_libdir = $::mco_server_settings['libdir']
+      if $mco_server_libdir {
+        $mco_server_libdir_array = split($mco_server_libdir, $::agent_upgrade::params::path_separator)
+        # Only add the new path if it's not already in libdir
+        if [] == $mco_server_libdir_array.filter |$x| { $x == $::agent_upgrade::params::mco_libdir } {
```
joshcooper: I thought about using puppet's [match](https://docs.puppetlabs.com/references/4.1.latest/function.html#match) function to check if the new path is already in libdir (instead of splitting and filtering). But it gets a little complicated. If newpath was `/a`, then you wouldn't want to match `/a/b`, and you'd have to handle cases where it's at the start, in the middle, at the end, or not present.

manifests/prepare.pp
```diff
@@ -52,14 +54,103 @@
...
+      $mco_client_libdir = $::mco_client_settings['libdir']
+      if $mco_client_libdir {
+        $mco_client_libdir_array = split($mco_client_libdir, $::agent_upgrade::params::path_separator)
+        # Only add the new path if it's not already in libdir
+        if [] == $mco_client_libdir_array.filter |$x| { $x == $::agent_upgrade::params::mco_libdir } {
+          ini_setting { 'client/libdir':
+            section => '',
+            setting => 'libdir',
+            path    => $mco_client,
+            value   => "${mco_client_libdir}:$::agent_upgrade::params::mco_libdir",
```
joshcooper: Same comment about using path_separator, and can wait for windows support.

spec/acceptance/class_spec.rb
```diff
@@ -97,12 +105,36 @@ class { 'agent_upgrade': }
...
+          assert_match(/^#{default}\s+time=/, stdout)
+        end
+    end
+
+    describe file('/etc/puppetlabs/mcollective/server.cfg') do
+      it { is_expected.to exist }
+      its(:content) {
+        is_expected.to include 'libdir = /usr/libexec/mcollective/plugins:/opt/puppetlabs/mcollective/plugins'
+        is_expected.to include 'logfile = /var/log/puppetlabs/mcollective.log'
+        is_expected.to include 'plugin.yaml = /etc/mcollective/facts.yaml:/etc/puppetlabs/mcollective/facts.yaml'
```
joshcooper: I see the ticket mentions to "Append the new libdir location to libdir" and we do that, but I'm wondering if that's the right thing to do?

Seems like the current behavior could be bad if you had an older version of an mco plugin in the old place (perhaps not compatible with puppet 4), and a newer version of the plugin in the new place, and we'd give precedence to the older one.

But perhaps mco doesn't allow multiple versions of the same plugin in the `libdir` and so this would never be a concern? /cc @richardc

spec/classes/agent_upgrade_spec.rb
```diff
@@ -4,47 +4,42 @@
...
-            else
-              [{}, {:service_names => []}].each do |params|
-                context "agent_upgrade class without any parameters" do
-                  let(:params) { params }
+        if Puppet.version < "3.8.0"
+          it { expect { is_expected.to contain_package('agent_upgrade') }.to raise_error(Puppet::Error, /upgrading requires Puppet 3.8/) }
+        else
+          [{}, {:service_names => []}].each do |params|
+            context "agent_upgrade class without any parameters" do
+              let(:params) { params }
```
joshcooper: It seems the value of this affects whether the compiled catalog contains service resources, but I didn't understand how that works, some rspec-puppet magic?

joshcooper:
I just have the one question about whether to prepend or append mco settings, but since that was specified in the ticket, I think we can merge as-is, and if needed, fixup later. The other issue about `params` is just a question for my own understand.

spec/acceptance/class_spec.rb
richardc: If there are multiple versions of the plugin on the `libdir` search path, mco just uses the first one it finds.

MikaelSmith: Meaning the new path should come first?

spec/classes/agent_upgrade_spec.rb
MikaelSmith: It's some rspec-puppet magic for compile. http://rspec-puppet.com/tutorial/, under `Specifying parameters`.

spec/acceptance/class_spec.rb
richardc: If we expect there to be newer/preferred plugins in that path, yes.

PR #12, opened 2015-05-21
(PUP-4567) Fixes for acceptance and Beaker 2.12

MikaelSmith:
Without this change, acceptance fails because `step` expects to be in
the context of a test. They also fail specifically in CI because the
master also includes the agent role.

* Pin to a working version of Beaker; BKR-283 tracks fixing it.
* Use default instead of agents for the machine being tested.
* Also only sync modules when needed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,6 @@ group :test do
 end
 
 group :system_tests do
-  gem "beaker"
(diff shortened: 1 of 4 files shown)


MikaelSmith:
This should be PUP-4567.

MikaelSmith:
Updated

PR #13, opened 2015-05-22
(PUP-4561) Extract platform-specific code

MikaelSmith:
Before this commit, the `prepare` class was full of platform-specific
code that needed to be refactored before non-RedHat platforms could be
supported.

Extract the platform-specific code to `params` and a new class that is
selected based on the `osfamily` fact.


MikaelSmith:
This will definitely have a merge conflict with #11

manifests/osfamily/redhat.pp
```diff
@@ -0,0 +1,45 @@
+class agent_upgrade::osfamily::redhat {
+  #  if $caller_module_name != $module_name {
+  #  fail("Use of private class ${name} by ${caller_module_name}")
+  #}
```
joshcooper: can be removed?

MikaelSmith: Oh right, I was going to try the guard I saw in another repo. I'll check whether I want to enable it or remove it completely.

manifests/prepare.pp
```diff
@@ -53,15 +54,25 @@
...
 
-  # Install PC1 yum repo; based off puppetlabs_yum
-  contain '::agent_upgrade::puppetlabs_yum'
+  # PLATFORM SPECIFIC CONFIGURATION
+  # Break out the platform-specific configuration into subclasses, dependent on
+  # the osfamily of the client being configured.
+
+  case $::osfamily {
+    'redhat', 'debian', 'windows', 'solaris', 'aix', 'suse': {
+      contain downcase("::agent_upgrade::osfamily::${::osfamily}")
```
joshcooper: Is this a common module paradigm for "contain"ing platform specific functionality?

MikaelSmith: It's one I borrowed from the pe-agent module.

PR #14, opened 2015-05-26
(maint) Use Puppet 3.8 gems

MikaelSmith:
Puppet 3.8 gems were released as Puppet 3.8.1. Use those for testing,
rather than git.


joshcooper:
Looks like there's a file named `-`!

PR #16, opened 2015-05-30
pup-4638-rename-all-the-things

kylog:


MikaelSmith:
Naming will overlap with https://forge.puppetlabs.com/bmurt/puppet_agent, but https://forge.puppetlabs.com/stephenrjohnson/puppet seems like the more frequently used and maintained module for managing agent installs.

PR #18, opened 2015-06-02
Add changelog

kylog:


kylog:
@mhaskel suggested we add a change log prior to release, /cc @MikaelSmith.

MikaelSmith:
The README mentions centos and fedora, we should probably bring these in-sync.

kylog:
Ah I was going off metadata.json, which says rhel 5/6/7.

MikaelSmith:
We should probably get those in-sync. It should work on any rpm-based distro, but haven't tested it everywhere. Looking at puppetlabs-concat, I'd probably make them all say RedHat 5/6/7 and CentOS 5/6/7.

kylog:
Yeah, I'll do that. Push coming.

kylog:
Okey doke, those should be synced up now.

PR #23, opened 2015-06-03
(maint) Use latest gpg key from yum.puppetlabs.com

MikaelSmith:
The prior file was sourced from https://github.com/stahnma/puppet-module-puppetlabs_yum/blob/master/files/RPM-GPG-KEY-puppetlabs, new one is http://yum.puppetlabs.com/RPM-GPG-KEY-puppetlabs.


MikaelSmith:
The prior file was sourced from https://github.com/stahnma/puppet-module-puppetlabs_yum/blob/master/files/RPM-GPG-KEY-puppetlabs, new one is http://yum.puppetlabs.com/RPM-GPG-KEY-puppetlabs.

kylog:
:+1: 

I'll let travis catch up and then push the button.

PR #25, opened 2015-06-04
(PUP-4658) Update acceptance for fix

MikaelSmith:
A prior commit for this ticket changed the order if mcollective's libdir
config, but not the acceptance test verifying it. Update the acceptance
test to pass again.


MikaelSmith:
Manually ran acceptance successfully.

PR #26, opened 2015-06-04
(PUP-4478) Add Deb package support

MikaelSmith:
Adds support for Debian 6 and 7, and Ubuntu 12.04, 14.04, and 14.10.


MikaelSmith:
Should we call out Fedora 21? We didn't support it in the Puppet 3.x line, so there's not much for the module to do yet.

jbondpdx:
Yeah, I think it's still good to mention it. We don't have to make a big
deal of it, though.

On Fri, Jun 5, 2015 at 8:34 AM, Michael Smith <notifications@github.com>
wrote:

> Should we call out Fedora 21? We didn't support it in the Puppet 3.x line,
> so there's not much for the module to do yet.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/26#issuecomment-109330968>
> .
>



-- 
Jean Bond
Technical Documentation
Puppet Labs

spec/spec_helper_acceptance.rb
```diff
@@ -145,9 +163,20 @@ def teardown_puppet_on(host)
...
-package { ['puppet-agent', 'puppet', 'mcollective', 'mcollective-client', 'activemq']: ensure => purged }
-file { ['/etc/puppet', '/etc/puppetlabs', '/etc/mcollective', '/etc/activemq']: ensure => absent, force => true, backup => false }
-yumrepo { 'pc1_repo': ensure => absent }
+package { ['puppet-agent', 'puppet', 'mcollective', 'mcollective-client']: ensure => purged }
+file { ['/etc/puppet', '/etc/puppetlabs', '/etc/mcollective']: ensure => absent, force => true, backup => false }
+#{
+  case host['platform']
+  when /debian|ubuntu/
+    on host, '/opt/puppetlabs/bin/puppet module install puppetlabs-apt', { :acceptable_exit_codes => [0,1] }
+    "include apt\napt::source { 'pc1_repo': ensure => absent }"
```
joshcooper: @MikaelSmith seeing a strange error when puppet_agent tries to purge the deb repo configs: https://gist.github.com/joshcooper/0be2740f2a1be0b194d4#file-gistfile1-txt so I think state is getting left behind?

MikaelSmith:
Fixed test cleanup on Debian - puppet apply was uninstalling itself before removing the repo, where it failed because a node terminus is missing.

spec/spec_helper_acceptance.rb
```diff
@@ -145,9 +163,20 @@ def teardown_puppet_on(host)
   # Note pc1_repo is specific to the module's manifests. This is knowledge we need to clean
   # the machine after each run.
   pp = <<-EOS
-package { ['puppet-agent', 'puppet', 'mcollective', 'mcollective-client', 'activemq']: ensure => purged }
-file { ['/etc/puppet', '/etc/puppetlabs', '/etc/mcollective', '/etc/activemq']: ensure => absent, force => true, backup => false }
-yumrepo { 'pc1_repo': ensure => absent }
+#{
```
joshcooper: typo?

spec/spec_helper_acceptance.rb
```diff
@@ -144,10 +162,21 @@ def teardown_puppet_on(host)
   step "Purge puppet from #{host}"
   # Note pc1_repo is specific to the module's manifests. This is knowledge we need to clean
   # the machine after each run.
+  case host['platform']
+  when /debian|ubuntu/
+    on host, '/opt/puppetlabs/bin/puppet module install puppetlabs-apt', { :acceptable_exit_codes => [0,1] }
```
joshcooper: Are there cases where we won't have the module installed already? Also minor nit, `on host, puppet('module', ...)`?

spec/spec_helper_acceptance.rb
```diff
@@ -144,10 +162,21 @@ def teardown_puppet_on(host)
...
+
   pp = <<-EOS
-package { ['puppet-agent', 'puppet', 'mcollective', 'mcollective-client', 'activemq']: ensure => purged }
-file { ['/etc/puppet', '/etc/puppetlabs', '/etc/mcollective', '/etc/activemq']: ensure => absent, force => true, backup => false }
-yumrepo { 'pc1_repo': ensure => absent }
+#{clean_repo}
+file { ['/etc/puppet', '/etc/puppetlabs', '/etc/mcollective']: ensure => absent, force => true, backup => false }
+package { ['puppet-agent', 'puppet', 'mcollective', 'mcollective-client']: ensure => purged }
   EOS
   on host, "/opt/puppetlabs/bin/puppet apply -e \"#{pp}\""
```
joshcooper: probably want to use `on host, puppet('apply', ...` here too

spec/spec_helper_acceptance.rb
MikaelSmith: Yes, always because the migration module doesn't move modules. It probably should, but I'd file another ticket for that.

On puppet helpers, that was BKR-273 (Beaker paths aren't setup for AIO after the upgrade), and is fixed in another PR.

spec/spec_helper_acceptance.rb
MikaelSmith: Can't for the same reason as above.

joshcooper:
ah, right, makes sense, will test and merge tomorrow.

PR #30, opened 2015-06-09
(maint) Use puppet helper for ssldir

MikaelSmith:
Retrieve ssldir from the host using the #puppet helper.


MikaelSmith:
Pulled into another PR.

PR #28, opened 2015-06-05
(PUP-4603) Use Puppet Server 2.1+ for acceptance tests

MikaelSmith:
- (maint) Update Beaker to latest and pin gems
- (maint) Remove work-around for BKR-262
- (PUP-4603) Use Puppet Server 2.1+

PR #31, opened 2015-06-09
(maint) Only test specs per osfamily

MikaelSmith:
Behavior only changes between os families, not between os. Specs are
slow doing many resource checks for each family, so do them per os
family instead.

### DIFF
--- a/spec/classes/puppet_agent_prepare_spec.rb
+++ b/spec/classes/puppet_agent_prepare_spec.rb
@@ -6,8 +6,8 @@ MCO_PLUGIN_YAML = '/etc/puppetlabs/mcollective/facts.yaml'
 MCO_LOGFILE = '/var/log/puppetlabs/mcollective.log'
 
 describe 'puppet_agent::prepare' do
-  context 'supported operating systems' do
-    on_supported_os.each do |os, facts|
+  context 'supported operating system families' do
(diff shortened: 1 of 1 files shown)


spec/classes/puppet_agent_prepare_spec.rb
```diff
@@ -6,8 +6,8 @@
 MCO_LOGFILE = '/var/log/puppetlabs/mcollective.log'
 
 describe 'puppet_agent::prepare' do
-  context 'supported operating systems' do
-    on_supported_os.each do |os, facts|
+  context 'supported operating system families' do
+    on_supported_os.to_a.uniq { |os, facts| facts[:osfamily] }.each do |os, facts|
```
joshcooper: This doesn't work as expected in ruby 1.8.7:

```
irb(main):005:0>  {'debian-7' => 'debian', 'debian-6' => 'debian' }.to_a.uniq { |k,v| v }.each {|k,v| puts "#{k}=#{v}"}
debian-7=debian
debian-6=debian
=> [["debian-7", "debian"], ["debian-6", "debian"]]
```

But does in ruby 1.9 and up:

```
irb(main):001:0> {'debian-7' => 'debian', 'debian-6' => 'debian' }.to_a.uniq { |k,v| v }.each {|k,v| puts "#{k}=#{v}"}
debian-7=debian
```

MikaelSmith:
Added to #28.

PR #32, opened 2015-06-09
(maint) Use BKR-314 and BKR-317 fixes

MikaelSmith:
BKR-314 and BKR-317 identified issues with using Beaker helpers for
upgrade scenarios. Simplify the acceptance setup by using Beaker helpers
now that they're fixed.

Beaker 2.15 was released with fixes for [BKR-314](https://tickets.puppetlabs.com/browse/BKR-314) and [BKR-317](https://tickets.puppetlabs.com/browse/BKR-317).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,6 @@ group :test do
 end
 
 group :system_tests do
-  gem 'beaker', '~> 2.15'
+  gem 'beaker', '~> 2.16'
   gem 'beaker-rspec', '~> 5.1'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,6 @@ group :test do
 end
 
 group :system_tests do
-  gem 'beaker', '~> 2.14'
+  gem 'beaker', '~> 2.15'
   gem 'beaker-rspec', '~> 5.1'
 end
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -100,7 +100,7 @@ end
 def setup_puppet_on(host, opts = {})
   opts = {:agent => false, :mcollective => false}.merge(opts)
 
-  step "Setup puppet on #{host}"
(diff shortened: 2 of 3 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -84,9 +81,7 @@ def setup_puppet_on(host, opts = {})
   opts = {:agent => false, :mcollective => false}.merge(opts)
 
   step "Setup puppet on #{host}"
-  install_package host, 'puppet'
-  add_foss_defaults_on host
-  add_puppet_paths_on host
+  install_puppet_on host
```
cyberious: We should be adding options here for version and architecture for windows users?  
https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/spec_helper_acceptance.rb#L7 as well as version numbers.  
Also the above `step` method should be removed as this has been yanked for beaker-rspec

MikaelSmith: All the `step` methods should be removed, but I hadn't heard they were officially yanked from beaker-rspec.

Anything that needs to be added for Windows would happen under PUP-4479.

MikaelSmith:
@ericwilliamson mind taking a look over this and merging if it looks good? Jenkins tests are broken because of Beaker 2.16, and this fixes them.

Gemfile
```diff
@@ -11,6 +11,6 @@ group :test do
 end
 
 group :system_tests do
-  gem 'beaker', '~> 2.14'
+  gem 'beaker', '~> 2.15'
```
ericwilliamson: you mention it's broken for beaker 2.16, should we be pinning to that instead of 2.15 as the minimum version?

MikaelSmith: The behavior changed, so that makes sense.

ericwilliamson:
:+1:

PR #33, opened 2015-06-10
(PUP-4479) Agent upgrade windows

cyberious:


lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+    "puppet-agent-#{resource[:version]}-#{arch}.msi"
+  end
+
+
+  def get_command
+    tempfile = new_logfile_path
+    debug "Spawning installer with log at '#{tempfile}'"
+    "cmd.exe /c \"\"#{native_path(command(:powershell))}\"" +
+      " -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass -Command " +
+      "\"Wait-Process #{Process.pid} -ErrorAction SilentlyContinue;msiexec /qn /i #{source_url} /l*v #{tempfile} \""
```
ferventcoder: SilentlyContinue :panda_face:

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+    native_path(Tempfile.new(['puppet-agent-installer', '.log']).path)
+  end
+
+  def spawn_upgrade_process
+    args = {
+      :command_line => get_command,
+      :creation_flags => Process::DETACHED_PROCESS,
+      :process_inherit => false,
+      :thread_inherit => true,
+      :cwd => 'C:\\'
```
ferventcoder: Don't assume C exists. Look for systemroot instead.

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    "puppet-agent-#{resource[:version]}-#{arch}.msi"
+  end
+
+
+  def get_command
+    tempfile = new_logfile_path
+    debug "Spawning installer with log at '#{tempfile}'"
+    "cmd.exe /c \"\"#{native_path(command(:powershell))}\"" +
```
ferventcoder: Why do you spawn cmd to call for powershell?

ferventcoder: Just call powershell directly

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  newproperty(:version) do
+    defaultto 'latest'
+    validate do |value|
+      if value !~ /^(latest|\d+\.\d+\.\d+)$/
+        fail "Valid values are 'latest' or full version numbers, you provided '#{value}'"
+      end
+    end
+  end
+end
```
ferventcoder: Was this type never created before for others?

manifests/params.pp
```diff
@@ -29,6 +29,9 @@
 
       $path_separator = ':'
     }
+    'windows' : {
+
```
ferventcoder: So nothing needs to be set then?

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+      :command_line => get_command,
+      :creation_flags => Process::DETACHED_PROCESS,
+      :process_inherit => false,
+      :thread_inherit => true,
+      :cwd => 'C:\\'
+    }
+
+    debug "Creating wait process for #{Process.pid}"
+    upgradeProcess = Process.create(args)
+    debug ("Created background process to wait for puppet exit at PID #{upgradeProcess.process_id}")
```
ferventcoder: With this debug message you actually need to tell the user what you are doing and where they should follow up if the agent is not upgraded. I would create a secondary log from that process that tells them that the process was successful or not (pointing to the msi install log as well for further review), plus how they move forward with getting support for failures.

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
...
+        fail "Please provide a valid http(s), puppet or unc path"
+      end
+      if value =~ /^https?/ && value !~ URI::regexp
+        fail("Please provide a valid URI")
+      end
+    end
+  end
+
+  newparam(:arch) do
+    newvalues /^(x86|x64)$/
```
ferventcoder: Default to?

ferventcoder: I see you did the following in the provider - Facter.value(:architecture)

I would ensure that you mention that here in the description of this you are going to write.

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
+
+Puppet::Type.type(:agent_upgrade).provide(:agent_upgrade) do
+  confine :osfamily => /windows/i
```
ferventcoder: This looks wrong to me?

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  mk_resource_methods
```
ferventcoder: I'm not sure what this either.

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Needs defaultfor as well.

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  commands :powershell =>
+             if File.exists?("#{ENV['SYSTEMROOT']}\\sysnative\\WindowsPowershell\\v1.0\\powershell.exe")
+               "#{ENV['SYSTEMROOT']}\\sysnative\\WindowsPowershell\\v1.0\\powershell.exe"
+             elsif File.exists?("#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe")
+               "#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe"
+             else
+               'powershell.exe'
+             end
```
ferventcoder: Typically you create this so you can `powershell(somearg, arg2, argN)` - in the way that you ended up using it, it may be better to use a variable instead.

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
...
+  newparam(:name, :namevar => true)
+
+  newparam(:source) do
+    newvalues /^((https?|puppet):\/\/)|\\\\\S+\\\w+|[a-z]:\\/i
+    validate do |value|
+      if value !~ /^((https?|puppet):\/\/)|\\\\\S+\\\w+|[a-z]:\\/i
+        fail "Please provide a valid http(s), puppet or unc path"
+      end
+      if value =~ /^https?/ && value !~ URI::regexp
+        fail("Please provide a valid URI")
```
ferventcoder: Interesting, so we are stating we know all the ways the user may choose to upgrade and we are going to limit them. I'd almost prefer to let the user decide, but check what they are attempting to install to be sure it is not an older version of Puppet.

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
+Puppet::Type.newtype(:agent_upgrade) do
+  ensurable
+
+  newparam(:name, :namevar => true)
+
+  newparam(:source) do
```
ferventcoder: Needs `desc` section. Yes you write this. :)

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
+Puppet::Type.newtype(:agent_upgrade) do
+  ensurable
+
+  newparam(:name, :namevar => true)
```
ferventcoder: Needs `desc` section.

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
...
+      if value !~ /^((https?|puppet):\/\/)|\\\\\S+\\\w+|[a-z]:\\/i
+        fail "Please provide a valid http(s), puppet or unc path"
+      end
+      if value =~ /^https?/ && value !~ URI::regexp
+        fail("Please provide a valid URI")
+      end
+    end
+  end
+
+  newparam(:arch) do
```
ferventcoder: Needs `desc` section.

lib/puppet/type/agent_upgrade.rb
```diff
@@ -0,0 +1,30 @@
...
+    newvalues /^(x86|x64)$/
+  end
+
+  newproperty(:version) do
```
ferventcoder: Needs `desc` section

lib/puppet/type/agent_upgrade.rb
ferventcoder: Mention this is the full path to the msi as well.

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: In the unlikely event that the spin up time for powershell is > than the ruby exit, than Wait-Process will Error stating the process does not exist.  We only care that it is no longer running.

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: PowerShell was not spawning up properly when called directly.  Used previous art from powershell module

lib/puppet/type/agent_upgrade.rb
cyberious: All others are using standard manifests not requiring a type and provider.

manifests/params.pp
cyberious: Everything is handled in the provider.  We could do the source here but doesn't make sense when we have a type and provider to handle the logic

ferventcoder: :+1:

lib/puppet/type/agent_upgrade.rb
ferventcoder: That's what I though.

lib/puppet/type/agent_upgrade.rb
cyberious: This is private provider and should be used from puppet_agent class

cyberious: Which we default to $::architecture in the class puppet_agent

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Makes sense. Include a comment here as to such.

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Oh right, that is just for wait process. Carry on. :)

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: This creates the accessor methods for properties allowing us to set them for self.instances

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Also, there is a bug in PowerShell v3 surrounding waiting processes.

ferventcoder: Nevermind, the bug is in Start-Process -Wait

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: This is to ensure we call the which for the command and demand that it is there.

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: :+1:

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Stdlib thing?

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: I just meant, this is a little heavy weight to use IMHO. You could achieve the same setting a variable.

lib/puppet/type/agent_upgrade.rb
cyberious: Again this is a private provider, if they abuse the usage and call agent_upgrade directly we are not going to stop them but ensure the URL they are provide is valid.

lib/puppet/type/agent_upgrade.rb
ferventcoder: Without a default, does it error when not set?

lib/puppet/provider/agent_upgrade/windows.rb
joshcooper: also it's an insecure world writable directory, and could lead to privilege escalation due to the way CreateProcess works

lib/puppet/type/agent_upgrade.rb
ferventcoder: Nevermind, that will be handled by the public class, not the private type.

lib/puppet/type/agent_upgrade.rb
ferventcoder: Nevermind, that will be handled by the public class, not the private type.

lib/puppet/type/agent_upgrade.rb
ferventcoder: Nevermind, that will be handled by the public class, not the private type.

lib/puppet/type/agent_upgrade.rb
ferventcoder: I'm actually worried about this one - it's not agnostic and the check should be implemented up in the public class, not in here. The check is going to need to be done anyway, implementing it here just means it needs to be reimplemented in the class. However I'm worried both about duplication of work and that one will be more restrictive than another, leading to incorrect documentation as such.

Plus it should likely be more agnostic.

joshcooper: Probably better to call `uri = URI.parse(value)` and them verify `uri.scheme` is one of the supported values. That said, I don't think we can support `puppet` URIs, as the agent does some magic converting `puppet:///modules/mymodule/file` into an HTTP request.

lib/puppet/type/agent_upgrade.rb
ferventcoder: Nevermind, that's up in the public class that calls this private type/provider.

joshcooper: for an enumeration of values, you can just list them directly: `newvalues "x86", "x64"` like we do for file checksums

ferventcoder:
How does one make the type private? I don't see this as part of the PR.

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
...
+  def new_logfile_path
+    native_path(Tempfile.new(['puppet-agent-installer', '.log']).path)
+  end
+
+  def spawn_upgrade_process
+    args = {
+      :command_line => get_command,
+      :creation_flags => Process::DETACHED_PROCESS,
+      :process_inherit => false,
+      :thread_inherit => true,
```
joshcooper: we do we need to specify inherit flags?

cyberious:
Can add, do we need to though?
https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/assert_private.rb

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: We need to ensure we have correct privileges to download and install.

lib/puppet/type/agent_upgrade.rb
ferventcoder: The agnostic sense is that I want to upgrade all of my clients to the latest. Most folks download once and make it available there. That means now that I need to fully qualify source to the file that I need a selector/case statement for all platforms to be sure I get the right file for each platform. 


cc @joshcooper @MikaelSmith @kylog

ferventcoder:
@cyberious that doesn't look like the actual type/provider can be set private. Am I missing something?

lib/puppet/type/agent_upgrade.rb
cyberious: @joshcooper I had the same thought and decided the user can use `file` resource and tell us where they download it to outside of the scope of this module

lib/puppet/provider/agent_upgrade/windows.rb
MikaelSmith: Why not use something like https://forge.puppetlabs.com/puppetlabs/powershell?

MikaelSmith:
This commit could use a description, like why we use a type/provider in the first place.

manifests/init.pp
```diff
@@ -29,14 +34,23 @@
     info('puppet_agent performs no actions on Puppet 4+')
   }
   else {
-    class { '::puppet_agent::prepare': } ->
-    class { '::puppet_agent::install': } ->
-    class { '::puppet_agent::config': } ~>
-    class { '::puppet_agent::service': }
+    if $::osfamily == 'windows' {
+      agent_upgrade { 'Windows Puppet Agent Upgrade':
```
MikaelSmith: This skips everything done to fix up `puppet.conf` and the `mcollective` configs.

MikaelSmith:
Does this pass acceptance tests?

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,85 @@
+
+Puppet::Type.type(:agent_upgrade).provide(:agent_upgrade) do
```
joshcooper: Since `puppet-agent` is a package, and can already be managed using the built-in `windows` package provider, I think it is confusing to have a type & provider that do a similar thing. For example, the instances method below returns `puppet-agent` packages.

Another option would be to create a class that:

1. Validates the parameters, e.g. source URL, arch, ...
1. Uses the powershell module to execute powershell code passing in the current ruby process id.
1. The powershell code spawns another detached powershell instance that waits for the ruby process to exit, and then executes msiexec

lib/puppet/type/agent_upgrade.rb
ferventcoder: @cyberious then it may be a good idea to allow the path to be to a file.

manifests/init.pp
ferventcoder: Yeah, so this still need to handle those as well.

joshcooper:
Since `puppet-agent` is a package, and can already be managed using the built-in `windows` package provider, I think it is confusing to have a type & provider that do a similar thing. For example, the instances method returns `puppet-agent` packages.

Another option would be to create a puppet class that:

1. Validates the parameters, e.g. source URL, arch, ...
1. Uses the powershell module to execute powershell code passing in the current ruby process id.
1. The powershell code spawns another detached powershell instance that waits for the ruby process to exit, and then executes msiexec.

lib/puppet/type/agent_upgrade.rb
cyberious: @joshcooper verifying that URI.parse will still require us to rescue failures for file paths so would be more code to validate instead.

ferventcoder:
@MikaelSmith good call on the commit message - it needs a whole lot of love.

lib/puppet/type/agent_upgrade.rb
cyberious: Updated

ferventcoder:
>Since puppet-agent is a package, and can already be managed using the built-in `windows` package provider, I think it is confusing to have a type & provider that do a similar thing. 

@joshcooper I'm confused on this a bit - we've already proven that Puppet doesn't do a good job handling upgrades to itself on Windows because it has to stop Puppet midstream and then perform the upgrade and runs into random issues. I'm thinking you meant this a different way than I understood it?

cyberious:
@joshcooper in Puppet Code, how do you get the Current Process ID to watch for?  The PowerShell provider is not something I want to depend on for this.  Which ruby process do you want to wait for? What if the user is running a irb session locally or another app that is Ruby?  It might never execute.

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Also a valid point.

cyberious:
@MikaelSmith @cowofevil has declared this to be manual testing for Windows

lib/puppet/provider/agent_upgrade/windows.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  @doc = <<-EOT
+    Windows provider to upgrade from 3.8 to 4.0 using latest puppet-agent version,
+    requires PowerShell to be already installed, only upgrade if < 4.0
+  EOT
+  confine :osfamily => /windows/i
```
ferventcoder: Back to this, what is `/windows/i`?

manifests/init.pp
cyberious: Is there something fundamentally different for windows puppet.conf and mcollective that I am unaware of?

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Do you need a defaultfor as well?

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: Preferred path ENV['SYSTEMROOT'] ? or somewhere else?

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: Do you have a suggestion for notification as well as verbiage about support?

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: Because we can not detect the Current Running process ID from puppet code, as well detached.  Below has some comments on this.

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: If there is only one and we are confine to windows and any other provider, which I don't see happening confines to there OS there is no need for default as they are all confined

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Doesn't sound like that would be a good place to end up. Perhaps temp location?

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: "We are going to upgrade the Puppet Agent as soon as Puppet finishes running the catalog. Please be patient and give it about 10-15 minutes to complete. We will log the progress in *logfilelocation* and the detailed installation log will be at *msiloglocation*. If the node doesn't come back up, please check one or both of these log files to determine what may have went wrong."

As far as what support would also want to state in there, best to ask them.

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: I agree and if not c:\ where?  I am open to suggestions.  ENV['TEMP']?

lib/puppet/provider/agent_upgrade/windows.rb
ferventcoder: Of course this means your powershell script needs to become a bit more, probably needs to be a file that this calls and runs.

lib/puppet/provider/agent_upgrade/windows.rb
MikaelSmith: Once upon a time there was https://projects.puppetlabs.com/issues/10112.

lib/puppet/provider/agent_upgrade/windows.rb
cyberious: So a double log?  What about a single log file that references the above text with the logfile we are passing to msiexec?

joshcooper:
@cyberious @ferventcoder this is what I'm talking about: https://github.com/joshcooper/detach

Use an exec resource to create a detached powershell instance. Have the powershell instance wait for puppet/ruby to exit, and then execute msiexec.

joshcooper:
```
C:\work\puppet>bundle exec puppet apply -e 'include detach' --modulepath c:\work\modules
DL is deprecated, please use Fiddle
Notice: Compiled catalog for win-qp47voha2p4.solar.lan in environment production in 0.67 seconds
Notice: /Stage[main]/Detach/File[C:\Users\albert\AppData\Local\Temp\install-puppet.ps1]/content: content changed '{md5}e
aaee688a2cdb8b1cde9ed08fad46fd1' to '{md5}5b68354b5cb6e7ad61ff4a2485cf5f25'
Notice: /Stage[main]/Detach/Exec[detach]/returns: executed successfully
Notice: /Stage[main]/Detach/Exec[sleep]/returns:
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Pinging 127.0.0.1 with 32 bytes of data:
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Notice: /Stage[main]/Detach/Exec[sleep]/returns:
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Ping statistics for 127.0.0.1:
Notice: /Stage[main]/Detach/Exec[sleep]/returns:     Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Notice: /Stage[main]/Detach/Exec[sleep]/returns: Approximate round trip times in milli-seconds:
Notice: /Stage[main]/Detach/Exec[sleep]/returns:     Minimum = 0ms, Maximum = 0ms, Average = 0ms
Notice: /Stage[main]/Detach/Exec[sleep]/returns: executed successfully
Notice: Applied catalog in 5.16 seconds
```

as soon as puppet finishes, powershell executes its command, in this case launching Notepad

cyberious:
That is a very clever way to go about it.  With that implimentation do you have any concerns around TOCTOU vulnerability as we are creating a tempfile?

joshcooper:
> do you have any concerns around TOCTOU vulnerability as we are creating a tempfile?

On Windows, temp directories are not world writable, and instead are per-user, even for LocalSystem. So as long as the owner/group/mode are specified on the file (or use the acl module) I'm not too concerned about it.

```
C:\work\puppet>icacls c:\users\albert\AppData\Local\Temp
c:\users\albert\AppData\Local\Temp NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
                                   BUILTIN\Administrators:(I)(OI)(CI)(F)
                                   WIN-QP47VOHA2P4\albert:(I)(OI)(CI)(F)
C:\work\puppet>icacls  C:\Windows\Temp
C:\Windows\Temp NT SERVICE\TrustedInstaller:(I)(F)
                NT SERVICE\TrustedInstaller:(I)(CI)(IO)(F)
                NT AUTHORITY\SYSTEM:(I)(F)
                NT AUTHORITY\SYSTEM:(I)(OI)(CI)(IO)(F)
                BUILTIN\Administrators:(I)(F)
                BUILTIN\Administrators:(I)(OI)(CI)(IO)(F)
                BUILTIN\Users:(I)(RX)
                BUILTIN\Users:(I)(OI)(CI)(IO)(GR,GE)
                CREATOR OWNER:(I)(OI)(CI)(IO)(F)
```
Alternatively, you could specify the powershell code to execute inline when specifying the `exec` resource.

joshcooper:
> That is a very clever way to go about it. 

So I'm using a fact to collect the pid, using an exec resource to execute powershell asynchronously, and using a two line powershell script to execute msiexec. I think that's about as simple as it gets.

As an aside, it would be nice if the `exec` type had a `background` parameter and created the process detached for us. It would be trivial to do on the windows side, just need to pass the appropriate parameter, and then use a powershell resource directly:

```
exec { 'install-puppet':
  command => "Wait-Process $pid\nmsiexec.exe /qn /i http://downloads.puppetlabs.com/windows/puppet-agent-1.1.0-x64.exe"
  background => true,
  provider => powershell,
}
```

ferventcoder:
Since it's a separate file as well, introducing the logging for the file is also nice.

cyberious:
@joshcooper I would love to see background true, how you'd you report back success for puppet reporting?  We should file that as a ticket for powershell module.

cyberious:
@ferventcoder wouldn't that be increase scope as well?

ferventcoder:
Nah, logging is pretty cheap to do.

cyberious:
We are currently messaging in the debug context, we can change that to a notify context as well instead of debug only.  I don't believe creating two files about the logging is necessary however.

MikaelSmith:
To jump on Josh's spike, I prefer to see as much done in puppet code to make this another example of how to use Puppet to do things on Windows.

cyberious:
@MikaelSmith I think using this as a springboard for how to do things in puppet code for windows has a time and place, if you look at this example it is pretty much do all this stuff except if windows.

cyberious:
@joshcooper @MikaelSmith @ferventcoder closing PR in favor of #34

PR #35, opened 2015-06-12
(PUP-4690) Migrate SSL stuff based on individual settings

branan:
Previously the ssldir was migrated monolithically. Unfortunately,
Puppet allows setting paths for several parts of the ssldir
individually. This now will take those disparate settings and combine
them into one ssldir post-migration.

It's a lot like voltron, but with more SSL.

This does not yet have acceptance, as beaker-rspec with vagrant and virtualbox is causing me grief. I'll get acceptance up ASAP, but for now the code itself is reviewable.

### DIFF
--- a/manifests/prepare.pp
+++ b/manifests/prepare.pp
@@ -16,11 +16,40 @@ class puppet_agent::prepare {
     ensure => directory,
   }
 
-  file { $::puppet_agent::params::ssldir:
-    ensure  => directory,
-    source  => $::puppet_ssldir,
-    backup  => false,
-    recurse => true,
+  $ssl_dir = $::puppet_agent::params::ssldir
+  file { $ssl_dir:
+    ensure   => directory,
+    source   => $::puppet_ssldir,
+    backup   => false,
+    recurse  => false,
+  }
+
+  $sslpaths = {
+    'certdir'       => 'certs',
+    'privatedir'    => 'private',
+    'privatekeydir' => 'private_keys',
+    'publickeydir'  => 'public_keys',
+    'requestdir'    => 'certificate_requests',
+  }
+
+  $sslpaths.each |String $setting, String $subdir| {
+    if $::puppet_sslpaths[$setting]['path_exists'] {
+      file { "${ssl_dir}/${subdir}":
+        ensure  => directory,
+        source  => $::puppet_sslpaths[$setting]['path'],
+        backup  => false,
+        recurse => true,
+      }
+    }
+  }
+
+  # The only one that's a file, not a directory.
+  if $::puppet_sslpaths['hostcrl']['path_exists'] {
+    file { "${ssl_dir}/crl.pem":
+      ensure  => file,
+      source  => $::puppet_sslpaths['hostcrl']['path'],
+      backup  => false
+    }
   }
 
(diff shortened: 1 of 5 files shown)


manifests/prepare.pp
```diff
@@ -16,11 +16,40 @@
...
-    ensure  => directory,
-    source  => $::puppet_ssldir,
-    backup  => false,
-    recurse => true,
+  $ssl_dir = $::puppet_agent::params::ssldir
+  file { $ssl_dir:
+    ensure   => directory,
+    source   => $::puppet_ssldir,
+    backup   => false,
+    recurse  => false,
```
MikaelSmith: We no longer recurse? What if none of the subdir settings are set?

branan: The settings have default values pointing inside the ssldir, so if the user hasn't mucked with anything behavior will be identical.

branan: well, unless their agent is a CA, which won't get migrated at the moment.

branan: @MikaelSmith updated the fact so its now "path_exists". Should hopefully be clearer what's happening now.

MikaelSmith:
This looks good once acceptance tests are accounted for.

manifests/prepare.pp
```diff
@@ -16,11 +16,40 @@
     ensure => directory,
   }
 
-  file { $::puppet_agent::params::ssldir:
-    ensure  => directory,
-    source  => $::puppet_ssldir,
-    backup  => false,
-    recurse => true,
+  $ssl_dir = $::puppet_agent::params::ssldir
```
cyberious: Any reason to not just use ```$::puppet_agent::params::ssldir``` instead of creating an additional var?

branan: It's really long and I got sick of typing it.

branan: I also think it does make the lines below where we append to that path slightly more readable

MikaelSmith:
All ssldir default options at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L633-L758.

spec/classes/puppet_agent_prepare_spec.rb
```diff
@@ -102,7 +128,22 @@
           'ensure'  => 'directory',
           'source'  => '/dev/null/ssl',
           'backup'  => 'false',
-          'recurse' => 'true',
+          'recurse' => 'false',
+        }) }
+
+        ['certificate_requests', 'certs', 'private', 'private_keys', 'public_keys'].each do |dir|
```
MikaelSmith: Some testing for `path_exists == false`?

branan: I'll add that

branan:
I finally got the damn acceptance to run! it passes for me!

I don't think it's worth adding new acceptance tests for this - the current ones should cover "we copy SSL from the individual directories"

MikaelSmith:
:+1:

spec/classes/puppet_agent_prepare_spec.rb
branan: Added a negative test case for the SSL paths

MikaelSmith:
:+1: waiting on CI.

PR #36, opened 2015-06-12
(PUP-4746, PUP-4735) Fix mco setting fact and add tests

MikaelSmith:
* (PUP-4746) Update mco settings custom fact
    Without this change, mco_*_settings will only return the 1st declaration
    of a setting in the config file. Update the fact to return values from
    all declarations.

* (PUP-4735) Modify acceptance test for mco config
    MCollective config files can contain multiple declarations of the same
    property. Ensure acceptance testing covers that case.

### DIFF
--- a/lib/facter/mco_config.rb
+++ b/lib/facter/mco_config.rb
@@ -16,13 +16,21 @@
       settings = nil
       config = Facter.fact("mco_#{node}_config".to_sym)
       if config and config.value
-        settings = Hash[File.readlines(config.value).select {|v|
+        settings = {}
+
+        File.readlines(config.value).select {|v|
           v.lstrip =~ /[^#].+=.+/
         }.map {|x|
           x.split('=', 2).map {|s| s.strip}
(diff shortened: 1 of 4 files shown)


branan:
I'm not really sure why we expect multiple libdir entries in some cases and joined on colons in others. What's the backstory there?

branan:
the fact appears to always be joining, so I'm not sure how we get to a place where things are unjoined?

MikaelSmith:
I'm not sure either, it's not ideal. Examples like https://github.com/puppetlabs/marionette-collective/blob/e1d952275973368f6b4137d3ec6fd507e545a728/ext/aio/common/client.cfg.dist#L4-L8 can exist, but the inifile tools aren't really prepared to deal with that file.

branan:
@MikaelSmith I asked my question in a stuipd way, but I think I've figured out how we're handling the client and server config files in the code, and your fixtures make sense now :)

PR #37, opened 2015-06-17
(maint) Add docs and assert_private to classes

cyberious:


MikaelSmith:
Thanks for cleaning that up. :+1:

cyberious:
Fixed broken tests that assumed we always executed but will never since they are nested and limited to only 3.8

PR #39, opened 2015-06-22
(FM-2911) Munge cmd path if rubyplatform =~ i386

cyberious:


joshcooper:
I'm confused why this is necessary (and if so seems like a bug in the system32 fact). I'd expect the fact to only return `sysnative` if that directory exists? https://github.com/puppetlabs/facter/blob/2.x/lib/facter/system32.rb#L15-L16 Or does the directory exist, but `cmd.exe` isn't present?

cyberious:
It is not the existence of it that is the problem.  It is that i386 ruby can not execute the command from sysnative. https://tickets.puppetlabs.com/browse/FM-2911

ferventcoder:
It appears this is the issue:

```sh
C:\Windows\sysnative\cmd.exe /c start /b C:\Windows\sysnative\cmd.exe /c
"C:\Users\ADMINI~1\AppData\Local\Temp\install_puppet.bat" returned 1 instead of
one of [0]
```

manifests/windows/install.pp
```diff
@@ -32,7 +37,7 @@
     content => template('puppet_agent/install_puppet.bat.erb')
   }->
   exec { 'install_puppet.bat':
-    command   => "${::system32}\\cmd.exe /c start /b ${::system32}\\cmd.exe /c \"${env_temp_variable}\\install_puppet.bat\"",
+    command   => "${_cmd_location} /c start /b ${_cmd_location} /c \"${env_temp_variable}\\install_puppet.bat\"",
```
ferventcoder: The first one should still use system32

```puppet
command   => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${env_temp_variable}\\install_puppet.bat\"",
```

cyberious: But it does not.  On PE 3.8 i386 it resolves as sysnative.  I opened a IRB session to verify that it said file exists and that cmd.exe was executable

ferventcoder: The call to `::system32` will resolve to `c:\windows\sysnative` and `c:\windows\sysnative\cmd.exe` will resolve just fine. However the rest of the args to the right `/c start /b $(_cmd_location)` are handed to the 64-bit cmd.exe process you just started up, one of which includes `c:\windows\sysnative\cmd.exe` (in the old code).

And then it states it can't find the file, but the error is from Cmd.exe, not from Puppet.

cyberious: I understand that it will work in that manner however the code reading is more confusing, it is more simplistic and with ${_cmd_location} in both places

spec/classes/puppet_agent_windows_install_spec.rb
```diff
@@ -129,6 +129,38 @@
...
+                             'command' => 'C:\windows\sysnative\cmd.exe /c start /b C:\windows\system32\cmd.exe /c "C:\tmp\install_puppet.bat"',
+                           } }
+
+          }
+        end
+        describe 'x86' do
+          let(:facts) { facts.merge({:rubyplatform => 'x86_64'}) }
+          it {
+            is_expected.to contain_exec('install_puppet.bat').with { {
+                             'command' => 'C:\windows\sysnative\cmd.exe /c start /b C:\windows\sysnative\cmd.exe /c "C:\tmp\install_puppet.bat"',
```
ferventcoder: This is technically incorrect still. I think this will fail testing.

ferventcoder: If it passes, then it is still wrong.

The second call to cmd.exe is always `c:\Windows\system32\cmd.exe`. It's probably fine to just use that instead of a private variable.

manifests/windows/install.pp
```diff
@@ -23,6 +23,11 @@
     }
   }
 
+  $_cmd_location = $::rubyplatform ? {
+    /i386/  => "C:\\Windows\\system32\\cmd.exe",
+    default => "${::system32}\\cmd.exe"
+  }
```
ferventcoder: Either way, this should always come out to `C:\Windows\system32\cmd.exe` - this can be simplified by removing it.

ferventcoder:
@cyberious @cowofevil I don't know if this is fixed yet.

cowofevil:
@ferventcoder Passed my initial testing, I'm about to retest the FM-2911 ticket and will post results.

cowofevil:
@cyberious @ferventcoder With build 0.1.0-b20020-6aef09ef the issue no longer exists, so from a QA perspective this looks fixed. Do we need to revisit this PR?

ferventcoder:
Just the test itself.

PR #41, opened 2015-06-23
(PUP-4478) Fix tests on Ubuntu

MikaelSmith:
### (PUP-4478) Fix apt refreshes on Ubuntu
The module requires that apt update before installing the package.
Class level dependencies don't seem to enforce that, so add an explicit
sentinel to force the apt_update exec to run before finishing the repo
setup.
### (maint) Skip testing service start on agent runs
In some cases having the `puppet` service running when trying to run
`puppet agent` manually to verify idempotency results in an error, because
the service has also started an agent run at the same time.

`puppet apply` tests that the service is started after upgrade. Disable
starting the `puppet` service during agent runs to avoid the race
condition.

### DIFF
--- a/manifests/osfamily/debian.pp
+++ b/manifests/osfamily/debian.pp
@@ -7,6 +7,14 @@ class puppet_agent::osfamily::debian {
     key        => {
       'id'     => '47B320EB4C7C375AA9DAE1A01054B7A24BD6EC30',
       'server' => 'pgp.mit.edu',
-    }
+    },
+    notify     => Notify['pc1_repo_force'],
+  }
+
+  # apt_update doesn't inherit the future class dependency, so it
+  # can wait until the end of the run to exec. Force it to happen now.
+  notify { 'pc1_repo_force':
+      message => 'forcing apt update for pc1_repo',
+      require => Exec['apt_update'],
   }
 }
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -72,7 +72,7 @@ describe 'puppet_agent class' do
   context 'agent run' do
     before(:all) {
       setup_puppet_on default, :agent => true
-      pp = "file { '#{master.puppet['codedir']}/environments/production/manifests/site.pp': ensure => file, content => 'class { \"puppet_agent\": }' }"
+      pp = "file { '#{master.puppet['codedir']}/environments/production/manifests/site.pp': ensure => file, content => 'class { \"puppet_agent\": service_names => [\"mcollective\"] }' }"
       apply_manifest_on(master, pp, :catch_failures => true)
     }
     after (:all) {
@@ -95,11 +95,6 @@ describe 'puppet_agent class' do
       it { is_expected.to be_installed }
     end
 
-    describe service('puppet') do
-      it { is_expected.to be_enabled }
-      it { is_expected.to be_running }
-    end
-
     describe service('mcollective') do
       it { is_expected.to be_enabled }
       it { is_expected.to be_running }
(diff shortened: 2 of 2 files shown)


manifests/osfamily/debian.pp
```diff
@@ -7,6 +7,14 @@
...
       'id'     => '47B320EB4C7C375AA9DAE1A01054B7A24BD6EC30',
       'server' => 'pgp.mit.edu',
-    }
+    },
+    notify     => Notify['pc1_repo_force'],
+  }
+
+  # apt_update doesn't inherit the future class dependency, so it
+  # can wait until the end of the run to exec. Force it to happen now.
+  notify { 'pc1_repo_force':
```
MikaelSmith: @joshcooper does this sound like an issue we know about?

joshcooper: good question, and not sure... @haus?

haus: I don't understand the question.

MikaelSmith: Not really a packaging question. `apt::source` introduces a notify of `Exec['apt_update']`. We use contain on this class from `puppet_agent::prepare`, and `puppet_agent` class should setup a dependency of the `puppet_agent::install` class on the `puppet_agent::prepare` class.

However, the Exec wasn't happening until after the install class.

MikaelSmith: `include apt` is what introduces the `Exec['apt_update']` resource; maybe that should be `contain apt` instead.

manifests/osfamily/debian.pp
```diff
@@ -7,6 +7,14 @@
     key        => {
       'id'     => '47B320EB4C7C375AA9DAE1A01054B7A24BD6EC30',
       'server' => 'pgp.mit.edu',
-    }
+    },
+    notify     => Notify['pc1_repo_force'],
```
joshcooper: I didn't follow why we need the sentinel? The puppetlabs-apt docs say "By default, Puppet runs apt-get update on the first Puppet run after you include the apt class, and anytime notify => Exec['apt_update'] occurs", could we just have the `apt::source` resource `notify => Exec['apt_update']`?

MikaelSmith: In at least one instance, the `Exec['apt_update']` was the last resource evaluated, after trying to install packages from PC1. `notify => Exec['apt_update']` wasn't enough to pull the priority forward.

MikaelSmith: The failure at http://jenkins-modules-dev.delivery.puppetlabs.net/job/modules_puppetlabs-puppet_agent_intn-sys_core-master-foss-stable/PLATFORM_CONFIG=ubuntu-1404-64mda,SLAVE_LABEL=beaker/8/console is what I'm fixing here.

manifests/osfamily/debian.pp
joshcooper: with this PR I see the refresh before we try to install puppet-agent (as required):

```
Notice: /Stage[main]/Apt::Update/Exec[apt_update]: Triggered 'refresh' from 1 events
Notice: forcing apt update for pc1_repo
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Notify[pc1_repo_force]/message: defined 'message' as 'forcing apt update for pc1_repo'
```

It's interesting that the sentinel is required. It's as though the refresh is not contained within this class, but adding a sentinel forces it. Seems like it might be a puppet bug?

MikaelSmith: @hunner @mhaskel have either of you seen this behavior before?

mhaskel: @MikaelSmith I'm assuming you're still on a 1.x puppetlabs-apt version? The use of anchors in those releases were ... not correct. It should be fixed in the 2.x series.

MikaelSmith: The metadata specifies puppetlabs-apt 2.0.1. I'm not exactly sure what acceptance is installing though, I'll double check.

mhaskel: o_O I thought we had fixed containment

PR #43, opened 2015-06-30
(maint) Update to have version requirements in metadata.json

cyberious:


MikaelSmith:
Technically it will run on 4.0.0, it just won't do anything.

metadata.json
```diff
@@ -40,6 +40,12 @@
       ]
     }
   ],
+  "requirements": [
+    {
+      "name": "puppet",
+      "version_requirement": ">= 3.8.0 < 4.0.0"
```
ferventcoder: Does this version requirement apply to the master that I would install this on to later upgrade my agents?

ferventcoder: Because I'd just state `>= 3.8.0` if so.

MikaelSmith: Also, does it have an impact on acceptance testing, where the tests run `puppet agent` after upgrading?

MikaelSmith:
Why is the forge claiming we have 7 warnings, when the lint run here doesn't say any?

MikaelSmith:
ping @cyberious

MikaelSmith:
Closed in favor of #47.

PR #44, opened 2015-07-01
(PE-10132) Add PE support for SLES 12, Debian, RedHat platforms

ericwilliamson:
This PR adds support for upgrading PE agents from 3.8 to 4.0.

Changes include:
* Passing the `source` parameter through to RedHat and Debian platforms
* Using the PE Masters repositories instead of FOSS for Debian and RedHat

TODO:
Update Windows source
Add support for following platforms:
SLES 10-11 (Zypper)
OSX

### DIFF
--- a/manifests/osfamily/suse.pp
+++ b/manifests/osfamily/suse.pp
@@ -1,0 +1,66 @@
+class puppet_agent::osfamily::suse {
+  if $::operatingsystem != 'SLES' or $::puppet_agent::is_pe == false {
+    fail("${::operatingsystem} not supported")
+  }
+
+  case $::operatingsystemmajrelease {
+    '12': {
+      # Import the GPG key
+      $keyname = 'RPM-GPG-KEY-puppetlabs'
+      $gpg_path = "/etc/pki/rpm-gpg/${keyname}"
+
+      file { ['/etc/pki', '/etc/pki/rpm-gpg']:
+        ensure => directory,
+      }
+
+      file { $gpg_path:
+        ensure => present,
+        owner  => 0,
+        group  => 0,
+        mode   => '0644',
+        source => "puppet:///modules/puppet_agent/${keyname}",
+      }
+
+      # Given the path to a key, see if it is imported, if not, import it
+      exec {  "import-${keyname}":
+        path      => '/bin:/usr/bin:/sbin:/usr/sbin',
+        command   => "rpm --import ${gpg_path}",
+        unless    => "rpm -q gpg-pubkey-`echo $(gpg --throw-keyids < ${gpg_path}) | cut --characters=11-18 | tr [A-Z] [a-z]`",
+        require   => File[$gpg_path],
+        logoutput => 'on_failure',
+      }
+
+
+      # Set up a zypper repository by creating a .repo file which mimics a ini file
+      $pe_server_version = pe_build_version()
+      $source = "${::puppet_agent::source}/${pe_server_version}/${::platform_tag}"
+
+      $repo_file = '/etc/zypp/repos.d/pc1_repo.repo'
+      $repo_name = 'pc1_repo'
+
+      # In Puppet Enterprise, agent packages are served by the same server
+      # as the master, which can be using either a self signed CA, or an external CA.
+      # Zypper has issues with validating a self signed CA, so for now disable ssl verification.
+      $repo_settings = {
+        'name'        => $repo_name,
+        'enabled'     => '1',
+        'autorefresh' => '0',
+        'baseurl'     => "${source}?ssl_verify=no",
+        'type'        => 'rpm-md',
+      }
+
+      $repo_settings.each |String $setting, String $value| {
+        ini_setting { "zypper ${repo_name} ${setting}":
+          ensure  => present,
+          path    => $repo_file,
+          section => $repo_name,
+          setting => $setting,
+          value   => $value,
+        }
+      }
+    }
+    default: {
+      fail("${::operatingsystem} ${::operatingsystemmajrelease} not supported")
+    }
+  }
+}
--- a/lib/puppet/parser/functions/uri_host_from_string.rb
+++ b/lib/puppet/parser/functions/uri_host_from_string.rb
@@ -1,0 +1,14 @@
+require 'uri'
+
+module Puppet::Parser::Functions
+  newfunction(:uri_host_from_string, :arity => 1, :type => :rvalue, :doc => <<-EOS
+  Return a uri host from a string
+  EOS
+  ) do |args|
+
+    uri = URI(args[0])
+
+    return uri.host
+  end
+end
+
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,6 +15,7 @@
 #
 class puppet_agent (
   $arch          = $::architecture,
+  $is_pe         = $::puppet_agent::params::_is_pe,
   $package_name  = $::puppet_agent::params::package_name,
   $service_names = $::puppet_agent::params::service_names,
   $source        = $::puppet_agent::params::_source,
--- a/manifests/osfamily/debian.pp
+++ b/manifests/osfamily/debian.pp
@@ -6,11 +6,30 @@ class puppet_agent::osfamily::debian {
   if $::puppet_agent::is_pe {
     $pe_server_version = pe_build_version()
     $source = "${::puppet_agent::source}/${pe_server_version}/${::platform_tag}"
-    $source_host = uri_host_from_string($source)
 
-    # If this is PE, we're using a self signed certificate, so need to disable SSL verification
+    # In Puppet Enterprise, agent packages are served by the same server
+    # as the master, which can be using either a self signed CA, or an external CA.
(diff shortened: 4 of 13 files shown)


ericwilliamson:
Ping @MikaelSmith @kylog @jpartlow Now that all the pe_repo changes are finished, I wanted to go ahead and get this PR review rolling for the simple platforms while I work on adding spec tests and the PE specific platforms.

A couple things I don't like with the current approach:

1) Using the fact `$::is_pe`. This fact is only valid for 3.8 and earlier agents. Once we go to upgrade 4.0 agents it will need a new detection method. I'm thinking maybe a check for if a PE specific function exists?

2) Having to disable SSL verification due to self signed CA certs. It *should* be safe in PE to use the 3.8 agent cert paths because they don't change in PE, but wasn't sure if there was an edge case I might be missing?

MikaelSmith:
Well it matter in 4.0+ whether the master is PE?

MikaelSmith:
On the self-signed cert, we can't bake that into the repo setup?

ericwilliamson:
> On the self-signed cert, we can't bake that into the repo setup?

We can - I'm just wondering if there is going to be any issues with cert paths changing between 3.8 and 4.0. For PE I don't believe there are any changes.

/cc @jpartlow

jpartlow:
I think we still need to determine whether or not the agent is in a PE environment 4.0+, because it effects whether or not our source is going to be from the FOSS servers, or from a local PE repository.  Unless we say that the user has to specify that themselves when they use the agent_upgrade module in PE?

ericwilliamson:
@MikaelSmith @jpartlow PR has been updated and is ready for another round of feedback

manifests/osfamily/suse.pp
```diff
@@ -0,0 +1,66 @@
+class puppet_agent::osfamily::suse {
+  if $::operatingsystem != 'SLES' and ! $::puppet_agent::is_pe {
+    fail("${::operatingsystem} not supported")
+  }
+
+  case $::operatingsystemmajrelease {
+    '12': {
```
jpartlow: Sles 11 and Sles 10 are works in progress?  (another PR?)

jpartlow:
Everything here looks great to me @ericwilliamson.  I had a question above about the state of pesky Sles10/11.

manifests/osfamily/suse.pp
ericwilliamson: Correct. I had SLES 11 in here but after further testing, I was running into issues with upgrades: 

```
Error: Execution of '/usr/bin/zypper --quiet install --auto-agree-with-licenses --no-confirm --name puppet-agent' returned 4: The following NEW package is going to be installed:
  puppet-agent
 
The following packages are going to be REMOVED:
  pe-agent pe-augeas pe-facter pe-hiera pe-libldap pe-libyaml pe-mcollective pe-mcollective-common pe-openssl pe-puppet pe-puppet-enterprise-release pe-ruby pe-ruby-augeas pe-ruby-ldap pe-ruby-rgen pe-ruby-shadow pe-ruby-stomp pe-rubygem-deep-merge pe-rubygem-net-ssh pe-virt-what
 
The following package is not supported by its vendor:
  puppet-agent
 
1 new package to install, 20 to remove.
Overall download size: 17.9 MiB. After the operation, additional 33.0 MiB will be used.
Continue? [y/n/?] (y): y
Installation of puppet-agent-1.2.0-1.sles11 failed:
(with --nodeps --force) Error: Subprocess failed. Error: RPM failed: warning: /etc/puppetlabs/mcollective/server.cfg created as /etc/puppetlabs/mcollective/server.cfg.rpmnew
warning: /etc/puppetlabs/puppet/puppet.conf created as /etc/puppetlabs/puppet/puppet.conf.rpmnew
warning: /etc/sysconfig/puppet created as /etc/sysconfig/puppet.rpmnew
insserv: script puppet: service puppet already provided!
insserv: script mcollective: service mcollective already provided!
insserv: Service network is missed in the runlevels 4 to use service puppet
insserv: Service syslog is missed in the runlevels 4 to use service puppet
insserv: exiting now!
/sbin/insserv failed, exit code 1
puppet                    0:off  1:off  2:off  3:off  4:off  5:off  6:off
insserv: script mcollective: service mcollective already provided!
insserv: script puppet: service puppet already provided!
insserv: Service network is missed in the runlevels 4 to use service puppet
insserv: Service syslog is missed in the runlevels 4 to use service puppet
insserv: exiting now!
/sbin/insserv failed, exit code 1
mcollective               0:off  1:off  2:off  3:off  4:off  5:off  6:off
error: %post(puppet-agent-1.2.0-1.sles11.x86_64) scriptlet failed, exit status 1
 
 
Abort, retry, ignore? [a/r/i] (a): a
Problem occured during or after installation or removal of packages:
Installation aborted by user
```

Was going to ping RE about it on monday

jpartlow:
Should we get this in and return to the todo list in another PR?

MikaelSmith:
Sounds good to me.

ericwilliamson:
Updated title. Working on adding spec tests then.

MikaelSmith:
Planning to add spec tests as part of this, or another commit?

ericwilliamson:
I'd prefer this PR, but I know time is tight so will let @jpartlow make that call. I'm working on them right now while testing out my acceptance test changes for this in the PE suite.

jpartlow:
@ericwilliamson I would finish specs in this PR; will save time in long run to see failures now.

ericwilliamson:
@MikaelSmith Pushed up specs for suse. I can't really add any for redhat or debian as I couldn't figure out how to mock `$::puppet_agent::{source,is_pe}`. Any suggestions?

MikaelSmith:
You could set the `is_pe` fact in any of the `:facts` initialization.

ericwilliamson:
@MichalBryxi I tried that and it didn't work.

Heres the patch i'm trying locally (Note the fix for the logic error). The output is the same if i set the default `is_pe` fact to true or false

```
diff --git a/manifests/osfamily/suse.pp b/manifests/osfamily/suse.pp
index 09c8604..3f6c69b 100644
--- a/manifests/osfamily/suse.pp
+++ b/manifests/osfamily/suse.pp
@@ -1,5 +1,5 @@
 class puppet_agent::osfamily::suse {
-  if $::operatingsystem != 'SLES' and ! $::puppet_agent::is_pe {
+  if $::operatingsystem != 'SLES' or $::puppet_agent::is_pe == false {
     fail("${::operatingsystem} not supported")
   }

diff --git a/spec/classes/puppet_agent_osfamily_suse_spec.rb b/spec/classes/puppet_agent_osfamily_suse_spec.rb
index bad057d..06ab8de 100644
--- a/spec/classes/puppet_agent_osfamily_suse_spec.rb
+++ b/spec/classes/puppet_agent_osfamily_suse_spec.rb
@@ -4,6 +4,7 @@ describe 'puppet_agent::osfamily::suse' do
   facts = {
     :osfamily        => 'Suse',
     :operatingsystem => 'SLES',
+    :is_pe           => false,
   }

   context "with a not supported operating system" do
@@ -20,6 +21,16 @@ describe 'puppet_agent::osfamily::suse' do
     end
   end

+  context "with a foss environment" do
+    let(:facts) do
+      facts.merge({
+        :is_pe => false
+      })
+    end
+
+    it { expect { catalogue }.to raise_error(Puppet::Error, /SLES not supported/) }
+  end
+
   context "with a supported operating system" do
     ['12'].each do |os_version|
       context "with SLES #{os_version}" do
```

and the output:
```
Failures:

  1) puppet_agent::osfamily::suse with a foss environment should raise Puppet::Error with message matching /SLES not supported/
     Failure/Error: it { expect { catalogue }.to raise_error(Puppet::Error, /SLES not supported/) }
       expected Puppet::Error with message matching /SLES not supported/, got #<Puppet::PreformattedError: Evaluation Error: Error while evaluating a Function Call, SLES  not supported at /s/modules/puppetlabs-puppet_agent/spec/fixtures/modules/puppet_agent/manifests/osfamily/suse.pp:63:7 on node localhost.delivery.puppetlabs.net> with backtrace:
         # ./spec/classes/puppet_agent_osfamily_suse_spec.rb:31:in `block (4 levels) in <top (required)>'
         # ./spec/classes/puppet_agent_osfamily_suse_spec.rb:31:in `block (3 levels) in <top (required)>'
     # ./spec/classes/puppet_agent_osfamily_suse_spec.rb:31:in `block (3 levels) in <top (required)>'
```

MikaelSmith:
When you use `describe "puppet_agent::osfamily::suse"`, I'm not sure how params gets initialized. @cyberious might have a thought; I recall him changing all the subclass tests to instantiate the full puppet_agent class.

ericwilliamson:
@MikaelSmith I should also probably update https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/metadata.json#L10 for sles and windows?

MikaelSmith:
That probably makes sense. I'm not sure why Windows wasn't updated before @cyberious.

MikaelSmith:
All ready to merge?

ericwilliamson:
Yep

PR #48, opened 2015-07-12
(PUP-4849) Make puppet_agent ensure platform type in tests

MikaelSmith:


spec/spec_helper_acceptance.rb
```diff
@@ -164,7 +163,6 @@ def teardown_puppet_on(host)
 package { ['puppet-agent', 'puppet', 'mcollective', 'mcollective-client']: ensure => purged }
   EOS
   on host, puppet('apply', '-e', "\"#{pp}\"")
-  remove_defaults_on host
```
branan: Why ditch the `remove_defaults_on`?

MikaelSmith: I should debug more, but it just caused everything to fail when the config file set `type: aio`.

MikaelSmith: Filing a BKR ticket about this, the remove_defaults_on code can't be called twice with `type: aio` set in the config file.

branan: OK. Since there's a beaker ticket can you up date this with a quick comment to that effect (or move the reomoval to its own commit that mentions the ticket) so we have some info in this repo about what we're blocked on?

Once that's done I'll be happy to click the button here

MikaelSmith: I think it turned out this whole fix is only needed due to BKR-395; install_puppet_on sets the type to `foss` all on its own. I'll update the commit.

MikaelSmith:
Updated

PR #51, opened 2015-07-16
(PUP-4886) Clear settings at global level

MikaelSmith:
Without this change, settings at the global level in puppet.conf
(outside a section) would not be cleared, and get inheritted by each
section. This would lead to incorrect configurations.

Also clear settings at the global level.


branan:
I didn't even know we allowed configuration before a section header *sigh*

branan:
@MikaelSmith should there be new unit or acceptance tests for this?

MikaelSmith:
Spec tests were updated. I could add something to the acceptance tests.

branan:
Nope, I just need more coffee :).

PR #54, opened 2015-07-21
(maint) Add Windows support to metadata

MikaelSmith:


MikaelSmith:
@cyberious look fair?

PR #56, opened 2015-07-22
Remove classfile puppet.conf setting, update for PE windows

ericwilliamson:
This PR contains two fixes for PE users:

The first is reseting the `classfile` puppet.conf setting for mco inventory. The second is specifying the agent version to that of the compiling master instead of latest.

### DIFF
--- a/manifests/windows/install.pp
+++ b/manifests/windows/install.pp
@@ -12,8 +12,17 @@ class puppet_agent::windows::install {
     default   => $::puppet_agent::arch
   }
 
+  if $::puppet_agent::is_pe {
+    $_agent_version = chomp(file('/opt/puppetlabs/puppet/VERSION'))
+    $_pe_server_version = pe_build_version()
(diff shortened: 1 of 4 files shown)


ericwilliamson:
Ping @cyberious @jpartlow

MikaelSmith:
:+1:

manifests/windows/install.pp
```diff
@@ -12,8 +12,17 @@
     default   => $::puppet_agent::arch
   }
 
+  if $::puppet_agent::is_pe {
```
ferventcoder: So crazy question, once someone is on the AIO agent and uses this for the next upgrade...

jpartlow: @ferventcoder I think this line is handling that, albeit in a somewhat roundabout way: https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/params.pp#L13

MikaelSmith: Also not explicitly supported, we'll have to verify when we enable running on Puppet 4 agents.

ferventcoder: :+1:

PR #59, opened 2015-07-24
(PUP-4921) Add PE version restriction

MikaelSmith:


jpartlow:
@MikaelSmith I'm not familiar with the PE version in module metadata.  What does that end up being compared against?

The supported path for PE ugprades is 3.8.1 -> 2015.2.0.

I'm also not sure what 5.0.0 would correspond to in PE.

MikaelSmith:
I'm following a pattern in puppetlabs-apt that says < 5.0. Maybe @mhaskel or @adreyer is better able to answer.

Should I bump the minimum to 3.8.1?

jpartlow:
@MikaelSmith based on our conversation in the modules channel; this is so the forge can properly filter the module's availability for particular versions of PE, and the requirements look right to me.  Merging.

PR #62, opened 2015-09-04
WIP (PUP-5222) Add upgrading the puppet-agent package in Puppet 4+

MikaelSmith:
Adds support for upgrading the puppet-agent package after install.
Upgrades of Puppet 4+ agents will only occur if a specific version is
specified.

Also adds a collection parameter to specify which Puppet Collection to
use.

Still needs some acceptance testing, and spec tests for Puppet 4+ behavior.


highb:
Closing in favor of https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/78

PR #64, opened 2015-09-16
service_names param fix in README

applewiskey:
Attribute name in documentation mismatch real attribute name.


MikaelSmith:
Merged in https://github.com/puppetlabs/puppetlabs-puppet_agent/commit/6b403adf8e96f054affae590372b6831137081e8.

PR #63, opened 2015-09-15
(PE-10915) Add SLES 10 upgrade for PE

jpartlow:
Introduces support for upgrading SLES 10 agents in PE. SLES 10 does not
have tools for securely handling remote package management.

* puppet-agent package is downloaded from a pe_packages fileserver mount
prepared separately on the master (in puppet_enterprise module)
* the package is downloaded with a File resource, but only if agent node
is not already at the current aio version
* because of limitations in the rpm package provider, it is not feasible
to get the install of the new puppet-agent package to run with '-U
--replacepkgs' flags that would allow rpm to correctly uninstall
conflicting superseded pe-* packages, so we remove those prior to
install. (Passing '-U --replacepkgs' in install_options parameter ends
up failing:

Error: Execution of '/bin/rpm -i "-U --replacepkgs"
/opt/puppetlabs/packages/puppet-agent-1.2.5-1.sles10.x86_64.rpm'
returned 1: -U --replacepkgs: not an rpm package (or package manifest)

* I'm threading package_file_name parameter through because it varies by
platform and ends up being needed both in the prepare and the install
phase for these special case platforms.

PR #65, opened 2015-09-21
(PE-12001) Add solaris 10

jpartlow:
Based on #63.  This should not be merged until that PR is.

### DIFF
--- a/manifests/osfamily/solaris.pp
+++ b/manifests/osfamily/solaris.pp
@@ -1,0 +1,39 @@
+class puppet_agent::osfamily::solaris(
+  $package_file_name = undef,
+) {
+  assert_private()
+
+  if $::operatingsystem != 'Solaris' or $::puppet_agent::is_pe == false {
+    fail("${::operatingsystem} not supported")
+  }
+
+  case $::operatingsystemmajrelease {
+    '10': {
+      class { 'puppet_agent::prepare::package':
+        package_file_name => $package_file_name,
+      }
+      contain puppet_agent::prepare::package
+
+      $_unzipped_package_name = regsubst($package_file_name, '\.gz$', '')
+      exec { "unzip ${package_file_name}":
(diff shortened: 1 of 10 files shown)


manifests/osfamily/solaris.pp
```diff
@@ -0,0 +1,37 @@
...
+      class { puppet_agent::prepare::package:
+        package_file_name => $package_file_name,
+      }
+      contain puppet_agent::prepare::package
+
+      $_unzipped_package_name = regsubst($package_file_name, '\.gz$', '')
+      exec { "unzip ${package_file_name}":
+        path      => '/bin:/usr/bin:/sbin:/usr/sbin',
+        command   => "gzip -d /opt/puppetlabs/packages/${package_file_name}",
+        creates   => "/opt/puppetlabs/ppackages/${_unzipped_package_name}",
```
highb: Is `ppuppet` a typo? I assume this is fixed in your local changes.

jpartlow: That's a typo; I'll fix that, and rebase for a change to sles10 pr

jpartlow:
Ah, right, I just figured out why the version constraints are on the osfamily classes; because the osfamily tests will see empty catalogs if the Puppet version is not 3.8.  I'll fix this up after arch.

jpartlow:
Had to add two resources to stop the pe-puppet and pe-mcollective.  It's also necessary that they (or at least the pe-puppet service stop) happen at the very end or the run is interrupted.  There may be better ways to do this then what I currently have arranged.

jpartlow:
Manually tested an upgrade of Solaris10 3.8.2 -> 2015.2.x with this patch.

highb:
Verified upgrade from 3.8.2->2015.2.1, Sol10 agent from vmpool upgraded sucessfully, `mco ping` succeeds after upgrade and puppet runs complete.

PR #66, opened 2015-09-25
(PE-10956) windows upgrade

highb:
Adds a PE-hosted MSI option for upgrading pre-2015 agents.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,11 +69,18 @@ class puppet_agent (
     } ->
     class { '::puppet_agent::install':
       package_file_name => $_package_file_name,
-    } ->
-    class { '::puppet_agent::service': }
+    }
 
     contain '::puppet_agent::prepare'
     contain '::puppet_agent::install'
-    contain '::puppet_agent::service'
+
+    # On windows, our MSI handles the services
+    if $::osfamily != 'windows' {
+      class { '::puppet_agent::service':
+        require => Class['::puppet_agent::install'],
+      }
+      contain '::puppet_agent::service'
+    }
+
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,40 +33,38 @@ class puppet_agent (
     if $::architecture == 'x86' and $arch == 'x64' {
       fail('Unable to install x64 on a x86 system')
     }
-    if $::osfamily == 'windows' {
-      class { '::puppet_agent::prepare': } ->
-      class { '::puppet_agent::windows::install': }
-    }
-    else {
 
-      if $::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10' {
-        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.sles10.${::architecture}.rpm"
-      } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '10' {
-        if $arch =~ '^sun4[uv]$' {
-          $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.sparc.pkg.gz"
-        } else {
-          $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.i386.pkg.gz"
-        }
-      } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ '10\.[9,10,11]' {
-        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
-      } elsif $::operatingsystem == 'aix' and $::architecture =~ 'PowerPC_POWER[5,6,7]' {
-        $aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
-        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.aix${aix_ver_number}.ppc.rpm"
+    if $::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10' {
+      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.sles10.${::architecture}.rpm"
+    } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '10' {
+      if $arch =~ '^sun4[uv]$' {
+        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.sparc.pkg.gz"
       } else {
-        $_package_file_name = undef
+        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.i386.pkg.gz"
       }
-
-      class { '::puppet_agent::prepare':
-        package_file_name => $_package_file_name,
-      } ->
-      class { '::puppet_agent::install':
-        package_file_name => $_package_file_name,
-      } ->
-      class { '::puppet_agent::service': }
-
-      contain '::puppet_agent::prepare'
-      contain '::puppet_agent::install'
-      contain '::puppet_agent::service'
+    } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ '10\.[9,10,11]' {
+      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
+    } elsif $::operatingsystem == 'aix' and $::architecture =~ 'PowerPC_POWER[5,6,7]' {
+      $aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
+      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.aix${aix_ver_number}.ppc.rpm"
+    } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ '10\.[9,10,11]' {
+      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
+    } elsif $::osfamily == 'windows' {
+      $_package_file_name = "${package_name}-${arch}.msi"
+    } else {
+      $_package_file_name = undef
     }
+
+    class { '::puppet_agent::prepare':
+      package_file_name => $_package_file_name,
+    } ->
+    class { '::puppet_agent::install':
+      package_file_name => $_package_file_name,
+    } ->
+    class { '::puppet_agent::service': }
+
+    contain '::puppet_agent::prepare'
+    contain '::puppet_agent::install'
+    contain '::puppet_agent::service'
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -50,7 +50,16 @@ class puppet_agent (
     } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ '10\.[9,10,11]' {
       $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
     } elsif $::osfamily == 'windows' {
-      $_package_file_name = "${package_name}-${arch}.msi"
+      $_arch = $::kernelmajversion ?{
+        /^5\.\d+/ => 'x86', # x64 is never allowed on windows 2003
+        default   => $arch
+      }
+
+      if $is_pe {
+        $_package_file_name = "${package_name}-${_arch}.msi"
+      } else {
+        $_package_file_name = "${package_name}-${_arch}-latest.msi"
+      }
     } else {
       $_package_file_name = undef
     }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -38,14 +38,23 @@ class puppet_agent::params {
       $mcodirs = [$mco_dir]
 
       $path_separator = ':'
+
+      $user  = 0
+      $group = 0
     }
     'windows' : {
+      $package_name = 'puppet-agent'
+      $service_names = ['puppet', 'mcollective']
+
       $confdir = $::puppet_confdir
       $mco_dir = $::mco_confdir
 
       $mcodirs = [$mco_dir] # Directories should already exists as they have not changed
       $puppetdirs = [regsubst($confdir,'\/etc\/','/code/')]
       $path_separator = ';'
+
+      $user  = 'S-1-5-32-544'
+      $group = 'S-1-5-32-544'
     }
     default: {
       fail("${::operatingsystem} not supported")
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -26,6 +26,8 @@ class puppet_agent::params {
       $package_name = 'puppet-agent'
       $service_names = ['puppet', 'mcollective']
 
+      $local_packages_dir = "/opt/puppetlabs/packages"
+
       $confdir = '/etc/puppetlabs/puppet'
       $mco_dir = '/etc/puppetlabs/mcollective'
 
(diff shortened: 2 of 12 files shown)


highb:
Spec tests need an update.

manifests/prepare/package.pp
```diff
@@ -0,0 +1,43 @@
...
+    if $::osfamily == 'windows' {
+      $tag = $::puppet_agent::arch ? {
+        'x64' => 'windows-x86_64',
+      }
+      $source = "puppet:///pe_packages/${pe_server_version}/${tag}/${package_file_name}"
+      $dest = "C:\\Program Files\\Puppet Labs\\Puppet Enterprise\\packages"
+      $file_dest = "C:\\Program Files\\Puppet Labs\\Puppet Enterprise\\packages\\${package_file_name}"
+    } else {
+      $source = "puppet:///pe_packages/${pe_server_version}/${::platform_tag}/${package_file_name}"
+      $dest = "/opt/puppetlabs/packages"
```
jpartlow: Ya, this was on my list of things to clean up.  @highb what do you think about pulling this back into params, and then either plumbing it through alongside $package_file_name or referencing it as something like $pupppet_agent::params::local_packages_dir.  Either way, we can then replace all the /opt/puppetlabs/packages references.

And then I think the reference to puppet_agent::prepare::package::file_dest in install.pp just becomes this variable + package_file_name.

highb: @jpartlow Yeah I think that makes more sense.

highb:
I was going crazy about this-- `be rake spec` locally showed all of the specs passing, so I couldn't figure out what was going on. Looking at the actual Travis failure... 
``` be rake test
---> syntax:manifests
---> syntax:templates
---> syntax:hiera:yaml
manifests/osfamily/windows.pp - WARNING: unquoted resource title on line 7
manifests/params.pp - WARNING: double quoted string containing no variables on line 29
manifests/params.pp - WARNING: double quoted string containing no variables on line 51```

I didn't even know there was a `rake test` Fixing the WARNINGs now.

**Comment by highb** [comment]
OK, latest push should resolve the linter errors and this should finally be good to review and/or merge?

**Comment by jpartlow** [comment]
Looks like it's having issues with Puppet 3.8.  Most of the specs and the module itself are constrained around the Puppet version.

**Comment by highb** [comment]
@jpartlow Fixed! :)

**Comment by jpartlow** [comment]
I think we just need a test run of a 3.8.3->2015.3.0 upgrade with a windows agent.

**Comment by highb** [comment]
Rebased this commit off the OSX/Solaris changes that just went in.

**Comment by highb** [comment]
Confirmed rebased code upgrades Win2012 agents by
1) Installing 3.8.3 on centos 7
2) Installing 3.8.3 on Win2012
3) Verifying mco/puppet runs work
4) Upgrading to 2015.3 on centos 7 master
5) Installing windows pe_repo
6) Blocking access to pm.puppetlabs.com on windows agent
7) Classifying puppet_agent with puppet_agent class
8) puppet run on windows agent, verified mco ping

**Comment by jpartlow** [comment]
rake test is exiting with linting failures

**Comment by jpartlow** [comment]
@highb there's something wrong with the merge on this branch:

https://github.com/highb/puppetlabs-puppet_agent/blob/highb-master-pe-10956_windows_upgrade/manifests/init.pp#L37

I'm not sure how you were able to verify it?  Did you do a rebase after testing?

**Comment by highb** [comment]
@jpartlow That merge hash isn't in my local copy and when I force push, there are no changes. I'm re-running specs and tests locally to figure out what is happening.

**Comment by highb** [comment]
retest this please

**Comment by highb** [comment]
@jpartlow Also https://github.com/highb/puppetlabs-puppet_agent/blob/highb-master-pe-10956_windows_upgrade/manifests/init.pp#L37 is the wrong branch. I accidentally pushed that. The real branch is: https://github.com/highb/puppetlabs-puppet_agent/blob/master-pe-10956_windows_upgrade/manifests/init.pp

**Comment by highb** [comment]
Oh, I've been running `rake spec` not `rake test`. I'll fix the syntax errors.

**Comment by jpartlow** [comment]
@highb Are your local changes synced with your remote?  I'm still seeing the <<<< in the branch from your remote.

**Comment by highb** [comment]
@jpartlow Yeah, my changes should be synced. Make sure you're not pointing to the other (broken) branch I accidentally pushed `highb-master-pe-10956_windows_upgrade`

**Comment by jpartlow** [comment]
@highb ok, that's sorted out on my end, so that just leaves the lint.

File: manifests/params.pp [hunk 2]
```diff
@@ -38,14 +40,25 @@
...
       $path_separator = ':'
+
+      $user  = 0
+      $group = 0
     }
     'windows' : {
+      $package_name = 'puppet-agent'
+      $service_names = ['puppet', 'mcollective']
+
+      $local_packages_dir = 'C:\\Program Files\\Puppet Labs\\Puppet Enterprise\\packages'
```
joshcooper: This may not be the correct directory in situations where the user has installed puppet elsewhere, or when running 32-bit puppet on 64-bit windows. I think there's an `env_windows_installdir` fact that you can use: https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L8 /cc @Iristyle 

You should be able to use the install_dir fact

highb: Ah, that's helpful.

joshcooper: Another thought is to use `C:\ProgramData\PuppetLabs` instead (I think there's a PE `common_appdata` fact for that). If you write to `C:\Program Files` then those files will prevent puppet from uninstalling cleanly (it will leave behind the new files and any parent directories). It's not a big deal, but not ideal. Also if you are running puppet as "non-root", puppet may not have permission to write to `C:\Program Files`, but it can write to `C:\ProgramData`

highb:
I'm going to run through upgrading a w2012 agent with this patch again.

highb:
Worked on upgrading a 3.8.3 agent to 2015.3, with one error in the run:
```Could not evaluate: Cannot get status of mcollective, error was: The specified service does not exist as an installed service. - OpenService: The specified service does not exist as an installed service. Wrapped exception: The specified service does not exist as an installed service. - OpenService: The specified service does not exist as an installed service.```

But subsequent runs can find mco/pxp-agent just fine and `mco ping`/runonce from the console work fine.

**Comment by highb** [comment]
Tests are failing on puppet 3.8.1. Fun. Looking into what is going on there.

**Comment by highb** [comment]
So the remaining test failure is due to `is_expected.to contain_file('C:\tmp\puppet-agent.msi').that_comes_before('File[C:\tmp\install_puppet.bat]')` seeing the file path as not absolute. I think this is occuring because https://github.com/puppetlabs/puppet/blob/3.8.4/lib/puppet/type/file.rb#L53 calls https://github.com/puppetlabs/puppet/blob/3.8.4/lib/puppet/util.rb#L230 which finally calls https://github.com/puppetlabs/puppet/blob/3.8.4/lib/puppet/util/platform.rb#L10 . I've been trying to stub the platform.windows? method but rspec doesn't seem to be agreeing with me. Should that even be possible? Is there some better way of getting rspec to mock out a windows environment?

File: manifests/prepare/package.pp [hunk 3]
```diff
@@ -17,20 +17,25 @@
   if $puppet_agent::params::master_agent_version != $::aio_agent_version {
     $pe_server_version = pe_build_version()
 
-    $source = "puppet:///pe_packages/${pe_server_version}/${::platform_tag}/${package_file_name}"
+    if $::osfamily == 'windows' {
+      $tag = $::puppet_agent::arch ? {
+        'x64' => 'windows-x86_64',
+      }
+      $source = "puppet:///pe_packages/${pe_server_version}/${tag}/${package_file_name}"
```
mwbutcher: ${tag} seems to need a suffix of the master's agent version number

MikaelSmith:
Want to roll https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/68 into this?

highb:
@MikaelSmith Or if you merge that PR I can rebase. Either way.

manifests/prepare/package.pp
highb: So that is a new behavior that was introduced in 2015.3.2, before `pe_repo` would just copy the agent into the ${tag} directory, now we add the version. On platforms that use `manifests/repo.pp` we use https://github.com/puppetlabs/puppetlabs-pe_repo/blob/2015.3.x/manifests/repo.pp#L104 to setup a symlink to that tag from the old path. Windows doesn't have that symlink.

spec/classes/puppet_agent_windows_install_spec.rb
```diff
@@ -7,8 +7,8 @@
       is_expected.to_not contain_class('::puppet_agent::windows::install')
     }
   elsif Puppet.version >= "3.8.0"
-    {'5.1' => {:expect_arch => 'x86', :appdata => 'C:/Document and Settings/All Users/Application Data/Puppetlabs'},
-     '6.1' => {:expect_arch => 'x64', :appdata => 'C:/ProgramData/Puppetlabs'}}.each do |kernelmajversion, values|
+    {'5.1' => {:expect_arch => 'x86', :appdata => 'C:\Document and Settings\All Users\Application Data'},
```
arafatm: @bh is `File.join` appropriate here?

highb: Probably not, since we're running these tests on Unix systems and `File::SEPARATOR` should be `/` there.

manifests/prepare/package.pp
jpartlow: This is a bug in pe_repo https://tickets.puppetlabs.com/browse/PE-13759 (created by PE-12030)

spec/classes/puppet_agent_windows_install_spec.rb
joshcooper: Fun with ruby and windows... So Ruby defines `File::SEPARATOR` to be `/` on both *nix and Windows, and Windows accepts both forward and backwards slashes. So you should be ok using `File.join`.

If you really want backslashes, you can use `File::ALT_SEPARATOR`. On Windows it's defined to be `\` and on *nix it's `nil`.

My recommendation is to be consistent with whichever you choose. For example, `File.join(`Dir.pwd`, 'bar')` can lead to mixed slashes, e.g. `C:\foo/bar`.

If you need to normalize the paths, you can do `File.expand_path(File.join(Dir.pwd, 'bar')` to produce `C:/foo/bar`

MikaelSmith:
Needs a rebase now.

highb:
Rebased everything, I'll run this through some testing.

highb:
image
Looks like Puppet might be having trouble dealing with the Windows service changing underneath it?

highb:
Aside from the strange service error on the first run with puppet_agent, the upgrade seems to work fine.

highb:
Seems like it might be a race condition with the MSI finishing install because post puppet run I have no problem interacting with the `puppet` and `mcollective` services:
image

MikaelSmith:
Oh, shouldn't be trying to manage the service for Windows installs. The actual install happens with a script scheduled to run after Puppet finishes, and the MSI will ensure the service is started.

manifests/init.pp
```diff
@@ -33,40 +33,47 @@
...
+      $_package_file_name = undef
     }
+
+    class { '::puppet_agent::prepare':
+      package_file_name => $_package_file_name,
+    } ->
+    class { '::puppet_agent::install':
+      package_file_name => $_package_file_name,
+    } ->
+    class { '::puppet_agent::service': }
```
MikaelSmith: This should be skipped or not do anything on Windows.

highb:
Ah, alright. I was starting to dive in to `start /wait` and how that works, but just leaving it all to the MSI is much easier. ;)

highb:
OK, tested this out on a mono 3.8.3->2015.3.2-rc upgrade with the change to not manage the service on windows, and I'm no longer seeing the error, as expected. I'd call this ready for someone to test/review again.

jpartlow:
@highb which Windows platform were you testing?  I'm seeing failures upgrading 2008r2 64.  (I'm updating the ticket with details.)

templates/install_puppet.bat.erb
```diff
@@ -6,7 +6,7 @@ FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=
 echo %_task% | findstr "No tasks are running" >nul
 IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
 
-msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>"
+start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>"
```
Iristyle: For reference, Beaker is always a good place to look to see how we install the Puppet MSI - https://github.com/puppetlabs/beaker/blob/f3d87f2b3463c2c49e1dd2b3e34b028336e7151e/lib/beaker/dsl/install_utils/windows_utils.rb#L18-L39

You've fixed the problem now with the `/wait` ... but I don't see the exit status being returned.

manifests/windows/install.pp
```diff
@@ -45,15 +44,15 @@
...
   notice ("Puppet upgrade log file at ${_logfile}")
   debug ("Installing puppet from ${_msi_location}")
-  file { "${::env_temp_variable}\\install_puppet.bat":
+  file { "${_installbat}":
     ensure  => file,
     content => template('puppet_agent/install_puppet.bat.erb')
   }->
   exec { 'install_puppet.bat':
-    command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${::env_temp_variable}\\install_puppet.bat\" ${::puppet_agent_pid}",
+    command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${_installbat}\" ${::puppet_agent_pid}",
```
Iristyle: I think this is more complicated than necessary - in Beaker, we just shell to {{cmd.exe}} and pass it the batch file.  I also don't see anything being done with the exit codes - `0`, `1641` and `3010` are all success.

https://github.com/puppetlabs/beaker/blob/f3d87f2b3463c2c49e1dd2b3e34b028336e7151e/lib/beaker/dsl/install_utils/windows_utils.rb#L120-L127

jpartlow: @Iristyle Are there any additional complications that come from installing the msi on top of an existing installation?  (In this case a puppet-agent msi replacing a pe-puppet/puppet 3.x msi?).  What we're seeing is that the puppet.conf is being overwritten, so the server setting is lost (I think).

templates/install_puppet.bat.erb
```diff
@@ -6,7 +6,7 @@ FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=
 echo %_task% | findstr "No tasks are running" >nul
 IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
 
-msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>"
+start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
```
highb: D'oh: note to future me, this setting is being read on the master. https://docs.puppetlabs.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#puppet-master-variables So if I want to get the existing value, I should write a custom fact. :/

highb:
Tested with this change https://github.com/puppetlabs/pe_acceptance_tests/pull/836 and it works, but I need to change the use of `settings::server` to a custom fact that reads the server value from the config.

jpartlow:
Is the problem that puppet.conf is being overwritten, or just the server setting is being overwritten?  If just the later, then this fix may be fine.

highb:
@jpartlow As far as I can tell, the MSI just overwrites the specific INI setting. So reading it before upgrading, then passing it in during the MSI run should preserve it. Even if it does mess up other settings, having the server correct should make it much easier to remediate any problems.

As for the `settings::server` problem, I've added the `puppet_master_server` fact to read the `server` value.

mwbutcher:
:+1:   successfully tested against windows2012r2 upgrading from puppet 3.8.5 to 4.3.2

PR #67, opened 2015-09-26
Issue/master/pe 10914 add osx 109 upgrade

jpartlow:
Not pretty, but it does seem to work.

In particular the remove_packages_osx class is ugly and needs review.  Deleting all of /opt/puppet seems dubious.  We might be able to instead use a script to selectively rm files from 'pkgutil --list' output.

Specs need to be finished.

### DIFF
--- a/manifests/install/remove_packages_osx.pp
+++ b/manifests/install/remove_packages_osx.pp
@@ -1,0 +1,60 @@
+# == Class puppet_agent::install::remove_packages_osx
+#
+# Sadly, special handling is required to clear up puppet_enterprise installation
+# on 3.8.
+#
+class puppet_agent::install::remove_packages_osx {
+  assert_private()
+
+  if $::puppet_agent::is_pe {
+    # shutdown services
+    service { 'pe-puppet':
+      ensure => stopped,
+    }->
+    service { 'pe-mcollective':
+      ensure => stopped,
+    }->
+
+    # remove old users and groups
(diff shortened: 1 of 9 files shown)


manifests/install/remove_packages_osx.pp
```diff
@@ -0,0 +1,60 @@
...
+    user { 'pe-mcollective':
+      ensure => absent,
+    }->
+
+    # remove old /opt/puppet files
+    file { '/opt/puppet':
+      ensure => absent,
+      force  => true,
+      backup => false,
+    }
```
jpartlow: As an alternative to the above, we could do something delightful like an Exec of:

``` sh
for old_puppet_file in $(pkgutil --pkgs | grep puppet | xargs -Isub pkgutil --files sub | sort -r | uniq | grep '^opt/puppet'); do
  
  if [ -d "/${old_puppet_file?}" ]; then
    rmdir "/${old_puppet_file?}"
  elif [ -f "/${old_puppet_file?}" ]; then
    rm "/${old_puppet_file?}"
  fi
done
```

To preserve files added by the user; this would leave behind /opt/puppet/libexec/mcollective, and /opt/puppet/ssl/man/man3 (which presumably gets built post install.  The mcollective plugins are about the only reason I can think of to prefer this route, as this preserves custom plugins that might have been added by the user.

@highb @haus any opinions?

jpartlow:
This was finished up in #70

PR #68, opened 2015-10-26
(MODULES-2750) Pass in Puppet agent PID as command line parameter to avoid recreatin…

dhs-rec:
…g install_puppet.bat at every agent run.

### DIFF
--- a/manifests/windows/install.pp
+++ b/manifests/windows/install.pp
@@ -53,7 +53,7 @@ class puppet_agent::windows::install {
     content => template('puppet_agent/install_puppet.bat.erb')
   }->
   exec { 'install_puppet.bat':
-    command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${::env_temp_variable}\\install_puppet.bat\"",
+    command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${::env_temp_variable}\\install_puppet.bat\" ${::puppet_agent_pid}",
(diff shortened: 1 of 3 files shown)


ferventcoder:
I like this.

What we'd like to do here is have you create a ticket on https://tickets.puppetlabs.com/browse/PUP and then update this issue and the commit message with that ticket number. See https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md for more details.

Then it appears you will need to fix up the tests that are now broken based on this change.

dhs-rec:
The ticket is: https://tickets.puppetlabs.com/browse/PA-13

Will look into the tests...

ferventcoder:
Please squash those commits down to one so that specs continue to pass for each commit. Also you are removing test cases, but perhaps it would be good to prove that the command is calling the .bat file with a PID.

ferventcoder:
There is some talk on the ticket about where this should go, PA is probably not the place for the puppet_agent module. I'll wait until there is a decision on where that should go. In the meantime, you can go ahead with the changes I requested. Thanks!

kylog:
This is now https://tickets.puppetlabs.com/browse/MODULES-2750

ferventcoder:
@dhs-rec would you mind updating the commits to be `(MODULES-2750)` instead of `(PA-13)`? Thanks!

MikaelSmith:
@jpartlow you see any problems with this? Tempted to merge it in.

PR #70, opened 2015-12-04
Issue/master/pe 10914 add osx 109 upgrade

mwbutcher:


MikaelSmith:
Commit still says WIP.

mwbutcher:
rebased, fixup commit squashed, message fixed

jpartlow:
There's no actual information in the commit about what we're doing here (probably notes about how we're removing 3.8 would be most relevant).

manifests/install/remove_packages_osx.pp
```diff
@@ -0,0 +1,60 @@
...
+      'pe-hiera',
+      'pe-puppet-enterprise-release',
+      'pe-stomp',
+      'pe-libyaml',
+      'pe-ruby-rgen',
+      'pe-deep-merge',
+      'pe-ruby-shadow',
+    ].each |$package| {
+      exec { "forget ${package}":
+        command => "/usr/sbin/pkgutil --forget com.puppetlabs.${package}",
```
highb: Wow, the puppet osx package provider doesn't handle this? I'm disappointed. :/

MikaelSmith: PR's welcome.

highb:
@jpartlow @mwbutcher Just tested this out on a 3.8.3->2015.3 upgrade. Installed 3.8.3 on centos7, osx10.9. Verified mco ping/agent runs work. Upgraded centos7 master, verified agent runs work and puppet is not upgraded on agent. Installed puppet_agent module from this PR, added osx10.9 node to group with puppet_agent class applied, ran puppet/mco successfully.

@jpartlow Did you have any additional comments/concerns on this PR or are we good to merge?

jpartlow:
:+1:

manifests/install/remove_packages_osx.pp
highb: I should say I'm more disappointed in Apple for not writing a real package manager. It's definitely a hard problem to solve. Is there any way of getting a listing of all installed signatures so you could regex for `package_name` and uninstall if there is only one match?

PR #71, opened 2015-12-05
(PE-13179) Puppet Agent Module: Update metadata.json to include Anken…

mwbutcher:
…y release

Prior to this change the puppet agent module has a note that it's to be
used with PE < 2015.3. This Commit updates the metadata.json to note
PE Ankeny versions.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -25,6 +25,14 @@
       ]
     },
     {
+      "operatingsystem": "Darwin",
+      "operatingsystemrelease": [
+        "10.9",
+        "10.10",
+        "10.11"
+       ]
+    },
+    {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -69,7 +77,7 @@
   "requirements": [
     {
       "name": "pe",
-      "version_requirement": ">= 3.8.1 < 2015.3.0"
+      "version_requirement": ">= 2015.3.0 < 2016.1.0"
```
MikaelSmith: This module requirement applies to where it's being run, which could be as part of `puppet apply`. That could still include `3.8.1`, so I'm not sure why the minimum is changing.

mwbutcher: that makes sense

highb:
:+1: Sane to me.

PR #72, opened 2015-12-07
(PE-12001) Add Solaris 10 sparc to supported arch

highb:
Prior to this commit the puppet_agent module did not accept
the sparc (sun4u/v) arch on Solaris 10.
This commit adds that arch to the accepted architectures, as well
as add special handling on Solaris 10 to use the "sparc" packages
on those arches.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -21,7 +21,7 @@ class puppet_agent (
   $source        = $::puppet_agent::params::_source,
 ) inherits ::puppet_agent::params {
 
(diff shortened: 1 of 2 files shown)


jpartlow:
:+1: Brandon we're you able to get this patch to update puppet-agent on the cisparc10 host from some hapless master with the pe_repo::platforms::solaris_10_sparc class applied to it?

highb:
@jpartlow Yep, I was able to repro the arch not matching the regex, then I applied this patch, and it upgraded fine.

MikaelSmith:
:+1:

PR #73, opened 2015-12-18
(PE-12299) Add Solaris 11 Support

demophoon:
This PR consists of everything needed to get puppet agents on Solaris 11 to upgrade from puppet 3.x to puppet 4.x in a PE environment.

This PR is ready for review, and potentially even a merge.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -42,9 +42,9 @@ class puppet_agent::install(
     contain puppet_agent::install::remove_packages
 
     exec { 'puppet_agent restore /etc/puppetlabs':
-      command   => 'cp -r /tmp/puppet_agent/puppetlabs /etc',
-      path      => '/bin:/usr/bin:/sbin:/usr/sbin',
-      require   => Class['puppet_agent::install::remove_packages'],
+      command => 'cp -r /tmp/puppet_agent/puppetlabs /etc',
+      path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+      require => Class['puppet_agent::install::remove_packages'],
     }
 
     exec { 'puppet_agent post-install restore /etc/puppetlabs':
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -38,6 +38,25 @@ class puppet_agent::install(
       source    => "/opt/puppetlabs/packages/${_unzipped_package_name}",
       require   => Class['puppet_agent::install::remove_packages'],
     }
+  } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '11' {
+    contain puppet_agent::install::remove_packages
+
+    exec { 'puppet_agent restore /etc/puppetlabs':
+      command   => 'cp -r /tmp/puppet_agent/puppetlabs /etc',
+      path      => '/bin:/usr/bin:/sbin:/usr/sbin',
+      require   => Class['puppet_agent::install::remove_packages'],
+    }
+
+    exec { 'puppet_agent post-install restore /etc/puppetlabs':
+      command     => 'cp -r /tmp/puppet_agent/puppetlabs /etc',
+      path        => '/bin:/usr/bin:/sbin:/usr/sbin',
+      refreshonly => true,
+    }
+
+    $_package_options = {
+      require => Exec['puppet_agent restore /etc/puppetlabs'],
+      notify  => Exec['puppet_agent post-install restore /etc/puppetlabs'],
+    }
   } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ '10\.[9,10,11]' {
     contain puppet_agent::install::remove_packages
 
--- a/manifests/osfamily/solaris.pp
+++ b/manifests/osfamily/solaris.pp
@@ -44,9 +44,9 @@ class puppet_agent::osfamily::solaris(
         ensure => directory,
       }
       exec { 'puppet_agent backup /etc/puppetlabs/':
-        command  => 'cp -r /etc/puppetlabs/ /tmp/puppet_agent/',
-        require  => [Exec['puppet_agent copy packages'], File['/tmp/puppet_agent/']],
-        path     => '/bin:/usr/bin:/sbin:/usr/sbin',
+        command => 'cp -r /etc/puppetlabs/ /tmp/puppet_agent/',
+        require => [Exec['puppet_agent copy packages'], File['/tmp/puppet_agent/']],
+        path    => '/bin:/usr/bin:/sbin:/usr/sbin',
       }
 
       $pkgrepo_dir = '/etc/puppetlabs/installer/solaris.repo'
--- a/manifests/osfamily/solaris.pp
+++ b/manifests/osfamily/solaris.pp
@@ -52,12 +52,23 @@ class puppet_agent::osfamily::solaris(
 
       $pkgrepo_dir = '/etc/puppetlabs/installer/solaris.repo'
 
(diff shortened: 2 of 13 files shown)


demophoon:
I'm waiting on this to be validated on Solaris 11 sparc before removing the 'Do not merge' flag.

jpartlow:
@demophoon it sounds like it has been validated on solaris i386?

demophoon:
Yes, this has been validated on solaris 11 i386. The service bug is still a problem however. I am trying the windows style fix for that however.

jpartlow:
@demophoon The solaris11/windows restart scripts look equally evil.  So I guess that's :+1:

Is this starting the new services by your testing?  Also, looks like a rebase is needed.

demophoon:
its_working_star_wars

This is ready for some review

manifests/install.pp
```diff
@@ -38,6 +38,25 @@
...
+
+    exec { 'puppet_agent post-install restore /etc/puppetlabs':
+      command     => 'cp -r /tmp/puppet_agent/puppetlabs /etc',
+      path        => '/bin:/usr/bin:/sbin:/usr/sbin',
+      refreshonly => true,
+    }
+
+    $_package_options = {
+      require => Exec['puppet_agent restore /etc/puppetlabs'],
+      notify  => Exec['puppet_agent post-install restore /etc/puppetlabs'],
```
jpartlow: @demophoon I'm lost here, we seem to be restoring from /tmp/puppet_agent/puppetlabs twice?  Is the first a typo where we should be copying /etc to /tmp/puppet_agent/puppetlabs?

If so, how is this currently working?

manifests/osfamily/solaris.pp
```diff
@@ -32,6 +32,57 @@
...
+        package_file_name => $package_file_name,
+      }
+      contain puppet_agent::prepare::package
+
+      # Backup user configuration because solaris 11 will blow away
+      # /etc/puppetlabs/ when uninstalling the pe-* modules.
+      file { '/tmp/puppet_agent/':
+        ensure => directory,
+      }
+      exec { 'puppet_agent backup /etc/puppetlabs/':
```
jpartlow: Does this exec need a test?

manifests/install.pp
demophoon: @jpartlow We are backing up `/etc/puppetlabs/` here: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/73/files#diff-0b65fa0c1283d2894d43d73f820fe08eR46

Also the restore is needed twice because when we remove the pe-* packages all of /etc/puppetlabs is moved into a lost+found directory determined by the system. The first restore is needed before the puppet-agent package is installed because pkg expects the repo we setup to be in `/etc/puppetlabs/installer/solaris.repo/` with the new agent package. The second restore is because when you install the puppet-agent package solaris overwrites any existing user configuration in `/etc/puppetlabs/` leaving the puppet.conf in a bad state where it doesn't know its master or any other configuration that a user may have specified.

manifests/osfamily/solaris.pp
```diff
@@ -32,6 +32,57 @@
...
+        command => 'cp -r /etc/puppetlabs/ /tmp/puppet_agent/',
+        require => [Exec['puppet_agent copy packages'], File['/tmp/puppet_agent/']],
+        path    => '/bin:/usr/bin:/sbin:/usr/sbin',
+      }
+
+      $pkgrepo_dir = '/etc/puppetlabs/installer/solaris.repo'
+
+      exec { 'puppet_agent remove existing repo':
+        command   => "pkgrepo remove -s '${pkgrepo_dir}' '*'",
```
jpartlow: Should we be removing this prior to removing the old packages?

manifests/service/solaris.pp
```diff
@@ -13,5 +13,18 @@
     service { 'pe-puppet':
       ensure => stopped,
     }
+  } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '11' {
+    file { '/tmp/solaris_start_puppet.sh':
+      ensure  => file,
+      content => template('puppet_agent/solaris_start_puppet.sh.erb'),
+      mode    => '0755',
+    } ->
+    exec { 'solaris_start_puppet.sh':
```
jpartlow: So where does mcollective get handled?

demophoon: Mcollective is handled in the same script: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/73/files#diff-f2d66952a483ed099ff3363202e04b69R14

manifests/osfamily/solaris.pp
demophoon: That is a good question.

I'm not sure where this step should fall with regard to package uninstallation. I basically copied these out of pe_repo to follow the same logic. Since pe_repo doesn't uninstall the old packages i just put this where I thought it would make the most sense.

I could try an agent upgrade with this step happening before the uninstall and see what that does. I was thinking specifically before the `/etc/puppetlabs/` backup since that directory is currently clobbered on package removal.

manifests/osfamily/solaris.pp
demophoon: To test if we have already done the backup?

manifests/osfamily/solaris.pp
demophoon: @jpartlow So moving this to after the package uninstallation has occurred seems to not have an effect on anything. The whole of `/etc/puppetlabs/` is still removed making the backup and restoration of that directory still required for the puppet-agent package to install and for puppet to be in an okay state after the puppet run. We also still need to have the package dependencies in puppet code otherwise the run fails for attempting to uninstall the packages when other dependant packages are still installed.

jbcraig: This fails on my solaris 11.2 & 11.3 i386 nodes.  The `pkgrepo remove` command removes packages from within the repository, not the repository itself.  The repo is already empty (part of install process I believe) and the remove command results in an  exit code of 1.  I fixed by changing the command into:

`command   => "rm -rf '${pkgrepo_dir}'/'*'",`

The `pkgrepo create` that follows creates a new empty repository.  After making this change I successfully tested it on my solaris 11.2 i386 node.

manifests/install.pp
jpartlow: O_o

Ok, I think I've got it.

manifests/osfamily/solaris.pp
jpartlow: Yes; looks like you already added a test, or I was blind when I wrote the comment.

demophoon: Nope, I added that test in per your comment :)

jpartlow:
Sadly I'm getting the following error from a frankenbuild:

```
Notice: /Stage[main]/Puppet_agent::Install/Exec[puppet_agent restore /etc/puppetlabs]/returns: cp: cannot access /tmp/puppet_agent/puppetlabs
      Error: cp -r /tmp/puppet_agent/puppetlabs /etc returned 2 instead of one of [0]
      Error: /Stage[main]/Puppet_agent::Install/Exec[puppet_agent restore /etc/puppetlabs]/returns: change from notrun to 0 failed: cp -r /tmp/puppet_agent/puppetlabs /etc returned 2 instead of one of [0]
```

I'll add the full details and log to the ticket.

MikaelSmith:
Needs a rebase.

MikaelSmith:
Closing in favor of a rebased/squashed version at https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/93.

PR #75, opened 2016-01-11
(PE-13608) puppet_agent windows spec tests failing on posix systems

mwbutcher:
Prior to this commit there were several pathing issues that were preventing
puppet_agent spec tests for windows from running on posix systems
like Travis.

This commit switches some /'s with \'s as MSI cannot accept / and gates
the windows specific file resource ordering tests on being run against
a windows system.


demophoon:
Closed in favor of #76.

PR #78, opened 2016-01-15
(PE-11531/PUP-5222) Allow puppet_agent to upgrade versions > 3.8

highb:
This PR is a rebase/modification of https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/62 to add a `package_version` parameter to `puppet_agent`, which will allow a user to upgrade to/install a specific version of the puppet-agent package, including versions > 3.8, even when already upgraded to > 3.8. I un-did a slight refactor where `manifests/windows/install.pp` was rename `manifests/install/windows.pp` for easier tracking of changes and so it doesn't break the windows PR that we also need to merge. We can always do this refactor later with no other changes.

Tested it out with centos-7 by installing 2015.2.0 on a mono/agent, upgrading mono, then installing puppet_agent and setting the version to the current one. Upgrade went fine (the package was upgraded and none of the pre 3.8 changes occured).

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -7,11 +7,17 @@
 #
 # === Parameters
 #
+# [arch]
+#   The package architecture.
+# [is_pe]
+#   Install from Puppet Enterprise repos.
 # [package_name]
 #   The package to upgrade to, i.e. `puppet-agent`.
 # [service_names]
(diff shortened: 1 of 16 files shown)


highb:
Closing in favor of https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/86

PR #79, opened 2016-01-26
(MODULES-3015) Fix SLES11 GPG key import issue

glarizza:
Previously, the exec to import the Puppetlabs GPG key made a call out to the
gpg binary to gather the hex value of the Puppetlabs GPG key. This call works
from the console with an ENV, but it doesn't work when automated because it
cannot find the relative homedir (~/.gnupg). To work around this, explicitly
set the value for root's .gnupg homedir in the unless command and expose it as
a variable should we need to deal with it later.

### DIFF
--- a/manifests/osfamily/suse.pp
+++ b/manifests/osfamily/suse.pp
@@ -16,8 +16,9 @@ class puppet_agent::osfamily::suse(
     }
     '11', '12': {
       # Import the GPG key
-      $keyname = 'RPM-GPG-KEY-puppetlabs'
-      $gpg_path = "/etc/pki/rpm-gpg/${keyname}"
+      $keyname     = 'RPM-GPG-KEY-puppetlabs'
+      $gpg_path    = "/etc/pki/rpm-gpg/${keyname}"
+      $gpg_homedir = '/root/.gnupg'
 
       file { ['/etc/pki', '/etc/pki/rpm-gpg']:
         ensure => directory,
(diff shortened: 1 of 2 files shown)


highb:
@glarizza Looks like the lint checker doesn't like something about the indentation of the hashrockets? I'm not seeing it in the changes tab, what does it do when you run `bundle install && bundle exec rake test` locally?

Log for future readers:
```
manifests/osfamily/suse.pp - WARNING: indentation of => is not properly aligned on line 37
manifests/osfamily/suse.pp - WARNING: indentation of => is not properly aligned on line 38
manifests/osfamily/suse.pp - WARNING: indentation of => is not properly aligned on line 39
manifests/osfamily/suse.pp - WARNING: indentation of => is not properly aligned on line 40
manifests/osfamily/suse.pp - WARNING: indentation of => is not properly aligned on line 41
```

manifests/osfamily/suse.pp
```diff
@@ -33,11 +34,11 @@
...
       # Given the path to a key, see if it is imported, if not, import it
       exec {  "import-${keyname}":
-        path      => '/bin:/usr/bin:/sbin:/usr/sbin',
-        command   => "rpm --import ${gpg_path}",
-        unless    => "rpm -q gpg-pubkey-`echo $(gpg --throw-keyids < ${gpg_path}) | cut --characters=11-18 | tr [A-Z] [a-z]`",
-        require   => File[$gpg_path],
-        logoutput => 'on_failure',
+        path           => '/bin:/usr/bin:/sbin:/usr/sbin',
+        command        => "rpm --import ${gpg_path}",
+        unless         => "rpm -q gpg-pubkey-$(echo $(gpg --homedir ${gpg_homedir} --throw-keyids < ${gpg_path}) | cut --characters=11-18 | tr [A-Z] [a-z])",
```
jpartlow: I'm trying to come up with less painful variations to this line and not coming up with much.

There is a minimal change that I think would be helpful:

rpm -q gpg-pubkey-$(gpg --homedir ${gpg_homedir} --throw-keyids --with-colons < ${gpg_path} | cut -d: -f5 | tr [A-Z] [a-z])

Should at least get away from specifying the character positions (which I think were off by one).  Also I /think/ the echo was unnecessary?

highb:
Alright, updated the specs, fixed the linter nonsense, specs ran fine. :+1: Thanks for the fix!

PR #80, opened 2016-01-28
Use slashes for regex

glarizza:
Puppet introduced the ability to use strings as the right operand of the regex
match operator, however, when testing this on 3.8.4 SLES11 boxes, I discovered
that they arose errors if the right operand wasn't a regular expression. This
commit corrects that to account for the errors.


highb:
Alright, rebased this and caught a few more strings on the right hand of `=~`. Specs pass so :+1:

PR #81, opened 2016-01-29
(PE-13179) remove lower version requirement in code

mwbutcher:
Prior to this commit there was a lower bound on what versions of
puppet could be upgraded by the agent module. This has caused
users to rely on supplemental tools in some cases to facilitate
upgrades. In order to encourage upgrades and make the process
easier, we're removing the lower bound.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,10 +23,7 @@ class puppet_agent (
 
   validate_re($arch, ['^x86$','^x64$','^i386$','^i86pc$','^amd64$','^x86_64$','^power$','^sun4[uv]$','PowerPC_POWER'])
 
-  if versioncmp("${::clientversion}", '3.8.0') < 0 {
-    fail('upgrading requires Puppet 3.8')
-  }
-  elsif versioncmp("${::clientversion}", '4.0.0') >= 0 {
+  if versioncmp("${::clientversion}", '4.0.0') >= 0 {
     info('puppet_agent performs no actions on Puppet 4+')
   }
   else {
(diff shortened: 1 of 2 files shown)


jpartlow:
Actually, it's not the metadata that needs to have lower bound removed. The manifest itself is constrained.  And we probably want the metadata to continue to express the fact that we only test from 3.8.  (See ticket comments)

jpartlow:
@mwbutcher there are spec failures that need to be looked at, and the commit message summary needs updating.

mwbutcher:
@jpartlow The specs and message should be fixed now.

demophoon:
:+1: from me, code looks good, ran with a PE 3.7.2 agent upgrading to PE 2015.3 on centos-7 x86_64.

jpartlow:
@mwbutcher @demophoon let's open a separate MODULES ticket (in our puppet agent epic) for this (and fix the commit message title) so that  PE-13179 can be closed, since that metadata change is already in place.

MikaelSmith:
The manifest is someone constrained because the module requires future parser.

MikaelSmith:
However if people want to try it, they're welcome to. So this seems ok.

MikaelSmith:
Repurposed PE-13179 for this, as it was already a duplicate of another ticket.

MikaelSmith:
As long as you're connecting to a Puppet 4 master (such as Puppet Server 2.1), this seems to actually work. I've tested back to Puppet 3.5.1 so far; the puppet apply acceptance tests fail in 3.5.1, but work in 3.7 (haven't tried 3.6 yet). I'll have a PR up in a minute for more flexible local testing.

PR #85, opened 2016-02-12
add support for Amazon Linux to puppet_agent::osfamily::redhat

oshaughnessy:
Amazon Linux looks enough like CentOS 6 that we can treat it that way for purposes of upgrading the puppet agent RPM. The end result of the change is that on Amazon Linux releases, the module configures the yum "pc1_repo" baseurl to the "el-6-${::architecture}" subdirectory instead of something
like "el-2015-x86_64" or "el-2014-x86_64".

This was tested on Amazon Linux 2015.03 and 2014.09.

### DIFF
--- a/manifests/osfamily/redhat.pp
+++ b/manifests/osfamily/redhat.pp
@@ -23,8 +23,15 @@ class puppet_agent::osfamily::redhat(
     $_sslclientcert_path = "${_ssl_dir}/certs/${::clientcert}.pem"
     $_sslclientkey_path = "${_ssl_dir}/private_keys/${::clientcert}.pem"
 
+    # Treat Amazon Linux just like Enterprise Linux 6
+    if $::operatingsystem == 'Amazon' {
+      $_repo_dir = "el-6-${::architecture}"
+    }
+    else {
+      $_repo_dir = $::platform_tag
+    }
     $pe_server_version = pe_build_version()
-    $source = "${::puppet_agent::source}/${pe_server_version}/${::platform_tag}"
+    $source = "${::puppet_agent::source}/${pe_server_version}/${_repo_dir}"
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
I'd like to see an update to test osfamily=Amazon in https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/spec/classes/puppet_agent_osfamily_redhat_spec.rb#L4 along with this change.

MikaelSmith:
@highb mind giving this a look? I'm fine merging it and putting up a new PR to add specs.

highb:
@MikaelSmith @oshaughnessy I'm not very familiar with AWS's images; are they all el-6 derivatives? Is there any possibility that it would be a different version that we would need to account for?

MikaelSmith:
I believe it's el-6 now, but that could certainly change in the future. I can fire up an Amazon AMI, we may be able to detect major version on it in a way that will (hopefully) stay relevant.

MikaelSmith:
I don't see an obvious way to detect which EL release an Amazon AMI is based on. Maybe it should be restricted to the 2015.09 release?

highb:
Or we document that it has only been tested on 2015.09. Amazon releases new AMIs fairly frequently, right? It would be a bit of overhead to keep that version up to date, if so.

oshaughnessy:
Thanks, guys. Sorry I didn't jump back in here in time to add a spec addition.

MikaelSmith:
That's ok, I added some as part of https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/92.

I also fixed up the OSS repo creation to correctly identify EL 6. I'm not sure that really helps anyone, because if they're using AMIs they're probably using the Puppet3/Facter2 packages, and puppet-agent conflicts with them.

PR #86, opened 2016-02-12
(PE-11531) Allow upgrades from Puppet 4+

highb:
Prior to this commit we only allowed upgrades with this
module if the current puppet version was pre-4.
This commit removes that restriction and hopefully handles
all the edge cases involved in upgrade between new versions
and doesn't break FOSS. This feature should be tested
thoroughly.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,7 @@
 fixtures:
   repositories:
     stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
+    transition: "https://github.com/puppetlabs/puppetlabs-transition.git"
     inifile: "https://github.com/puppetlabs/puppetlabs-inifile.git"
     apt: "https://github.com/puppetlabs/puppetlabs-apt.git"
   symlinks:
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,49 +1,70 @@
 # == Class: puppet_agent
 #
-# Upgrades Puppet 3.8 to Puppet 4+ (Puppet-Agent from Puppet Collection 1).
-# Makes the upgrade easier by migrating SSL certs and config files to the new
-# Puppet-Agent paths and removing deprecated settings that are no longer
+# Upgrades Puppet 3.8 and newer to the requested version.
+# Makes Puppet 4 upgrades easier by migrating SSL certs and config files to the
+# new Puppet-Agent paths and removing deprecated settings that are no longer
 # supported by Puppet 4.
 #
 # === Parameters
 #
+# [arch]
+#   The package architecture. Defaults to the architecture fact.
+# [collection]
+#   The Puppet Collection to track. Defaults to 'PC1'.
+# [is_pe]
+#   Install from Puppet Enterprise repos. Enabled if communicating with a PE master.
 # [package_name]
 #   The package to upgrade to, i.e. `puppet-agent`.
+# [package_version]
+#   The package version to upgrade to. Defaults to undef, meaning only upgrade from Puppet < 4.0.
 # [service_names]
 #   An array of services to start, normally `puppet` and `mcollective`.
 #   None will be started if the array is empty.
+# [source]
+#   The location to find packages.
 #
 class puppet_agent (
-  $arch          = $::architecture,
-  $is_pe         = $::puppet_agent::params::_is_pe,
-  $package_name  = $::puppet_agent::params::package_name,
-  $service_names = $::puppet_agent::params::service_names,
-  $source        = $::puppet_agent::params::_source,
+  $arch            = $::architecture,
+  $collection      = 'PC1',
+  $is_pe           = $::puppet_agent::params::_is_pe,
+  $package_name    = $::puppet_agent::params::package_name,
+  $package_version = undef,
+  $service_names   = $::puppet_agent::params::service_names,
+  $source          = $::puppet_agent::params::_source,
 ) inherits ::puppet_agent::params {
 
   validate_re($arch, ['^x86$','^x64$','^i386$','^i86pc$','^amd64$','^x86_64$','^power$','^sun4[uv]$','PowerPC_POWER'])
 
-  if versioncmp("${::clientversion}", '4.0.0') >= 0 {
-    info('puppet_agent performs no actions on Puppet 4+')
+  if ($package_version == undef and $puppet_agent::params::master_agent_version != undef and versioncmp("${::clientversion}", '4.0.0') < 0) {
+    $_package_version = $puppet_agent::params::master_agent_version
+  } elsif $package_version != undef {
+    $_package_version = $package_version
   }
-  else {
+
+  if $_package_version == undef  {
+    info("puppet_agent performs no actions if a package_version is not specified on Puppet 4, or if the master's agent version cannot be determined on Puppet 3.8")
+  } else {
+    if $package_version != undef and $package_version !~ /^\d+\.\d+\.\d+([.-]?\d*|\.\d+\.g[0-9a-f]+)$/ {
+      fail("invalid version ${package_version} requested")
+    }
+
     if $::architecture == 'x86' and $arch == 'x64' {
       fail('Unable to install x64 on a x86 system')
     }
 
     if $::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10' {
-      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.sles10.${::architecture}.rpm"
+      $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.sles10.${::architecture}.rpm"
     } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '10' {
       if $arch =~ /^sun4[uv]$/ {
-        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.sparc.pkg.gz"
+        $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.sparc.pkg.gz"
       } else {
-        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.i386.pkg.gz"
+        $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.i386.pkg.gz"
       }
     } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /10\.[9,10,11]/ {
-      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
+      $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.osx${$::macosx_productversion_major}.dmg"
     } elsif $::operatingsystem == 'aix' and $::architecture =~ /PowerPC_POWER[5,6,7]/ {
       $aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
-      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.aix${aix_ver_number}.ppc.rpm"
+      $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.aix${aix_ver_number}.ppc.rpm"
     } elsif $::osfamily == 'windows' {
       $_arch = $::kernelmajversion ?{
         /^5\.\d+/ => 'x86', # x64 is never allowed on windows 2003
@@ -52,6 +73,8 @@ class puppet_agent (
 
       if $is_pe {
         $_package_file_name = "${package_name}-${_arch}.msi"
+      } elsif $_package_version != undef {
+        $_package_file_name = "${package_name}-${_arch}-${_package_version}.msi"
       } else {
         $_package_file_name = "${package_name}-${_arch}-latest.msi"
       }
@@ -61,16 +84,19 @@ class puppet_agent (
 
     class { '::puppet_agent::prepare':
       package_file_name => $_package_file_name,
+      package_version   => $_package_version,
     } ->
     class { '::puppet_agent::install':
       package_file_name => $_package_file_name,
+      package_version   => $_package_version,
     }
 
     contain '::puppet_agent::prepare'
     contain '::puppet_agent::install'
 
     # On windows, our MSI handles the services
-    if $::osfamily != 'windows' {
+    # On PE AIO nodes, PE Agent nodegroup is managing the services
+    if $::osfamily != 'windows' and (!is_pe or versioncmp($::clientversion, '4.0.0') < 0) {
       class { '::puppet_agent::service':
         require => Class['::puppet_agent::install'],
       }
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -7,9 +7,12 @@
 # [package_file_name]
 #   The puppet-agent package file name.
 #   (see puppet_agent::prepare::package_file_name)
+# [version]
+#   The puppet-agent version to install.
 #
 class puppet_agent::install(
   $package_file_name = undef,
+  $package_version   = 'present',
 ) {
   assert_private()
 
@@ -50,20 +53,36 @@ class puppet_agent::install(
   }
 
   if $::osfamily == 'windows' {
-    if $::puppet_agent::is_pe == true and empty($::puppet_agent::source) and defined(File["${::puppet_agent::params::local_packages_dir}/${package_file_name}"]) {
-      class { 'puppet_agent::windows::install':
-        package_file_name => $package_file_name,
-        source            => windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}"),
+    # Prevent re-running the batch install
+    if versioncmp("${::aio_agent_version}", "${package_version}") < 0 {
+      if $::puppet_agent::is_pe == true and empty($::puppet_agent::source) and defined(File["${::puppet_agent::params::local_packages_dir}/${package_file_name}"]) {
+        class { 'puppet_agent::windows::install':
+          package_file_name => $package_file_name,
+          source            => windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}"),
+        }
+      } else {
+        class { 'puppet_agent::windows::install':
+          package_file_name => $package_file_name,
+          source            => $::puppet_agent::source,
+        }
       }
-    } else {
-      class { 'puppet_agent::windows::install':
-        package_file_name => $package_file_name,
-        source            => $::puppet_agent::source,
-      }
+    }
+  } elsif ($::osfamily == 'Solaris' and $::operatingsystemmajrelease == '10') or $::osfamily == 'Darwin' or $::osfamily == 'AIX' or ($::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10') {
+    # Solaris 10/OSX/AIX/SLES 10 package provider does not provide 'versionable'
+    # Package is removed above, then re-added as the new version here.
+    package { $::puppet_agent::package_name:
+      ensure => 'present',
+      *      => $_package_options,
+    }
+  } elsif ($::osfamily == 'RedHat') and ($package_version != 'present') {
+    # Workaround PUP-5802/PUP-5025
+    package { $::puppet_agent::package_name:
+      ensure => "${package_version}-1.el${::operatingsystemmajrelease}",
+      *      => $_package_options,
     }
   } else {
     package { $::puppet_agent::package_name:
-      ensure => present,
+      ensure => $package_version,
       *      => $_package_options,
     }
   }
--- a/manifests/install/remove_packages.pp
+++ b/manifests/install/remove_packages.pp
@@ -4,33 +4,37 @@
 # manager, where we are required to manually remove the old pe-* packages prior
 # to installing puppet-agent.
 #
-class puppet_agent::install::remove_packages {
+class puppet_agent::install::remove_packages(
+  $package_version = undef
+  ) {
   assert_private()
 
-  if versioncmp("${::clientversion}", '4.0.0') < 0 {
-
-    if $::operatingsystem == 'Darwin' {
-
-      contain '::puppet_agent::install::remove_packages_osx'
-
-    } else {
-
-      $package_options = $::operatingsystem ? {
-        'SLES'  => {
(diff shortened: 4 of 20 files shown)


highb:
I've tested this out with 3.8->2015.3.2 as well as 2015.2.1->2015.3.2 upgrades on centos7 and windows 2012 and it appears that `agent_upgrade.rb` passes. We'll definitely need to test more platforms.

jpartlow:
@highb tests are failing lint of manifests/osfamily/debian.pp - WARNING: variable not enclosed in {} on line 75.  I tried to push a commit fix to your branch but hit a permission issue (?)

Once that's fixed I'm seeing 70 spec errors around puppet_agent::service.  Are you not seeing those locally?

highb:
@jpartlow Hm, nope I don't think I was seeing those spec failures locally. Maybe something is off in my bundler environment, I'll poke around after I've had a change to check out these other PRs on the board.

highb:
@jpartlow OK, yeah was able to repro the spec failures locally. I'm not sure how I didn't see those before.  I'll get those fixed up.

jpartlow:
@highb Now we're seeing spec failures just for Puppet 3.7.5 and Puppet 4.0

highb:
@jpartlow Yep, I'm going to look into those once I've finished reviewing PRs.

manifests/init.pp
```diff
@@ -66,16 +87,19 @@
...
+      package_version   => $package_version,
     }
 
     contain '::puppet_agent::prepare'
     contain '::puppet_agent::install'
 
     # On windows, our MSI handles the services
-    if $::osfamily != 'windows' {
+    # On PE AIO nodes, PE Agent nodegroup is managing the services
+    if $::osfamily != 'windows' and versioncmp($::clientversion, '4.0.0') < 0 {
```
MikaelSmith: Shouldn't this be restricted to is_pe as well then?

highb: Oh yes, that'd be good to check.

manifests/init.pp
```diff
@@ -1,54 +1,75 @@
...
+    $_package_version = $package_version
   }
-  elsif versioncmp("${::clientversion}", '4.0.0') >= 0 {
-    info('puppet_agent performs no actions on Puppet 4+')
+
+  if versioncmp("${::clientversion}", '3.8.0') < 0 {
+    fail('upgrading requires at least Puppet 3.8')
   }
-  else {
+  elsif $is_pe and $::aio_agent_version != undef and $_package_version != undef and versioncmp("${::aio_agent_version}", "${_package_version}") >= 0 {
```
thrnio: Why does this do nothing if up-to-date only if running on PE?

manifests/init.pp
```diff
@@ -66,16 +87,19 @@
...
     class { '::puppet_agent::prepare':
       package_file_name => $_package_file_name,
+      package_version   => $package_version,
     } ->
     class { '::puppet_agent::install':
       package_file_name => $_package_file_name,
+      package_version   => $package_version,
     }
 
     contain '::puppet_agent::prepare'
```
MikaelSmith: I think MODULES-3118 implies this needs to happen even if the aio_agent_version and package_version match.

manifests/init.pp
MikaelSmith: We should probably still manage all the resources, and handle ensuring we don't make changes when the package installed is the correct version on a more granular level.

highb: That makes sense to me. I think the only `is_pe` checks we should be doing are for the service check (since PE already manages the services) and for the PE only platforms (AIX, etc)

thrnio: :+1:

highb:
@MikaelSmith Hm... we should probably think up a method for preventing that.

MikaelSmith:
Or support it. I'm ok with simply documenting it for now.

highb:
@MikaelSmith @thrnio I've finished up fixing the specs to test Puppet 4+, and addressed the lack of resource management in already upgraded PE nodes. I'm going to run some tests to verify PE doesn't break in some new and surprising way.

highb:
OK, one more spec fix is needed for pre-4 behavior.

Gemfile
```diff
@@ -2,7 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 group :test do
   gem 'rake', '~> 10.4'
-  gem 'puppet', ENV['PUPPET_VERSION'] || '~> 3.8'
+  gem 'puppet', ENV['PUPPET_VERSION'] || '~> 4.0'
```
MikaelSmith: Why is this change needed?

highb: Oh, it isn't really. That's just what I've been using for testing. All the tests should be specifying PUPPET_VERSION, though so it seems like a sensible default to me?

MikaelSmith: Ok.

manifests/init.pp
```diff
@@ -1,54 +1,72 @@
 # == Class: puppet_agent
 #
-# Upgrades Puppet 3.8 to Puppet 4+ (Puppet-Agent from Puppet Collection 1).
-# Makes the upgrade easier by migrating SSL certs and config files to the new
-# Puppet-Agent paths and removing deprecated settings that are no longer
+# Upgrades Puppet 3.8 and newer to the requested version.
```
MikaelSmith: Worth being clear that it will only upgrade to a puppet-agent package.

manifests/init.pp
```diff
@@ -1,54 +1,72 @@
...
   validate_re($arch, ['^x86$','^x64$','^i386$','^i86pc$','^amd64$','^x86_64$','^power$','^sun4[uv]$','PowerPC_POWER'])
 
-  if versioncmp("${::clientversion}", '3.8.0') < 0 {
-    fail('upgrading requires Puppet 3.8')
-  }
-  elsif versioncmp("${::clientversion}", '4.0.0') >= 0 {
-    info('puppet_agent performs no actions on Puppet 4+')
+  if ($package_version == undef and versioncmp("${::clientversion}", '4.0.0') < 0) {
+    $_package_version = $puppet_agent::params::master_agent_version
+  } elsif ($package_version != undef and versioncmp("${::clientversion}", $package_version) != 0) {
```
MikaelSmith: I'm not clear what this check is doing?

Also, is there an attempt to skip installing if the correct version is already installed?

MikaelSmith: Can we copy the `aio_agent_version` fact to this module, so OSS upgrades can use it? We should be checking whether `aio_agent_version` matches requested version, and not performing an install on platforms without package managers (or maybe just skip it on all platforms).

thrnio: It's kind of silly IMO that the `aio_agent_version` fact is only shipped with PE.

MikaelSmith: This may not be a major problem... rerunning the MSI on Windows takes about 5 seconds.

MikaelSmith: Filed https://tickets.puppetlabs.com/browse/PA-241 to add it.

MikaelSmith: Something like
```
diff --git a/manifests/install.pp b/manifests/install.pp
index 60cdde2..806f219 100644
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -16,6 +16,7 @@ class puppet_agent::install(
 ) {
   assert_private()

+  if ($::aio_agent_version != $package_version) {
     if ($::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10') or ($::operatingsystem == 'AIX' and  $::architecture =~ /PowerPC_POWER[5,6,7]/) {
       contain puppet_agent::install::remove_packages

@@ -78,3 +79,4 @@ class puppet_agent::install(
       }
     }
   }
+}
```
would work. We could look at adding a custom fact, or just plan to address it for OSS in a future puppet-agent release.

manifests/init.pp
```diff
@@ -1,54 +1,72 @@
...
     } elsif $::operatingsystem == 'aix' and $::architecture =~ /PowerPC_POWER[5,6,7]/ {
       $aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
-      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.aix${aix_ver_number}.ppc.rpm"
+      $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.aix${aix_ver_number}.ppc.rpm"
     } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /10\.[9,10,11]/ {
-      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
+      $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.osx${$::macosx_productversion_major}.dmg"
     } elsif $::osfamily == 'windows' {
       $_arch = $::kernelmajversion ?{
         /^5\.\d+/ => 'x86', # x64 is never allowed on windows 2003
```
MikaelSmith: This section needs to be updated to use the package version.

manifests/init.pp
highb: Hm, I guess we should just set `_package_version` if `package_version` is set. The version mismatch check was from when I was thinking the module would do a no-op if an upgrade wasn't needed.

manifests/init.pp
highb: I recall going through a similar line of thought where I assumed that there would be a version on the package, but I think the Windows MSIs don't have versions specified on the package name?

MikaelSmith: Might depend on where we get them. OSS versions do: https://downloads.puppetlabs.com/windows/

highb: Interesting. Currently the windows prepare step is very PE-centric. I think it depends on pe_repo being setup on the master. We could just document that if you want to use this module with FOSS, you should re-name your MSI?

jpartlow: There's something odd about that, because I thought windows upgrades were working in FOSS, and what we did for PE, was just deal with a) we needed to limit to installing windows packages provided with that pe version via pm.puppetlabs.com (downloaded locally to pe_repo), b) the fact that the master's cert was not trusted by windows boxes as far as transferring the windows package.

MikaelSmith: Upgrades work, but they always get the latest release. With the change to specify a specific version, that logic needs to be updated.

highb: Ah, right.

manifests/init.pp
MikaelSmith: This should only be needed for package providers that don't support versionable (Mac and Solaris 10), and Windows because we do an out-of-band upgrade. It looks like every provider we'll encounter that supports versionable also supports upgradeable, so no extra concerns there.

manifests/install.pp
```diff
@@ -61,9 +64,16 @@
         source            => $::puppet_agent::source,
       }
     }
+  } elsif ($::osfamily == 'Solaris' and $::operatingsystemmajrelease == '10') or $::osfamily == 'Darwin' {
+    # Solaris 10/OSX package provider does not provide 'versionable'
+    # Package is removed above, then re-added as the new version here.
+    package { $::puppet_agent::package_name:
+      ensure => 'present',
```
MikaelSmith: How does 'present' ensure the correct package? Because we supply a source?

highb: Yeah, since Solaris doesn't support versioned, we can't give a version here. So we just ensure that the source is the correct version.

highb:
@MikaelSmith Alright, I think I've addressed all your feedback and rebased the PR since it had some new merge conflicts.

manifests/init.pp
```diff
@@ -1,51 +1,72 @@
...
+  $source          = $::puppet_agent::params::_source,
 ) inherits ::puppet_agent::params {
 
   validate_re($arch, ['^x86$','^x64$','^i386$','^i86pc$','^amd64$','^x86_64$','^power$','^sun4[uv]$','PowerPC_POWER'])
 
-  if versioncmp("${::clientversion}", '4.0.0') >= 0 {
-    info('puppet_agent performs no actions on Puppet 4+')
+  if ($package_version == undef and versioncmp("${::clientversion}", '4.0.0') < 0) {
+    $_package_version = $puppet_agent::params::master_agent_version
+  } else
```
MikaelSmith: Missing an open curly brace, caused CI to fail.

manifests/install.pp
```diff
@@ -7,9 +7,12 @@
 # [package_file_name]
 #   The puppet-agent package file name.
 #   (see puppet_agent::prepare::package_file_name)
+# [version]
+#   The puppet-agent version to install.
 #
 class puppet_agent::install(
   $package_file_name = undef,
+  $package_version   = 'present',
```
MikaelSmith: Will this ever be used? We always pass an argument in init.pp, so seems like it will either be a version or undef.

manifests/init.pp
MikaelSmith: I didn't see any changes corresponding to this problem (not repeating installs on Mac, Solaris 10, and Windows).

manifests/init.pp
highb: Odd, I remember fixing that.

highb:
I'll finish up the fixes for the specs and another rebase tomorrow.

highb:
Rebased again, now going to work on the platforms that need their packages removed before upgrading.

manifests/init.pp
```diff
@@ -1,46 +1,67 @@
...
       } else {
-        $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.i386.pkg.gz"
+        $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.i386.pkg.gz"
       }
     } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /10\.[9,10,11]/ {
-      $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.osx${$::macosx_productversion_major}.dmg"
+      $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.osx${$::macosx_productversion_major}.dmg"
     } elsif $::operatingsystem == 'aix' and $::architecture =~ /PowerPC_POWER[5,6,7]/ {
       $aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
       $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.aix${aix_ver_number}.ppc.rpm"
```
highb: Hm, this `${puppet_agent::params::master_agent_version}` should be `${package_version}`.

MikaelSmith: Looks like it.

highb:
Specs were failing due to the linter not liking some double quoted strings w/o variables in them that I added. Fixed up. I've run the specs with Puppet 3.8 locally, hopefully nothing broke on 4.0

manifests/install/remove_packages.pp
```diff
@@ -69,14 +67,22 @@
           'pe-ruby-ldap',
         ]
       }
+    } elsif versioncmp("${::aio_agent_version}", "${::package_version}") < 0 {
+      if $::operatingsystem == 'Solaris' {
+        $packages = [
+          'PUPpuppet-agent',
```
MikaelSmith: This is necessary?

MikaelSmith:
Looks good, except I think Windows will still unnecessarily re-install the MSI. You should guard windows/install.pp if aio_agent_version matches clientversion.

manifests/install/remove_packages.pp
highb: I think so, solaris is weird. I'll test it out to be sure.

Also, I still need to add SLES/AIX below to get around their weird, non-versionable package managers.

highb:
@MikaelSmith Then it will only be a problem on FOSS until we ship `aio_agent_version` with facter, right?

highb:
I'm testing all this out as I go along with pe_acceptance_tests/agent_upgrade.rb. I just added a minor change to agent_upgrade to add the package_version if we're upgrading from a PE_VER > 3.8.

manifests/install/remove_packages.pp
highb: Oh no, this is going to cause a duplicate resource declaration...

MikaelSmith: Why?

highb: Ah, just ran a solaris upgrade and it *is* just `puppet-agent` now! Fixing...

highb:
Bah, didn't mean to push that up yet. Specs will fail because I haven't fixed some syntax issues and the specs still need another update. Update on testing: I've been working around the yum ensure => version issue that has been causing grief. It requires the `-1.el$majorversion` bit on the version, also just running 3.8.4/2015.3 tests takes quite a while...

highb:
Most recent changes have been tested on 3.8.4->2016.1 upgrade for centos7, solaris10 and sles10. Going to try out 2015.3.3->2016.1 with the same platforms plus OSX1010.

.fixtures.yml
```diff
@@ -1,6 +1,7 @@
 fixtures:
   repositories:
     stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
+    transition: "https://github.com/puppetlabs/puppetlabs-transition.git"
```
highb: Needed because otherwise we would have duplicate resource declarations on platforms that need to do the package remove/install dance to upgrade (Solaris/SLES/AIX)

manifests/install.pp
```diff
@@ -50,20 +53,37 @@
...
-      class { 'puppet_agent::windows::install':
-        package_file_name => $package_file_name,
-        source            => windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}"),
-      }
-    } else {
-      class { 'puppet_agent::windows::install':
-        package_file_name => $package_file_name,
-        source            => $::puppet_agent::source,
+    # Prevent re-running the batch install
+    if versioncmp("${::agent_version}", "${package_version}") < 0 {
```
MikaelSmith: So this blocks downgrades. That seems ok for now, since we don't officially support them.

MikaelSmith:
Failed CI. I'm :+1: on this in general though.

manifests/install/remove_packages.pp
```diff
@@ -69,13 +73,24 @@
...
+      if (versioncmp("${::clientversion}", '4.0.0') < 0) {
         package { $old_package:
-          ensure => absent,
-          *      => $package_options,
+          * => $package_options,
+        }
+      } else {
+        # We must use transition here because we would have a duplicate package
+        # declaration if we used a Package.
+        notify { "using puppetlabs-transition to remove ${old_package}: ${::operatingsystem} does not support versionable": }
```
MikaelSmith: You could use a name and message here to make the notify simpler.

MikaelSmith: And don't you need to declare `package { $old_Package }` still?

highb: Yeah, if I did declare it, that's where the duplicate resource errors come up.

highb: Naming it would probably be a good call. :P

highb:
@MikaelSmith Yeah, because I've added a module the Puppetfile isn't updated, so I had to hack around in https://github.com/puppetlabs/pe_acceptance_tests/blob/2016.2.x/setup/agent_upgrade.rb#L112 to add the transition module. :/

highb:
Damn it, now the dev build version format is causing problems again. I guess I'll just make dev builds acceptable to the `package_version` regex.

manifests/init.pp
```diff
@@ -1,49 +1,72 @@
...
-  if versioncmp("${::clientversion}", '4.0.0') >= 0 {
-    info('puppet_agent performs no actions on Puppet 4+')
+  if ($package_version == undef and $puppet_agent::params::master_agent_version != undef and versioncmp("${::clientversion}", '4.0.0') < 0) {
+    $_package_version = $puppet_agent::params::master_agent_version
+  } elsif $package_version != undef {
+    $_package_version = $package_version
   }
-  else {
+
+  if versioncmp("${::clientversion}", '3.8.0') < 0 {
```
MikaelSmith: This didn't rebase correctly.

spec/classes/puppet_agent_osfamily_aix_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 
-describe 'puppet_agent', :unless => Puppet.version < "3.8.0" || Puppet.version >= "4.0.0" do
+describe 'puppet_agent', :unless => Puppet.version < "3.8.0" do
```
MikaelSmith: We can stop guarding against version less than 3.8. Missed that in the commit that dropped the 3.8 guard.

spec/classes/puppet_agent_spec.rb
```diff
@@ -29,31 +34,58 @@
...
 
             it { is_expected.to compile.with_all_deps }
+            it { expect { catalogue }.not_to raise_error }
+          end
+        end
 
-            it { is_expected.to contain_class('puppet_agent') }
-            it { is_expected.to contain_class('puppet_agent::params') }
-            if Puppet.version < "4.0.0"
+        if Puppet.version < "3.8.0"
```
MikaelSmith: This rebased incorrectly.

manifests/install.pp
```diff
@@ -50,20 +53,36 @@
...
+    # Solaris 10/OSX/AIX/SLES 10 package provider does not provide 'versionable'
+    # Package is removed above, then re-added as the new version here.
+    package { $::puppet_agent::package_name:
+      ensure => 'present',
+      *      => $_package_options,
+    }
+  } elsif ($::osfamily == 'RedHat') and ($package_version != 'present') {
+    # Workaround PUP-5802/PUP-5025
+    package { $::puppet_agent::package_name:
+      ensure => "${package_version}-1.el${::operatingsystemmajrelease}",
```
MikaelSmith: Yes, operatingsystemmajrelease is the right way to do this.

spec/classes/puppet_agent_spec.rb
highb: D'oh. Good catch.

manifests/install.pp
thrnio: I thought 3.8 and stringified facts disabled was a requirement for this module?

MikaelSmith: Stringified facts yes, but the 3.8 requirement (on agents) is going away.

highb:
Hopefully that's the last spec fix. Ever closer...

highb:
@MikaelSmith Commits squashed, I preserved the rest of the commit messages in the squash commit for posterity. Windows specs updated to check against `1.2.1.1`.

PR #90, opened 2016-03-07
(maint) Obtain apt signing key from MoM on PE

thrnio:
This updates `puppet_agent::osfamily::debian` to pull the package
signing key from the source specified in `$::puppet_agent::source`
rather than from pgp.mit.edu. This allows the module to work properly
when a node has access to the master but does not have general external
web access.

This doesn't change the behavior on FOSS installations since the module
already assumes that the node can reach public servers. It also doesn't
change the behavior for other agent operating systems because Debian is
the only OS where we pull the signing key from pgp.mit.edu.


thrnio:
This doesn't actually work because I misunderstood what the `apt` module was doing here.

PR #91, opened 2016-03-11
(maint) Fix acceptance using Vagrant configs, allow flexible starting Puppet version

MikaelSmith:
Changes to puppetserver defaults and beaker VM naming caused acceptance
to start failing when using Vagrant. Add the full VM name to
dns_alt_names and increase memory allocated to the master.

Previously acceptance tests would install puppet-agent 1.2.1, then
install the latest puppetserver prompting an upgrade to a new version of
puppet-agent. Now always install the latest puppet-agent and
puppetserver.

Also provide an environment variable, `PUPPET_CLIENT_VERSION` to allow
specifying the starting Puppet version for acceptance tests. This
provides a flexible way to run tests of Puppet from any prior version to
the latest.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -91,7 +91,7 @@ end
 
 def server_opts
   {
-    :master => {:autosign => true},
+    :master => {:autosign => true, :dns_alt_names => master},
   }
 end
 
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -45,7 +45,7 @@ unless ENV['BEAKER_provision'] == 'no'
     master['puppetservice'] = 'puppetserver'
     master['puppetserver-confdir'] = '/etc/puppetlabs/puppetserver/conf.d'
     master['type'] = 'aio'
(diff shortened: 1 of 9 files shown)


MikaelSmith:
This is helpful for testing https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/81.

spec/spec_helper_acceptance.rb
```diff
@@ -45,7 +45,7 @@ def install_modules_on(host)
     master['puppetservice'] = 'puppetserver'
     master['puppetserver-confdir'] = '/etc/puppetlabs/puppetserver/conf.d'
     master['type'] = 'aio'
-    install_puppet_agent_on master, {:puppet_agent_version => '1.2.1'}
+    install_puppet_agent_on master, {}
```
highb: Is the `{}` necessary, since it's the default on the method signature?

MikaelSmith: Yes, because the description did not match the actual call signature. See https://github.com/puppetlabs/beaker/pull/1080.

highb: Interesting. :+1: , then.

PR #92, opened 2016-03-14
(maint) Update supported platforms

MikaelSmith:
Updates supported platforms for new package releases, AIX PowerPC 8
hardware, and platforms that were added but not noted as supported.

Note that specific upgrade scenarios will still depend on Puppet being
installed from packages prior to upgrade. `puppet-agent` packages should
exist for all supported platforms.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -19,7 +19,7 @@
 
 ## Overview
 
-A module for upgrading Puppet 3.8 agents to puppet-agent in Puppet Collection 1 (i.e., Puppet 4).
+A module for upgrading Puppet 3 agents to puppet-agent in Puppet Collection 1 (i.e., Puppet 4).
 
 ## Module Description
 
@@ -37,7 +37,9 @@ The puppet_agent module installs the Puppet Collection 1 repo (on systems that s
 
 ### Setup Requirements
 
-Your agents must be running Puppet 3.8 with `stringify_facts` set to 'false'. Agents should already be pointed at a master running Puppet Server 2.1 or greater, and thus successfully applying catalogs compiled with the Puppet 4 language.
+Your agents must be running Puppet 3 with `stringify_facts` set to 'false'. Agents should already be pointed at a master running Puppet Server 2.1 or greater, and thus successfully applying catalogs compiled with the Puppet 4 language.
+
+Puppet 3.7 with future parser is required to compile this module, meaning it can be applied to masterless Puppet 3.7+, or earlier Puppet 3 agents connecting to a Puppet 3.7+ master.
 
 ### Beginning with puppet_agent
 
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -44,8 +44,6 @@
     } elsif $::operatingsystem == 'aix' and $::architecture =~ /PowerPC_POWER[5,6,7]/ {
       $aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
       $_package_file_name = "${puppet_agent::package_name}-${puppet_agent::params::master_agent_version}-1.aix${aix_ver_number}.ppc.rpm"
-    } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /10\.[9,10,11]/ {
```
MikaelSmith: This section was redundant, see lines 42-43 above.

highb: Hm, must have been a bad re-base.

manifests/params.pp
```diff
@@ -67,6 +67,12 @@
     }
   }
 
+  # Treat Amazon Linux just like Enterprise Linux 6
+  $pe_repo_dir = ($::operatingsystem == 'Amazon') ? {
```
highb: :+1: to keeping the platform variables contained to `params.pp`

highb:
:+1: sane changes to me, travis is passing. We just need to ensure that we test the module out on AWS AMI images at some point.

README.markdown
```diff
@@ -116,6 +112,7 @@ Specifically in the 1.1.0 Release:
 * For Windows, you must trigger an agent run after upgrading so that Puppet can create the necessary directory structures.
 * Upgrading from 2015.2.x to 2015.3.x is not yet supported.
 * Solaris 11 is not yet supported.
+* AIX package names are based on PowerPC architecture version. PowerPC 8 is not yet supported.
```
MikaelSmith: We could fix this by dropping the exclusion of PowerPC 8. Happy to do that here or another PR.

MikaelSmith:
I tested the Amazon AMI manually for OSS. It's a challenge to setup, they don't provide `ruby(selinux)`, and the packages assume Ruby 1.8, so you have to mess with the environment a lot and manually install from RPMs. It also won't upgrade from Puppet3/Facter2, which are the Amazon-provided packages. So it doesn't really work for OSS, but that's not new.

I expect PE works, but that requires setting up a PE master in AWS to test against.

README.markdown
highb: Yeah, either is fine to me. Maybe lets get it in another PR so we don't overload this one.

highb:
:+1: Let's just make another PR for the Power8 change.

PR #93, opened 2016-03-17
(PE-12299) Add Solaris 11 support

MikaelSmith:
This commit plumbs the puppet agent module to add Solaris 11 support.

This commit should be fairly consistent with how pe_repo handles
upgrading agents on Solaris 11 by updating the repo in
`/etc/puppetlabs/installer/solaris.repo` with the new puppet-agent
package and installing puppet-agent from there

Lots of things in this commit are less than ideal but the current state
of how Solaris 11 packaging handles files through
uninstallation/installation.

Since Solaris 11 will not uninstall packages if dependant packages are
also installed so the packages must be removed in a specific order or
else the puppet run will fail.

When uninstalling the pe-* packages, Solaris 11 decides to completely
remove everything in `/etc/puppetlabs/` which is why this commit has to
go through and copy it to a temporary location during the upgrade.
However to install puppet-agent, the package needs to be in the repo
that lives in `/etc/puppetlabs/installer/solaris.repo` so the that repo
needs to be put back into place. After installing the puppet-agent
package it overwrites the existing configuration files with the default
configuration files. After installing the puppet-agent module the
/etc/puppetlabs directory is again copied into place so that the user
doesn't lose the ability to talk to the master server or any other
configurations that live in /etc/puppetlabs.

When upgrading Solaris 11 agents, the services are in a maintenance
state where Puppet is unable to start them. Use a script that waits for
the currently running agent run to stop and then calls puppet to ensure
that puppet and mcollective are running so that users do not have to
login to the nodes to perform these steps by hand.

This behavior is currently what windows is doing to start puppet and
mcollective agents when upgrading with the puppet_agent module.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -56,11 +56,25 @@ class puppet_agent (
 
     if $::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10' {
       $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.sles10.${::architecture}.rpm"
-    } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '10' {
+    } elsif $::operatingsystem == 'Solaris' {
       if $arch =~ /^sun4[uv]$/ {
-        $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.sparc.pkg.gz"
+        if $::operatingsystemmajrelease == '10' {
+          $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.sparc.pkg.gz"
+        } elsif $::operatingsystemmajrelease == '11' {
+          # Strip letters from development builds.
+          $_version_without_letters = regsubst($_package_version, /[a-zA-Z]/, '', 'G')
+          $_solaris_version = regsubst($_version_without_letters, /(^-|-$)/, '', 'G')
+          $_package_file_name = "${puppet_agent::package_name}@${_solaris_version},5.11-1.sparc.p5p"
+        }
       } else {
-        $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.i386.pkg.gz"
+        if $::operatingsystemmajrelease == '10' {
+          $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.i386.pkg.gz"
+        } elsif $::operatingsystemmajrelease == '11' {
+          # Strip letters from development builds.
+          $_version_without_letters = regsubst($_package_version, /[a-zA-Z]/, '', 'G')
+          $_solaris_version = regsubst($_version_without_letters, /(^-|-$)/, '', 'G')
+          $_package_file_name = "${puppet_agent::package_name}@${_solaris_version},5.11-1.i386.p5p"
+        }
       }
     } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /10\.[9,10,11]/ {
       $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.osx${$::macosx_productversion_major}.dmg"
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -16,7 +16,10 @@
 # [package_name]
 #   The package to upgrade to, i.e. `puppet-agent`.
 # [package_version]
-#   The package version to upgrade to. Defaults to undef, meaning only upgrade from Puppet < 4.0.
+#   The package version to upgrade to. When upgrading from Puppet < 4.0, defaults to
+#   the puppet master's latest supported version if compiled with a PE master or
+#   undef otherwise (meaning get the latest Open Source release). Explicitly specify
+#   a version to upgrade from puppet-agent packages (implying Puppet >= 4.0).
 # [service_names]
 #   An array of services to start, normally `puppet` and `mcollective`.
 #   None will be started if the array is empty.
@@ -28,22 +31,16 @@ class puppet_agent (
   $collection      = 'PC1',
   $is_pe           = $::puppet_agent::params::_is_pe,
   $package_name    = $::puppet_agent::params::package_name,
-  $package_version = undef,
+  $package_version = $::puppet_agent::params::package_version,
   $service_names   = $::puppet_agent::params::service_names,
   $source          = $::puppet_agent::params::_source,
 ) inherits ::puppet_agent::params {
 
   validate_re($arch, ['^x86$','^x64$','^i386$','^i86pc$','^amd64$','^x86_64$','^power$','^sun4[uv]$','PowerPC_POWER'])
 
-  if ($package_version == undef and $puppet_agent::params::master_agent_version != undef and versioncmp("${::clientversion}", '4.0.0') < 0) {
-    $_package_version = $puppet_agent::params::master_agent_version
-  } else {
-    $_package_version = $package_version
-  }
-
-  if $_package_version == undef and versioncmp("${::clientversion}", '4.0.0') >= 0 {
+  if $package_version == undef and versioncmp("${::clientversion}", '4.0.0') >= 0 {
     info('puppet_agent performs no actions if a package_version is not specified on Puppet 4')
-  } elsif $_package_version == undef and $is_pe {
+  } elsif $package_version == undef and $is_pe {
     info("puppet_agent performs no actions if the master's agent version cannot be determed on PE 3.x")
   } else {
     if $package_version != undef and $package_version !~ /^\d+\.\d+\.\d+([.-]?\d*|\.\d+\.g[0-9a-f]+)$/ {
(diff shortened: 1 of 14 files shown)


MikaelSmith:
This looks pretty good, planning to do some testing tomorrow.

templates/solaris_start_puppet.sh.erb
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+puppet_pid="<%= @puppet_agent_pid %>"
```
MikaelSmith: An improvement was made to the Windows script to pass in the PID (so the template isn't regenerated on each run). I'll do that here.

MikaelSmith:
Updated passing PID, but it appears to have some ordering issues in prepare. Working on fixing them.

MikaelSmith:
Tested and it looks good, this should be ready to go. @demophoon @highb

MikaelSmith:
Trying again. This needs to be re-tested on Solaris 11.

MikaelSmith:
Solaris 11 upgrades from 3.8 still appear to work.

highb:
Arg, this needs one more rebase. >_>

MikaelSmith:
Rebased.

highb:
Tested 3.8 -> 2016.1 PE upgrades with this change on a Sol11 box, it passed. Now running 2015.3.3 -> 2016.1.

highb:
Hm, seems that we still aren't installing the transition module:
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Invalid resource type transition at /etc/puppetlabs/code/environments/production/modules/puppet_agent/manifests/install/remove_packages.pp:152:9 on node s2f7gzq2qhruma5.delivery.puppetlabs.net
```

MikaelSmith:
But only on 215.3.3 -> 2016.1 upgrades?

highb:
Seems so. I'm re-examining what we're doing in https://github.com/puppetlabs/pe_acceptance_tests/blob/2016.2.x/setup/agent_upgrade.rb to try to figure out what might be happening.

highb:
Oh, r10k doesn't manage module dependencies. https://github.com/puppetlabs/pe_acceptance_tests-control/blob/production/Puppetfile needs to be updated. Going to put up a PR.

highb:
Hm, looks like something is going wrong with the Transition resource during 2015.3.3 to 2016.1.x upgrade:
```
      Notice: /Stage[main]/Puppet_agent::Osfamily::Solaris/Exec[puppet_agent copy packages]/returns: executed successfully
      Notice: /Stage[main]/Puppet_agent::Osfamily::Solaris/File[/tmp/puppet_agent/]/ensure: created
      Notice: /Stage[main]/Puppet_agent::Osfamily::Solaris/Exec[puppet_agent backup /etc/puppetlabs/]/returns: executed successfully
      Notice: /Package[puppet-agent]/ensure: removed
      Info: Creating state file /opt/puppetlabs/puppet/cache/state/state.yaml
      Error: No such file or directory @ dir_s_rmdir - /opt/puppetlabs/puppet/cache/state/state.yaml20160328-2456-gj7vgv.lock
      Notice: /Stage[main]/Puppet_agent::Install::Remove_packages/Notify[using puppetlabs-transition to remove puppet-agent: Solaris does not support versionable]: Dependency Transition[remove puppet-agent] has failures: true
      Notice: /Stage[main]/Puppet_agent::Install/Exec[puppet_agent restore /etc/puppetlabs]: Dependency Transition[remove puppet-agent] has failures: true
      Error: /Stage[main]/Puppet_agent::Install::Remove_packages/Transition[remove puppet-agent]/enable: change from enabled to transition Package[puppet-agent] to state {"ensure"=>"absent", "adminfile"=>"/opt/puppetlabs/packages/solaris-noask"} failed: No such file or directory @ dir_s_rmdir - /opt/puppetlabs/puppet/cache/state/state.yaml20160328-2456-gj7vgv.lock
      Warning: /Stage[main]/Puppet_agent::Install::Remove_packages/Notify[using puppetlabs-transition to remove puppet-agent: Solaris does not support versionable]: Skipping because of failed dependencies
      Warning: /Stage[main]/Puppet_agent::Install/Exec[puppet_agent restore /etc/puppetlabs]: Skipping because of failed dependencies
      Notice: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]: Dependency Transition[remove puppet-agent] has failures: true
      Warning: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]: Skipping because of failed dependencies
      Notice: /Stage[main]/Puppet_agent::Install/Exec[puppet_agent post-install restore /etc/puppetlabs]: Dependency Transition[remove puppet-agent] has failures: true
      Warning: /Stage[main]/Puppet_agent::Install/Exec[puppet_agent post-install restore /etc/puppetlabs]: Skipping because of failed dependencies
      Info: Class[Puppet_agent]: Unscheduling all events on Class[Puppet_agent]
```

MikaelSmith:
Only on Solaris 11?

highb:
@MikaelSmith I'm checking out Sol10 now.

highb:
Yeah, it appears to only be a problem on Solaris 11: https://gist.github.com/highb/a4f0179af9149ed5a34ba6dfdda157cb

MikaelSmith:
I think I've fixed it, running through some tests to be sure.

MikaelSmith:
@highb passed acceptance runs testing PE 3.8 -> 2015.3.3 and 2015.2.3 -> 2015.3.3. I think this is ready.

manifests/init.pp
```diff
@@ -55,18 +52,32 @@
...
+        }
       } else {
-        $_package_file_name = "${puppet_agent::package_name}-${_package_version}-1.i386.pkg.gz"
+        if $::operatingsystemmajrelease == '10' {
+          $_package_file_name = "${puppet_agent::package_name}-${package_version}-1.i386.pkg.gz"
+        } elsif $::operatingsystemmajrelease == '11' {
+          # Strip letters from development builds.
+          $_version_without_letters = regsubst($package_version, /[a-zA-Z]/, '', 'G')
+          $_solaris_version = regsubst($_version_without_letters, /(^-|-$)/, '', 'G')
+          $_package_file_name = "${puppet_agent::package_name}@${_solaris_version},5.11-1.i386.p5p"
```
highb: Couldn't the i386 vs sparc part be set in a variable? Then we wouldn't need to duplicate all the regex for both platforms.

manifests/params.pp
```diff
@@ -80,6 +82,12 @@
     default => undef,
   }
 
+  if ($master_agent_version != undef and versioncmp("${::clientversion}", '4.0.0') < 0) {
```
highb: Moving this to `params.pp` is a great improvement for readability!

manifests/init.pp
MikaelSmith: Yeah, probably.

MikaelSmith: Does it also make sense for Solaris 10 packages?

highb: Well, the arch is the same between 10 and 11, right? There could be a `$_arch` that is set to either `sparc` or `i386` depending on the `sun4[uv]` regex, and then we just use `$_arch` regardless of wether it is on 10 or 11.

MikaelSmith: ping @demophoon where did needing to strip letters out of development builds come from? If I look at current Solaris 11 package names, I see things like `puppet-agent@1.4.1.36.778531,5.11-1.i386.p5p`.

demophoon: @MikaelSmith That is how our packaging works with solaris package names. That was taken directly out of Vanagon here: https://github.com/puppetlabs/vanagon/blob/5236eefe97963f3cfc312bbef1e3e1b388d6ffc4/lib/vanagon/platform/solaris_11.rb#L100-L107

Solaris 11 seems to be the only platform that has this version naming oddity.

MikaelSmith: Oh, I see. Ok, I'll put it back to Solaris 11 only.

MikaelSmith:
Updated.

MikaelSmith:
Ping @highb

manifests/init.pp
```diff
@@ -55,18 +52,27 @@
...
+      }
+
+      if $::operatingsystemmajrelease == '10' {
+        $_package_file_name = "${puppet_agent::package_name}-${package_version}-1.${pkg_arch}.pkg.gz"
+      } elsif $::operatingsystemmajrelease == '11' {
+        # Strip letters from development builds. Unique to Solaris 11 packaging.
+        $_version_without_letters = regsubst($package_version, /[a-zA-Z]/, '', 'G')
+        $_solaris_version = regsubst($_version_without_letters, /(^-|-$)/, '', 'G')
+
+        $_package_file_name = "${puppet_agent::package_name}@${_solaris_version},5.11-1.${pkg_arch}.p5p"
```
highb: Yep, this is what I had in mind! Glad it worked!

PR #96, opened 2016-03-25
(maint) Re-enable open-source upgrades from 3.x

MikaelSmith:
The prior commit to enable puppet-agent to puppet-agent upgrades set an
expectation that a target version would always be required, either from
the master or explicitly specified as a parameter. That broke the
previous open-source upgrade behavior when upgrading from Puppet 3.x.

Restore that behavior by allowing the target version to be unspecified
if not on Puppet Enterprise and upgrading from Puppet 3.x. Also add the
new dependency puppetlabs-transition for acceptance testing.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -84,7 +84,6 @@ class puppet_agent (
 
     class { '::puppet_agent::prepare':
       package_file_name => $_package_file_name,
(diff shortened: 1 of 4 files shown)


highb:
:+1:

PR #97, opened 2016-03-29
Remove pluginsync setting if upgrading to 1.4.x

tphoney:


manifests/prepare/puppet_config.pp
```diff
@@ -26,7 +26,14 @@
       # Deprecated for global config
       'config_version', 'manifest', 'modulepath',
       # Settings that should be reset to defaults
-      'disable_warnings', 'vardir', 'rundir', 'libdir', 'confdir', 'ssldir', 'classfile'].each |$setting| {
+      'disable_warnings', 'vardir', 'rundir', 'libdir', 'confdir', 'ssldir', 'classfile']
```
MikaelSmith: Indentation looks off.

manifests/prepare/puppet_config.pp
```diff
@@ -26,7 +26,14 @@
...
+      # When upgrading to 1.4.x or later remove pluginsync
+      if versioncmp($puppet_agent::params::master_agent_version, '1.4.0') > 0 {
```
MikaelSmith: Generally a good approach, but we should base this off the target version. That means adding a new parameter to this class to pass _package_version in from init.pp

highb:
This will only cleanup the settings on upgrades from <= 3.8 (see `prepare.pp`)

MikaelSmith:
Oh yeah... might need another class to cover new deprecations.

manifests/prepare/puppet_config.pp
```diff
@@ -26,7 +26,14 @@
       # Deprecated for global config
       'config_version', 'manifest', 'modulepath',
       # Settings that should be reset to defaults
-      'disable_warnings', 'vardir', 'rundir', 'libdir', 'confdir', 'ssldir', 'classfile'].each |$setting| {
+      'disable_warnings', 'vardir', 'rundir', 'libdir', 'confdir', 'ssldir', 'classfile']
+    # When upgrading to 1.4.x or later remove pluginsync
+    if versioncmp($::puppet_agent::package_version, '1.4.0') >= 0 {
```
MikaelSmith: This only works if we move https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/init.pp#L38-L42 to params.pp.

MikaelSmith: We should also handle undef. If we get this far with undef, it means we're updating to FOSS latest, so we should also remove pluginsync in that case.

tphoney:
@MikaelSmith @highb I made the suggested changes. Sorry for the ping pong nature of this.

MikaelSmith:
This needs a rebase now, and you could remove the WIP note.

MikaelSmith:
:+1:

MikaelSmith:
Oh, should probably add a spec. I'll add it and put up a new PR.

MikaelSmith:
Put up a new PR at https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/101.

PR #99, opened 2016-04-05
(maint) Fix a dependency cycle introduced on Windows by Sol11 work

MikaelSmith:
Fixes a dependency cycle introduced by Solaris 11 work. Solaris 11
requires managing config files before preparing packages, because part
of preparing the package is saving /etc/puppetlabs where the configs are
stored.

On Windows, a dependency was introduced because the package is stored to
`C:/ProgramData/Puppet Labs/...`, and config files are also stored
there. So creating an order of configs before package prepare caused a
requirement that C:/ProgramData/Puppet Labs exists both before and after
managing config files. Fix that dependency cycle by mananging
`C:/ProgramData/Puppet Labs` earlier.

### DIFF
--- a/manifests/prepare.pp
+++ b/manifests/prepare.pp
@@ -17,6 +17,7 @@ class puppet_agent::prepare(
   include puppet_agent::params
   $_windows_client = downcase($::osfamily) == 'windows'
   if $_windows_client {
+
     File{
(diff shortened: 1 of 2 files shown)


MikaelSmith:
@highb

highb:
:+1:

Issue #98, opened 2016-04-05
Unknown function: 'pe_compiling_server_aio_build'


olevole:
puppet-agent                    1.4.1-1trusty
puppetserver                     2.3.1-1puppetlabs1
Ubuntu 14.04.4 LTS

Troubles with 1.1.0 and latest code from https://github.com/puppetlabs/puppetlabs-puppet_agent.git i've got error:
~~~
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Unknown function: 'pe_compiling_server_aio_build'. at /etc/puppetlabs/code/environments/production/modules/public/puppet_agent/manifests/params.pp:81:16 on node 
~~~

MikaelSmith:
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/params.pp#L81 probably needs to be guarded on the function existing. I'll look at fixing that up in a bit.

MikaelSmith:
Are you using a PE master? It should be guarded by `is_pe`.

olevole:
Look's like it was a one-time problem - after synchronization .rb files by client problem went away. Thank you for your response, ive close this issue.

MikaelSmith:
Hmm, I'm curious how it got into that state. Glad it worked now though.

olevole:
it is observed on all of my nodes when moving from puppet 3  to puppet 4: http://pastebin.com/iK2NB9gU . So, it neccessary to run 
~~~
puppet agent -t
~~~

twice for me.

MikaelSmith:
Are you also changing the master you point to when you do the upgrade?

olevole:
No, master was originally on 4 version. Somehow (by mistake) the clients  was  on puppet3 and this installation (puppet3 client <-> puppet4 ) worked successfully until recently.

MikaelSmith:
Put up a PR to try and avoid that problem anyway.

PR #100, opened 2016-04-05
(maint) Fix issue #98, don't require `pe_compiling_server_aio_build`

MikaelSmith:
Compilation should not require `pe_compiling_server_aio_build`. Skip
calling it if not defined even if `is_pe` is true.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -77,9 +77,10 @@ class puppet_agent::params {
 
   # The aio puppet-agent version currently installed on the compiling master
   # (only used in PE)
-  $master_agent_version = $_is_pe ? {
-    true    => pe_compiling_server_aio_build(),
-    default => undef,
+  if ($_is_pe and is_function_available('pe_compiling_server_aio_build')) {
+    $master_agent_version = pe_compiling_server_aio_build()
+  } else {
+    $master_agent_version = undef
   }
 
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -77,9 +77,8 @@
 
   # The aio puppet-agent version currently installed on the compiling master
   # (only used in PE)
-  $master_agent_version = $_is_pe ? {
-    true    => pe_compiling_server_aio_build(),
-    default => undef,
+  if ($_is_pe && is_function_available('pe_compiling_server_aio_build')) {
+    $master_agent_version = pe_compiling_server_aio_build()
```
highb: Can we add an else to explicitly set it to `undef`? We're going to get warnings all over the place otherwise.

MikaelSmith: Ok. I also want to add a spec test.

MikaelSmith:
Updated.

highb:
:+1: once specs pass.

MikaelSmith:
Specs are proving tricky.

MikaelSmith:
I think I got all the tests working. Got a little complicated.

highb:
Heh, that did get a bit complicated. Still, the fundamental change make sense and the specs mostly cover its behavior so :+1:

PR #103, opened 2016-04-06
(maint) Fix spec that should've skipped Solaris

MikaelSmith:
Typo in previous spec meant it only skipped SLES (as a PE-only
platform). It must also skip Solaris.


highb:
:+1:

PR #106, opened 2016-04-13
(PE-11531) Don't remove packages if package_version undef

highb:
Prior to this commit we were removing the puppet-agent package
if the `package_version` happened to be unset, due to the
behavior of `versioncmp`. This is obviously not what we want to
have happen.
This commit updates all of the calls that compare `package_version`
with `aio_agent_version` with `versioncmp` to be gated on
`package_version` not being undef, thus preventing the packages
from being removed on every run.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -77,7 +77,7 @@ class puppet_agent::install(
 
   if $::osfamily == 'windows' {
     # Prevent re-running the batch install
-    if versioncmp("${::aio_agent_version}", "${package_version}") < 0 {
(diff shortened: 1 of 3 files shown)


highb:
As discussed in channel, `versioncmp` returns -1 if the second argument is nil:
```
[root@azvyfad4cb0gnv2 ~]# puppet apply -e 'warning(versioncmp("1.0.1", nil))'
Warning: Scope(Class[main]): -1
Notice: Compiled catalog for azvyfad4cb0gnv2.delivery.puppetlabs.net in environment production in 0.03 seconds
```

MikaelSmith:
👍

PR #107, opened 2016-04-14
(maint) Fix handling dev versions on Solaris 11

MikaelSmith:
Version numbers for development packages on Solaris 11 need to be
munged, which happened for the file name. However Solaris 11
puppet-agent upgrades use the version number, which was not correctly
munged. Fix that, and add testing for development version numbers.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -63,10 +63,11 @@ class puppet_agent (
         $_package_file_name = "${puppet_agent::package_name}-${package_version}-1.${pkg_arch}.pkg.gz"
       } elsif $::operatingsystemmajrelease == '11' {
         # Strip letters from development builds. Unique to Solaris 11 packaging.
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -90,13 +91,20 @@
       $_package_file_name = undef
     }
 
+    # Allow for normalizing package_version for the package provider via _package_version.
+    # This only needs to be passed through to install, as elsewhere we want to
+    # use the full version string for comparisons.
+    if !$_package_version {
```
highb: Instead of relying on puppet's falsiness, maybe we should check `$_package_version != undef`?

MikaelSmith: Sure.

highb:
:+1: Once CI passes. I ran this through a install.rb/agent_upgrade.rb beaker run, and followed up with acceptance/tests/ and it successfully upgraded.

PR #108, opened 2016-04-14
(PE-11531) Centralize check for aio upgrade

highb:
Prior to this commit we had the same AIO upgrade expression scattered
in 3 different places with slightly different syntax. Also, we did not
account for the fact that `aio_agent_version` is a PE only fact.
This commit creates a `$aio_upgrade_required` var in the main class
that will 1) always be true if not PE and `package_version` is set 2) be
true if `aio_agent_version` and `package_version` are set and
the `aio_agent_version` is less than (-1) the `package_version`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -47,6 +47,10 @@ class puppet_agent (
       fail("invalid version ${package_version} requested")
     }
 
+    $aio_upgrade_required = ($is_pe == false and $package_version != undef) or
+      ($::aio_agent_version != undef and $package_version != undef and
+        versioncmp("${::aio_agent_version}", "${package_version}") < 0)
+
     if $::architecture == 'x86' and $arch == 'x64' {
       fail('Unable to install x64 on a x86 system')
     }
--- a/spec/classes/puppet_agent_windows_install_spec.rb
+++ b/spec/classes/puppet_agent_windows_install_spec.rb
@@ -44,27 +44,39 @@ RSpec.describe 'puppet_agent' do
 
         let(:facts) { facts.merge({:is_pe => true}) }
 
-        it {
-          is_expected.to contain_file('C:\tmp\install_puppet.bat').with_content(
-            %r[#{Regexp.escape("msiexec.exe /qn /norestart /i \"#{values[:appdata]}\\Puppetlabs\\packages\\puppet-agent-#{values[:expect_arch]}.msi\"")}])
-        }
+        if Puppet.version < '4.0.0'
+          context 'with aio_agent_version unset' do
+            let(:facts) { facts.merge({:is_pe => true}) }
 
-        context 'with up to date aio_agent_version matching server' do
-          let(:facts) { facts.merge({
-            :is_pe => true,
-            :aio_agent_version => package_version
-          })}
-
-          it { is_expected.not_to contain_class('puppet_agent::windows::install') }
+            it { is_expected.to contain_class('puppet_agent::windows::install') }
+            it { is_expected.to contain_file('C:\tmp\install_puppet.bat').with_content(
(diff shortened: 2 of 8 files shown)


highb:
Ran this through a basic pe_acceptance install/agent_upgrade, and then checked the storeconfigs test and it passed!

manifests/init.pp
```diff
@@ -47,6 +47,10 @@
       fail("invalid version ${package_version} requested")
     }
 
+    $aio_upgrade_required = ($is_pe == false and $package_version != undef) or
+      ($::aio_agent_version != undef and $package_version != undef and
```
MikaelSmith: We still don't know why aio_agent_version might not be set?

Eventually we can drop the is_pe check, when we add a custom fact to this module.

spec/classes/puppet_agent_windows_install_spec.rb
```diff
@@ -44,10 +44,17 @@
 
         let(:facts) { facts.merge({:is_pe => true}) }
 
-        it {
-          is_expected.to contain_file('C:\tmp\install_puppet.bat').with_content(
-            %r[#{Regexp.escape("msiexec.exe /qn /norestart /i \"#{values[:appdata]}\\Puppetlabs\\packages\\puppet-agent-#{values[:expect_arch]}.msi\"")}])
-        }
+        context 'with out of date aio_agent_version' do
```
MikaelSmith: This looks like it could be rolled into the other context below (line 68).

Also, we could keep a context where aio_agent_version isn't set to ensure upgrades don't happen.

MikaelSmith:
A few nitpicks on tests, otherwise looks good.

manifests/init.pp
highb: Yeah I'm not sure what is happening there and yeah part of the reason I decided to centralize this check is so we can easily update it later if we port the fact over.

highb:
I'm so confused. In the process of making these changes and testing with PUPPET_VERSION='~> 4.0.0', it seems that Solaris 11 stopped being versionable? I'm really confused as to how that happened, seeing as it looks like the current version of the module is able to update just fine.

MikaelSmith:
You're picking up the default provider of whatever system you run specs on (Mac presumably). The Travis failure looks like a different issue.

I've previously done something like https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/spec/classes/puppet_agent_osfamily_solaris_spec.rb#L58-L60 to remedy getting the wrong provider.

highb:
Wow ok, thanks. Yeah I've been working around the issue seen in Travis, but I ran into the versionable issue in the process.

highb:
Tested this out on a 3.8.4->2015.3.x upgrade on the platforms below and the upgrade succeeded. I also ran the puppetdb_storeconfigs.rb test and that appears to be passing for these platforms now.
```
 sol-10-agent:
    hypervisor: vmpooler
    platform: solaris-10-i386
    template: solaris-10-x86_64
    roles:
    - frictionless
    - agent
  sol-11-agent:
    hypervisor: vmpooler
    platform: solaris-11-i386
    template: solaris-11-x86_64
    roles:
    - frictionless
    - agent
  sles-11-agent:
    hypervisor: vmpooler
    platform: sles-11-x86_64
    template: sles-11-x86_64
    roles:
    - frictionless
    - agent
  osx-109-agent:
    hypervisor: vmpooler
    platform: osx-109-x86_64
    template: osx-109-x86_64
    roles:
    - frictionless
    - agent
```

Should I also manually run a 2015.2.3 upgrade or perhaps we should just merge into CI and kick the jobs to see what else falls out?

MikaelSmith:
I'm ok relying on CI, that's why we have it.

PR #111, opened 2016-04-19
(RE-7037) Use updated gpg key

underscorgan:
We have an updated gpg key that won't expire until January. We should
start using that.


MikaelSmith:
👍

PR #112, opened 2016-04-20
(PE-15036) Fix Windows permission inheritance

highb:
Prior to this commit there was an issue where the inherit permission
value on the `client_datadir` folder was being lost, causing the folder
to be inaccessible to the MCO/PCP daemons.

This commit ensures that directory has the proper inherit permission,
if it appears that it does not have it. For more details: PE-15036

### DIFF
--- a/manifests/windows/install.pp
+++ b/manifests/windows/install.pp
@@ -56,4 +56,11 @@ class puppet_agent::windows::install(
     command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${_installbat}\" ${::puppet_agent_pid}",
     path    => $::path,
   }
+
+  # PUP-5480/PE-15037 Cache dir loses inheritable SYSTEM perms
+  exec { 'fix inheritable SYSTEM perms':
+    command => "${::system32}\\icacls.exe \"${::puppet_client_datadir}\" /grant \"SYSTEM:(OI)(CI)(F)\"",
+    unless  => "${::system32}\\icacls.exe \"${::puppet_client_datadir}\" | findstr \"SYSTEM:(OI)(CI)(F)\"",
+    require => Exec['install_puppet.bat'],
+  }
 }
--- a/lib/facter/settings.rb
+++ b/lib/facter/settings.rb
@@ -40,6 +40,12 @@ Facter.add('puppet_confdir') do
   end
 end
 
+Facter.add('puppet_client_datadir') do
+  setcode do
+    Puppet.settings['client_datadir']
+  end
+end
+
 Facter.add('mco_confdir') do
   setcode do
     File.expand_path(File.join(Puppet.settings['confdir'],'../../mcollective/etc'))
--- a/spec/classes/puppet_agent_windows_install_spec.rb
+++ b/spec/classes/puppet_agent_windows_install_spec.rb
@@ -52,6 +52,7 @@ RSpec.describe 'puppet_agent' do
             it { is_expected.to contain_file('C:\tmp\install_puppet.bat').with_content(
                 %r[#{Regexp.escape("msiexec.exe /qn /norestart /i \"#{values[:appdata]}\\Puppetlabs\\packages\\puppet-agent-#{values[:expect_arch]}.msi\"")}])
             }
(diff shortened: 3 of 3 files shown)


highb:
I've tested this with `r10k_puppet_agent_fork=highb r10k_puppet_agent_ref=fix/master/pe-15037_fix_inhieritable_permissions_windows ./frankenbuilder -v 2015.3.2 --upgrade_from 2015.2.3 --tests /s/pe_acceptance_tests/setup/install.rb,/s/pe_acceptance_tests/setup/agent_upgrade.rb --vmpooler` and it passed!

```
      Test Suite: tests @ 2016-04-19 20:59:18 -0700

      - Host Configuration Summary -


              - Test Case Summary for suite 'tests' -
       Total Suite Time: 2193.59 seconds
      Average Test Time: 1096.79 seconds
              Attempted: 2
                 Passed: 2
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 2

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```

This should resolve the failing windows puppet_agent tests in CI.

manifests/windows/install.pp
```diff
@@ -56,4 +56,11 @@
     command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${_installbat}\" ${::puppet_agent_pid}",
     path    => $::path,
   }
+
+  # PE-15037 Cache dir loses inheritable SYSTEM perms
```
Iristyle: I would also note the original PUP ticket here for ref - https://tickets.puppetlabs.com/browse/PUP-5480

manifests/windows/install.pp
```diff
@@ -56,4 +56,11 @@
...
+  exec { 'fix inheritable SYSTEM perms':
+    command => "${::system32}\\icacls.exe ${::puppet_client_datadir} /grant \"SYSTEM:(OI)(CI)(F)\"",
```
Iristyle: On older versions of windows this path could contains spaces, so I think you want to wrap the path in double quotes to be safe like:

```
"${::system32}\\icacls.exe \"${::puppet_client_datadir}\" /grant \"SYSTEM:(OI)(CI)(F)\""
```

Iristyle:
I can't really comment on the tests, but everything looks good to me - and it passed - hurray!

highb:
@Iristyle Updated with the PUP ticket and quoted the paths. Thanks again for helping so much on this! :)

manifests/windows/install.pp
```diff
@@ -56,4 +56,11 @@
     command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${_installbat}\" ${::puppet_agent_pid}",
     path    => $::path,
   }
+
+  # PUP-5480/PE-15037 Cache dir loses inheritable SYSTEM perms
+  exec { 'fix inheritable SYSTEM perms':
+    command => "${::system32}\\icacls.exe \"${::puppet_client_datadir}\" /grant \"SYSTEM:(OI)(CI)(F)\"",
+    unless  => "${::system32}\\icacls.exe \"${::puppet_client_datadir}\" | findstr \"SYSTEM:(OI)(CI)(F)\"",
+    require => Exec['install_puppet.bat'],
```
MikaelSmith: Not sure why this requires install_puppet, but it shouldn't hurt because install_puppet won't do anything until Puppet exits.

MikaelSmith:
👍

manifests/windows/install.pp
highb: Yeah that was my thinking; we don't want it to run before the upgrade or during it, so I did this.

MikaelSmith: I'm not sure I follow... this will always run before the upgrade, because install_puppet.bat just waits until the current Puppet run finishes before triggering the MSI.

highb: Hm, right. So really this is sort of a race condition. :/

MikaelSmith: As in, the current version of Puppet may break the ACL settings on client_datadir again before we exit?

PR #113, opened 2016-04-20
(PE-11531) Fix upgrade issue with dev builds

highb:
Prior to this commit the module would always upgrade with dev
builds because they have a git SHA appended to the end, which
`versioncmp` considers to be greater than the version reported
by `aio_agent_version`, which lacks the extra git sha characters.
This commit updates the update check to strip the git sha when
comparing the `aio_agent_version` to the requested `packace_version`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -47,9 +47,16 @@ class puppet_agent (
       fail("invalid version ${package_version} requested")
     }
 
-    $aio_upgrade_required = ($is_pe == false and $package_version != undef) or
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Failing specs.

highb:
Gr, I thought I ran `rake test` on this. That's what I get for coding tired.

highb:
There, syntax passes locally. Sorry about that.

MikaelSmith:
Specs are getting complicated. Looks good though.

PR #115, opened 2016-04-22
(maint) Add maintainers

MikaelSmith:
@highb formatting ok? any others we should add?


highb:
👍 That's probably good enough for now.

PR #116, opened 2016-04-26
(MODULES-3304/PE-15256) Fix Windows 2008 upgrades

highb:
Prior to this commit the path being used to locate the local
prepared copy of the windows MSI had a bug where one area of
the module would use a forward slash, and the other would use a
backslash. Additionally, the install class did not depend on the
prepare step's file resource that is created and instead used the
`defined` function to check for the file.
This commit updates the slashes to be consistent and adds an
explicit require on the File resource that prepares the local
copy of the MSI.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -78,10 +78,12 @@ class puppet_agent::install(
   if $::osfamily == 'windows' {
     # Prevent re-running the batch install
     if $old_packages or $puppet_agent::aio_upgrade_required {
-      if $::puppet_agent::is_pe == true and empty($::puppet_agent::source) and defined(File["${::puppet_agent::params::local_packages_dir}/${package_file_name}"]) {
+      if $::puppet_agent::is_pe == true and empty($::puppet_agent::source) {
+        $local_package_file_path = windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}")
         class { 'puppet_agent::windows::install':
           package_file_name => $package_file_name,
-          source            => windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}"),
+          source            => $local_package_file_path,
+          require           => File[$local_package_file_path],
         }
       } else {
         class { 'puppet_agent::windows::install':
--- a/manifests/prepare/package.pp
+++ b/manifests/prepare/package.pp
@@ -30,7 +30,14 @@ class puppet_agent::prepare::package(
     file { $::puppet_agent::params::local_packages_dir:
       ensure => directory,
     }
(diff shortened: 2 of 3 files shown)


highb:
I have tested this with a Frankenbuild
```
r10k_puppet_agent_fork=highb r10k_puppet_agent_ref=testing ./frankenbuilder -v 2016.1.2-rc0-55-g5a542e9 --upgrade_from 2015.3.1 --tests /s/pe_acceptance_tests/setup/install.rb,/s/pe_acceptance_tests/setup/agent_upgrade.rb --agent windows-2008-x86_64 --vmpooler
```

MikaelSmith:
I'm still not a fan of using closed tickets for these fixes.

highb:
@MikaelSmith I could open FM ticket for the problem?

manifests/install.pp
```diff
@@ -75,13 +75,20 @@
     $_package_options = {}
   }
 
+  if $::osfamily =~ /windows/ {
+    $local_package_file_path = windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}")
+  } else {
+    $local_package_file_path = "${::puppet_agent::params::local_packages_dir}/${package_file_name}"
```
MikaelSmith: Is this branch used anywhere?

manifests/install.pp
```diff
@@ -75,13 +75,20 @@
     $_package_options = {}
   }
 
+  if $::osfamily =~ /windows/ {
```
MikaelSmith: Maybe do this in init and pass it to install and prepare::package?

highb: Yeah I was thinking about that. It seemed redundant to pass it in alongside the package_file_name, but the reduction in duplicated logic may make it worth it.

manifests/install.pp
highb: Hm, apparently not. It seems like it should be for the platforms w/o package managers.

manifests/install.pp
highb: OK I started plumbing out passing the `local_package_file_path` in from `init` and it is going to touch a lot more than this change.

MikaelSmith: Meaning you'd rather not do it? Or more work is needed?

highb: Meaning it would take more work, and I don't have time to tackle that right now.

highb:
Created MODULES-3304 for public visibility, amended it to the commit.

manifests/install.pp
MikaelSmith: Ok.

MikaelSmith:
👍

PR #119, opened 2016-05-23
(maint) Update metadata.json for Puppet 3.7

Iristyle:
- As part of the 1.2 release, the scope of upgrades was changed to include 3.7.  Makes sure that
   metadata.json reflects this.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -2,7 +2,7 @@
   "name": "puppetlabs-puppet_agent",
   "version": "1.2.0",
   "author": "puppetlabs",
-  "summary": "Upgrades Puppet 3.8 and All-In-One Puppet Agents",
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -2,7 +2,7 @@
   "name": "puppetlabs-puppet_agent",
   "version": "1.2.0",
   "author": "puppetlabs",
-  "summary": "Upgrades Puppet 3.8 and All-In-One Puppet Agents",
+  "summary": "Upgrades Puppet 3.7 and All-In-One Puppet Agents",
```
MikaelSmith: Seems like this could be clearer, as in supports `Puppet 3.7+`.

MikaelSmith: ping

Iristyle: Thanks - updated.

PR #124, opened 2016-06-06
(MODULES-3449) Stop Windows pxp-agent service

Iristyle:
- Previously, upgrades could fail as a result of the pxp-agent service
   holding file locks.  While we have been previously aware of the
   potential for this isssue, given pxp-agent runs under nssm.exe on
   Windows, prior investigation had shown this to not be an issue on
   silent installations (see PA-65).

   However, in practice, while upgrading from puppet-agent 1.4.2 to
   puppet-agent 1.5.0, a running pxp-agent service is preventing
   installs from proceeding as requested with a 1603 failure. NSSM.exe
   is not affiliated with the pxp-agent service in the MSI and therefore
   the Restart Manager has no understanding that it will be later
   shutdown and unlocked.

   The solution is to simply request a `net stop pxp-agent` prior to
   running the MSI.  It's very important that this be done *after* the
   puppet run responsible for triggering the upgrade has been performed.
   If the pxp-agent service shutdown occurs before, then the puppet run
   will typically turn it back on as part of managing the node with the
   Puppet Enterprise modules.

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -15,6 +15,14 @@ FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=
 echo %_task% | findstr "No tasks are running" >nul
 IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
 
+REM This *must* occur after Puppet Agent has finished applying its
+REM prior catalog which manages the pxp-agent service state. If not,
+REM the catalog includes the PE module which starts the service and
+REM sets its startup type, which prevents installs from proceeding.
+REM This may fail on agents without pxp-agent, but since this is not
+REM run interactively and the next command sets ERRORLEVEL, it's OK.
+net stop pxp-agent
+
 start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
 
 if exist %pid_path% del %pid_path%
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -23,7 +23,7 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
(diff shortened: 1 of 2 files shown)


templates/install_puppet.bat.erb
```diff
@@ -15,7 +15,13 @@ FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=
 echo %_task% | findstr "No tasks are running" >nul
 IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
 
-start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
+REM This *must* occur after Puppet Agent has finished applying its
+REM prior catalog which manages the pxp-agent service state.
```
glennsarti: Should you mention why it must occur if it's so important?

Iristyle: All set

glennsarti:
:+1: From me

templates/install_puppet.bat.erb
```diff
@@ -15,7 +15,15 @@ FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=
...
 IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
 
-start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*v "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
+REM This *must* occur after Puppet Agent has finished applying its
+REM prior catalog which manages the pxp-agent service state. If not,
+REM the catalog includes the PE module which starts the service and
+REM sets its startup type, which prevents installs from proceeding.
+REM This may fail on agents without pxp-agent, but since this is not
+REM run interactively and the next command sets ERRORLEVEL, it's OK.
+net stop pxp-agent
```
ferventcoder: Did you determine the issue? 

Also keep in mind that 3.x agents upgrading won't have this - so if it fails that should be ignored.

Iristyle: Yes, was able to fix the upgrade with just the `net stop pxp-agent` here.  Unfortunately, Restart Manager doesn't tell us what files are locked, though it's safe to assume its {{nssm.exe}}.  Still don't know why this behavior deviates from what we observed in PA-65 though.

Your note about 3.x is in the comment block already.

PR #125, opened 2016-06-10
Adds install_path for windows. Also fixes puppet service startup

Renelast:
This patch allows you to specify an installation path for the windows msi installation (We had puppet installed in a different path on our servers)

We also noticed that, after upgrading, the puppet service would not always be started by the msi so I also added an "sc start puppet" to the upgrade script to ensure the service is running after upgrading.

I tested this on live systems running windows 2008R2 and 2012R2

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -27,6 +27,5 @@ start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_l
 
 if exist %pid_path% del %pid_path%
 
-
 :End
 ENDLOCAL
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -23,9 +23,7 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
-start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>" <% unless @install_dir.empty? -%>INSTALDIR="<%= @install_dir %>"<% end -%>
-
-start /wait sc start puppet
+start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>" <% unless @install_dir.empty? -%>INSTALLDIR="<%= @install_dir %>"<% end -%>
 
 if exist %pid_path% del %pid_path%
 
--- a/README.markdown
+++ b/README.markdown
@@ -131,6 +131,12 @@ An array of services to start, normally `puppet` and `mcollective`. If the array
 
 Alternate source from which you wish to download the latest version of Puppet.
 
+####`install_dir`
+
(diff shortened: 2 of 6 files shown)


MikaelSmith:
@Iristyle would you mind taking a look at this? The service not starting after upgrade is strange, the MSI should be doing that (unless you pass an option to disable it).

Iristyle:
An `sc start puppet` should never be required.  The MSI is responsible for putting the Puppet service into the state that was requested by the install initially based on the `PUPPET_AGENT_STARTUP_MODE` MSI property, which defaults to `Automatic`.  https://docs.puppet.com/puppet/latest/reference/install_windows.html#puppetagentstartupmode

@Renelast I would suggest in the case of needing the `sc start puppet` in your environment, that you take a closer look at your MSI installation logs.  There should be some detail around the startup mode there that we can take a look at.  If you ever previously installed with a `Manual` startup mode, that property is "remembered" in the registry for subsequent upgrades.

templates/install_puppet.bat.erb
```diff
@@ -23,9 +23,12 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
-start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
+start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>" <% unless @install_dir.empty? -%>INSTALDIR="<%= @install_dir %>"<% end -%>
```
Iristyle: This property is `INSTALLDIR`, not `INSTALDIR`
https://docs.puppet.com/puppet/latest/reference/install_windows.html#installdir

Renelast:
@Iristyle I've removed the sc start command from this PR. Also corrected the INSTALLDIR typo.

The startup type was indeed set to manual on the previous installation so that would explain our "issue" 
Doe you happen to know where I can find the registry setting to change this? Or perhaps should we add an auto startup parameter to this module?

MikaelSmith:
That makes sense to me: if the module specifies the service should be running, we should add the flag to the MSI to ensure it starts.

Renelast:
@MikaelSmith Do you want me to add that to this PR or shall I open another one that adds that functionality?

MikaelSmith:
A separate PR is fine.

How are you using this change? It's not exposed in the main class.

Renelast:
@MikaelSmith We are setting this in hiera

e.g.
<code>
puppet_agent::windows::install::install_dir: 'C:\Management\Programs\Puppet'
</code>

I could update the PR so it's exposed through the main class if that is preferred?

MikaelSmith:
I'd like to see it exposed in documentation and the main class. Definitely note that it's Windows-only.

james-powis:
FYI the travis checks are failing due to updated puppet-lint gem... I added the exclusion of the newly defined checks in PR #127

MikaelSmith:
https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/127 also adds a simple test around `INSTALLDIR`. It'd be nice to see one here that tests with and without `install_dir` set.

MikaelSmith:
You should be able to rebase against master to fix the CI issue.

MikaelSmith:
Closing in favor of https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/130.

Issue #126, opened 2016-06-22
install_puppet.bat.erb does not respect nodes running in non-"production" environments


james-powis:
upgrading puppet causes puppet.conf file environment setting to revert to default explicitly. For nodes living in other environments this can cause issues. This does not impact situations where an enc is overriding environment. 

I will submit a PR with the appropriate fix.

MikaelSmith:
@Iristyle is this the expected behavior for the MSI?

MikaelSmith:
Thanks! Filed [MODULES-3517](https://tickets.puppetlabs.com/browse/MODULES-3517) to help us track it.

james-powis:
For completeness before upgrade puppet.conf had:
[agent]
...
environment = customers
server = puppet.master.mydomain.com

Post upgrade:
[agent]
environment =production
server =puppet.master.mydomain.com

Notice the changing of spacing, all other lines were left alone spacing wise and no settings were missing.

brettswift:
If you remove the environment completely, because you are relying on it getting it's environment from the console node group / environment classifier..   then this will assign `environment=master`. 

Is someone assigned to this? If not I might be able to look at this, as we have ~100 windoze to migrate fairly soon.

MikaelSmith:
The Node Classifier will be authoritative, unless you specifically set it up to defer to the agent. However, I agree that the default should probably not be to touch the environment setting.

james-powis:
I think that regardless of the behavior of the MSI in the future, ultimately for backwards compatibility the MSI execution this module is performing will have to pass PUPPET_AGENT_ENVIRONMENT="$::environment" in order to provide full matrix support.

glennsarti:
@james-powis I think I'm beginning to lean that way too.

MikaelSmith:
Merged https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/127.

PR #127, opened 2016-06-22
Explicitly setting environment to nodes current environment

james-powis:
Per #126 the windows agent upgrade batch script does not preserve the agent configured environment. This change will persist the environment config post upgrade.

Additionally the updated gems for Lint cause 140 character check issues, and complains about puppet:// source paths where modules/ is not in the string. I have suppressed these lint checks as they cause issues with the existing untouched code base.

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -6,6 +6,9 @@ set pid=
 for /f "tokens=2" %%a in ('tasklist /v ^| findstr /c:"%windowTitle%"') do set pid=%%a
 set pid_path=%TEMP%\puppet_agent_upgrade.pid
 
+set environment=
+for /f "delims=" %%i in ('puppet config print --section agent environment') do set environment=%%i
+
 if exist %pid_path% del %pid_path%
 @echo %pid%> %pid_path%
 
@@ -23,7 +26,7 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
-start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
+start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>" PUPPET_AGENT_ENVIRONMENT="%environment%"
 
 if exist %pid_path% del %pid_path%
 


spec/classes/puppet_agent_windows_install_spec.rb
```diff
@@ -52,6 +52,7 @@
             it { is_expected.to contain_file('C:\tmp\install_puppet.bat').with_content(
                 %r[#{Regexp.escape("msiexec.exe /qn /norestart /i \"#{values[:appdata]}\\Puppetlabs\\packages\\puppet-agent-#{values[:expect_arch]}.msi\"")}])
             }
+            it { is_expected.to contain_file('C:\tmp\install_puppet.bat').with_content(/PUPPET_AGENT_ENVIRONMENT="rp_env"/) }
```
MikaelSmith: Where does "rp_env" come from?

james-powis: It is the value of the $::environment fact while running in rspec-puppet... interestingly I have yet to figure out how to override that specific fact/top scope variable in rspec... but for this use case it works to test absence of typo/quoting issues...

MikaelSmith:
Thanks! This looks pretty good, but I pinged some people for feedback on the issue I'd like to get before merging this.

templates/install_puppet.bat.erb
```diff
@@ -23,7 +23,7 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
-start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>"
+start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>" PUPPET_AGENT_ENVIRONMENT="<%= @environment %>"
```
glennsarti: Where is `@environment` being set?  Can it be blank for nil?

MikaelSmith: It's a built-in variable: https://docs.puppet.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#serverfacts-variable

glennsarti: Hrmm, given the comment about it being overridable, should we be using `@server_facts['environment']` (or similar) instead?

MikaelSmith: That doesn't appear to be available in Puppet 3.8.

glennsarti: I image ideally you just want to read what's in the current puppet.conf and convert that to the appropriate MSI public properties

MikaelSmith: We could extend this module to explicitly manage the `environment` setting, defaulting to undef (which would imply no change). That would at least avoid accidental changes, and allow working around the MSI's behavior.

james-powis: `server_facts['environment']` would return the environment setting in the puppetserver(puppet-master)'s puppet.conf file and would be bad all around... 

For the purposes of discussion:
Assume the following, the agent configured environment is "testing", and the ENC override-ed environment is "lab",

The puppet catalog will compile under lab but there will be a warning about the requested environment being set to testing. This at least in my opinion is a bad behavior that should be corrected...

Interestingly, when the agent environment setting is `production` (default), and the enc override setting is `testing`, the agent has no issues with functioning (regardless of a local environment directory existing to match the ENC override environment). However if the agent environment setting is explicitly updated to match the enc enforced environment `testing` and the `{puppet_code_dir}/environments/testing` directory does not exist, then the agent will die and fail to run. 

So at a bare minimum this PR needs to add code to ensure that topscopped `$::environment` directory exists...

There has been an long lived bug (or in this case its kind of a feature) in 3.x which made it into 4.x which set local scoped `$environment` to Agent requested environment, and `$::environment` to the ENC override-ed environment setting. I have observed this with the hiera.yaml directory environments path directives. It _MAY_ be possible that this bug/feature also exists in the catalog compilation... However in 4.x we gained access to `$::agent_specified_environment` fact which would lets us effectively set the MSI Installer ARG to be that of the previous puppet.conf setting... Unfortunately that fact does not exist in 3.8... Unfortunately this is not something which we can easily test with rspec-puppet.

I will investigate this "feature" in 3.8 master agent config to see if it functions in a catalog the same way it does in the hiera.conf

Marking this as a WIP until I get those answers.

MikaelSmith: The warning issue was fixed in Puppet 4.2.2 ([PUP-3318](https://tickets.puppetlabs.com/browse/PUP-3318)).

In more recent versions, if Puppet is successfully running and `$::environment` hasn't been overridden by a custom fact, then the directory should exist. That probably doesn't hold true for Puppet 3.7 and older (not quite sure about 3.8).

james-powis: @MikaelSmith FYI I retracted my previous rant and came up with a much better solution...

MikaelSmith: I'm not sure that's necessary over `$environment` if you account for PUP-3318. If you come up with a scenario where that won't work, then the new fact should probably follow Puppet's implementation at https://github.com/puppetlabs/puppet/blob/master/lib/puppet.rb#L149-L155.

james-powis:
@MikaelSmith so I can think of a few use cases where `$environment` being used would be undesirable.

Namely when using puppet agent or apply --environment is used or when ENC is enforcing environment and the puppet.conf environment should not be aligned with the ENC's override... Unfortunately duplicating the code using agent_specified_environment will not work in 3.8.7... This is because set_by_config sub is not present in 3.8.7. Additionally this would not work when using --environment switch. 

This leaves me with really only one realistic option... load the puppet.conf and read the ini setting out of it defaulting to production... I really wish it was not this messy but it appears to has to be...

templates/install_puppet.bat.erb
glennsarti: That is not going to be trivial in a BAT file.

glennsarti:
That is not going to be trivial in a BAT file.

MikaelSmith:
I'd expect a custom fact. It also doesn't cover people running `puppet agent/apply --environment ...`.

james-powis:
well Facter::Core::Execution.exec('puppet config print --section agent environment') in a custom fact works beautifully... but I dont want to shell out and spin a whole new puppet process to get one little detail so i am trying to figure out how to invoke that function via ruby code... Lost in the land of face right now...

james-powis:
OK so after spending a couple of hours trying to figure out how to interface with the ruby libs the same way that puppet agent config does I have given up... this custom fact works but it is heavy handed... Perhaps someone over there at puppet can advice on a better methodology. Other wise I am calling this done. 

Testing the fact with apply works very well:
```puppet
20:07:18:~/git/personal/puppetlabs-puppet_agent
james@MacBook-Pro (master +%=)> FACTERLIB=/Users/james/git/personal/puppetlabs-puppet_agent/lib/facter puppet apply --environment foo -e 'notify { $::puppet_agent_environment: }'
Notice: Compiled catalog for macbook-pro.local in environment foo in 0.03 seconds
Notice: production
Notice: /Stage[main]/Main/Notify[production]/message: defined 'message' as 'production'
Notice: Applied catalog in 0.26 seconds```

File: lib/facter/settings.rb [hunk 3]
```diff
@@ -51,3 +51,16 @@
...
+Facter.add('puppet_agent_environment') do
+  setcode do
+    # There has to be a more graceful way to accomplish this, but I cannot seem to find a way
+    # via the existing ruby library in puppet... I think face action :config is worth looking at
+    # but have not been able to make it work, so this will have to do.
+    #
+    # This fact gets the puppet.conf config environment setting independent of the enc override or 
+    # the cli --environment override. The goal is to specify the agent environment to put into the 
+    # puppet.conf file by the msi installer without accidentally overriding it.
+    Facter::Core::Execution.exec('puppet config print --section agent environment');
```
MikaelSmith: I think the following should work (from https://github.com/puppetlabs/puppet/blob/master/lib/puppet/face/config.rb#L57-L84)
```
values_from_the_selected_section = Puppet.settings.values(nil, :agent)

loader_settings = {
  :environmentpath => values_from_the_selected_section.interpolate(:environmentpath),
  :basemodulepath => values_from_the_selected_section.interpolate(:basemodulepath),
}

Puppet.override(Puppet.base_context(loader_settings),
    "New environment loaders generated from the requested section.") do
  # And now we can lookup values that include those from environments configured from
  # the requested section
  values = Puppet.settings.values(Puppet[:environment].to_sym, :agent)
  value = values.interpolate(:environment)
end
```

MikaelSmith:
Another update was merged to fix the CI issues that conflicts with yours.

lib/facter/settings.rb
james-powis: @MikaelSmith I am a bit confused by config_by_cli I am not able to find where this is coming from can you provide a blob?

I am also a bit perplexed by only invoking on older versions... Assuming you are talking about `puppet config` and not the blob provided (due to puppet.settings not being available in 3.8 per your previous comment). 

In the interests of code simplicity, perhaps a custom fact is not well placed in this case we need the value exclusively at upgrade and only for windows... Might I recommend executing the `puppet config print --section agent environment` chunk from within the .bat

james-powis:
OK i have resolved the merge conflict. Additionally instead of using a fact which only provides the value when the bat is performing an upgrade and only on windows, I have instead moved the `puppet config print --section agent environment` into install_puppet.bat.erb in this way we do not need to exception code based on version and efficiency. I feel this provides the best of both worlds, the code remains simple, and the result is accomplished.

templates/install_puppet.bat.erb
```diff
@@ -6,6 +6,9 @@ set pid=
 for /f "tokens=2" %%a in ('tasklist /v ^| findstr /c:"%windowTitle%"') do set pid=%%a
 set pid_path=%TEMP%\puppet_agent_upgrade.pid
 
+set environment=
+for /f "delims=" %%i in ('puppet config print --section agent environment') do set environment=%%i
```
MikaelSmith: This is probably the most reasonable approach. Eventually we may be able to guard it depending on which version of the MSI is installing, but currently the MSI always updates the `environment` setting, so we should ensure it uses whatever is configured for the agent section.

PR #130, opened 2016-07-06
(MODULES-3571) Allow setting install_path for MSI

MikaelSmith:
Allow setting the INSTALLDIR option for the Windows MSI.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -131,6 +131,12 @@ An array of services to start, normally `puppet` and `mcollective`. If the array
 
 Alternate source from which you wish to download the latest version of Puppet.
 
+####`install_dir`
+
+The directory the puppet agent should be installed to. This is only applicable for windows operating systems.
+This only applies when upgrading the agent to a new version; it will not cause re-installation of the same version to a new location.
+
+
 ## Limitations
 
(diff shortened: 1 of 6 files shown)


README.markdown
```diff
@@ -131,6 +131,12 @@ An array of services to start, normally `puppet` and `mcollective`. If the array
 
 Alternate source from which you wish to download the latest version of Puppet.
 
+####`install_dir`
+
+The directory the puppet agent should be installed to. This is only applicable for windows operating systems.
+This only applies to an agent upgrade and will not change an already installed agent.
```
Renelast: @mikaelsmith I meant to say that setting the install_dir option will not reinstall an agent package to a new directory. This setting only has effect if the agent on the client is being updated. Does that make any sense?

MikaelSmith: Yes, thanks. I'm going to tweak the wording slightly to `This only applies when upgrading the agent to a new version; it will not cause re-installation of the same version to a new location.`

MikaelSmith:
@Iristyle @highb mind giving this a review?

highb:
👍 from me

PR #131, opened 2016-07-12
Test if stringify_facts = true on agent

hpcprofessional:
and give a helpful error message if it is.

stringify_facts = false is a stated requirement in the documentation.

However,  if that is missed, or if the occasional agent has it unset (and therefore true) then the error message is indecipherable. To aid the unlucky troubleshooter, this PR will identify the consequence of stringify_facts being true and a particular data structure being a String instead of a Hash.


MikaelSmith:
This seems useful, thanks. I'm going to squash commits and rewrite the commit message a bit.

PR #134, opened 2016-07-25
add support for non-US-English systems

aharden:
We found non-US-English systems (specifically, Japanese) kept spawning additional copies of this script.  Adding a code-page-change to 437 (US English) corrected that behavior and allowed agent upgrade to proceed.

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -2,6 +2,9 @@ set AGENT_PID=%1
 set windowTitle=Puppet Agent Upgrade
 title %windowTitle%
 
+REM set code page to US (English) to adjust non-US English systems if required
+chcp 437
+
 set pid=
 for /f "tokens=2" %%a in ('tasklist /v ^| findstr /c:"%windowTitle%"') do set pid=%%a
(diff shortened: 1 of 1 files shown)


Iristyle:
@aharden thanks for the report and the PR - I filed a ticket for this issue at https://tickets.puppetlabs.com/browse/MODULES-3636 and have a PR up at #135 to address the problem.

Closing this in favor of that PR.

templates/install_puppet.bat.erb
```diff
@@ -2,6 +2,9 @@ set AGENT_PID=%1
 set windowTitle=Puppet Agent Upgrade
 title %windowTitle%
 
+REM set code page to US (English) to adjust non-US English systems if required
+chcp 437
```
Iristyle: For future reference, this can cause some unintended consequences as Ruby will set `Encoding.default_external` to this value.  It's also a global setting in Windows, which will impact other applications.

We sometimes use this in tests, but typically revert it back immediately.

In any event, we've found a simpler method of checking for running processes in the other PR.

Thanks again!

PR #136, opened 2016-07-27
(MODULES-3434) Wait 120 seconds for Windows agent

Iristyle:
- Previously, the batch file would wait for an infinite amount of
   time for the calling Puppet process to exit, before it would
   proceed with an upgrade.

   Prior to rewriting the PID check with wmic, this could cause an
   infinite loop when trying to match an unlocalized string.

   Since infinite loops are never a good idea, cap the wait limit to
   120 seconds, and log an event log message under Puppet / ID 101
   that the agent upgrade failed.

   This should improve the end user experience under failing
   circumstances.

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -14,9 +14,8 @@ if exist %pid_path% del %pid_path%
 
 :wait_for_pid
 timeout /t 5 /nobreak > NUL
-FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=%%A
-echo %_task% | findstr "No tasks are running" >nul
-IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
+wmic path Win32_Process where handle=%AGENT_PID% get handle /format:textvaluelist | findstr /I "Handle=%AGENT_PID%"
+IF %errorlevel% == 0 ( GOTO wait_for_pid )
 
 REM This *must* occur after Puppet Agent has finished applying its
 REM prior catalog which manages the pxp-agent service state. If not,
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -12,9 +12,21 @@ for /f "delims=" %%i in ('puppet config print --section agent environment') do s
 if exist %pid_path% del %pid_path%
 @echo %pid%> %pid_path%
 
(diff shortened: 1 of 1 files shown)


Iristyle:
Builds on #135

templates/install_puppet.bat.erb
```diff
@@ -12,11 +12,22 @@ for /f "delims=" %%i in ('puppet config print --section agent environment') do s
...
-echo %_task% | findstr "No tasks are running" >nul
-IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
+wmic path Win32_Process where handle=%AGENT_PID% get handle /format:textvaluelist | findstr /I "Handle=%AGENT_PID%"
+
+SET /A pid_checks_performed+=1
+REM Wait for a max of 120 seconds (or 24 iterations) before aborting the upgrade
+IF %errorlevel% == 0 IF %pid_checks_performed% == 24 (
+  REM Adding CustomSource allows EventCreate to work properly
+  reg add HKLM\SYSTEM\CurrentControlSet\services\eventlog\Application\Puppet /v CustomSource /t REG_DWORD /d 1 /f
+  EVENTCREATE /T ERROR /L APPLICATION /SO Puppet /ID 101 /D "Puppet agent upgrade failed while waiting for Puppet process [ID: %AGENT_PID%] to exit."
```
MosesMendoza: Just out of curiosity, is there a reason you picked `101`?

Iristyle: That's a good question - thanks for pointing it out.

We only have 3 different error messages (which could theoretically be localized) defined in the Puppet event log.  

They were introduced back in this commit - https://github.com/puppetlabs/puppet/commit/d88f3e7c387eff270101604f2fec2e086f88ebdd

I intentionally selected a value that didn't conflict with any of these (really, the error codes should be unique to a particular failure type), with the hopes that we would later define 101 to be an upgrade failure.

Iristyle: Unfortunately, in the event log, this can lead to the following message (until we define 101):

```
The description for Event ID 101 from source Puppet cannot be found. Either the component that raises this event is not installed on your local computer or the installation is corrupted. You can install or repair the component on the local computer.

If the event originated on another computer, the display information had to be saved with the event.

The following information was included with the event: 

Puppet agent upgrade failed while waiting for process [4] to exit.

the message resource is present but the message is not found in the string/message table
```

PR #137, opened 2016-07-29
(MODULES-3657) Fix waiting on Windows during upgrade

glennsarti:
Previously the script was using the timeout.exe application however this
attempts to redirect STDIN and during a headless operation, for example during
an actual puppet run, this throws an error and does not wait the proper time.
This commit changes the wait method to use ping.  Ping waits 1 second between
attempts therefore a simple `ping 127.0.0.1 -n 6` will pause for 5 seconds.

This commit also changes the location of the PID file slightly.  Previously it
used `%TEMP%` however this variable can be removed or be in an unexpected place.
Instead the PID file is always placed in the same directory as the
script (`%~dp0`).

To assist in debugging upgrade failures the install batch file also outputs
all environment variables prior to running.

This PR builds on #136 and #135

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -14,9 +14,8 @@ if exist %pid_path% del %pid_path%
 
 :wait_for_pid
 timeout /t 5 /nobreak > NUL
-FOR /F "tokens=*" %%A IN ('tasklist /FI "PID eq %AGENT_PID%" /NH') DO set _task=%%A
-echo %_task% | findstr "No tasks are running" >nul
-IF NOT %errorlevel% == 0 ( GOTO wait_for_pid )
+wmic path Win32_Process where handle=%AGENT_PID% get handle /format:textvaluelist | findstr /I "Handle=%AGENT_PID%"
+IF %errorlevel% == 0 ( GOTO wait_for_pid )
 
 REM This *must* occur after Puppet Agent has finished applying its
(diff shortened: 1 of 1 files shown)


templates/install_puppet.bat.erb
```diff
@@ -1,22 +1,36 @@
+SET
+
 set AGENT_PID=%1
 set windowTitle=Puppet Agent Upgrade
 title %windowTitle%
 
 set pid=
 for /f "tokens=2" %%a in ('tasklist /v ^| findstr /c:"%windowTitle%"') do set pid=%%a
-set pid_path=%TEMP%\puppet_agent_upgrade.pid
+set pid_path=%~dp0puppet_agent_upgrade.pid
```
Iristyle: Hmm... and we're certain this is always writable?  And that there are no side effects of writing into this Puppet module template dir (consider PMT install vs pluginsync)?

I think it would be preferable to just set `%TEMP%` to `%windir%\Temp` if it's not already set:

```batch
IF "%TEMP%"=="" (SET TEMP=%windir%\Temp)
```

Iristyle: Comment withdrawn.  I was mistakenly thinking that `~dp0` is the template directory.  Its the location of the `install_puppet.bat` created from the ERB template, which is known to be the temp directory.

👍 to the change.

Iristyle:
Builds on #135 and #136

PR #140, opened 2016-08-09
(MODULES-3453) Disable services before upgrade

MikaelSmith:
Ensure old services are stopped and disabled before upgrade. This is
necessary on some platforms where uninstalling or upgrading a package
doesn't halt the service, especially when moving from pe-puppet to
puppet-agent packages.


MikaelSmith:
@highb

highb:
@MikaelSmith Do you have the results of an upgrade on Solaris with this to post?

MikaelSmith:
No, I'll setup some test runs.

Iristyle:
As general commentary, this solution looks like a lot of work / code.  Is there a reason to not just add a few lines to the batch file to forcefully terminate `puppet` / `pe-puppet` / `mcollective`  `pe-mcollective' in the batch file?

I filed https://tickets.puppetlabs.com/browse/MODULES-3757 with some more details about that

What happens when it's the `puppet` or `pe-puppet` service that hangs?  You can't manage that from within Puppet during the run initiated by the service... and would have to kill the PID in the batch file.

The service hang issues were originally fixed in https://tickets.puppetlabs.com/browse/PUP-2987

Iristyle:
Ahh... I now see after reading the MODULES-3453 ticket, that the problem this is intended to address affects all platforms.

So my prior comment was a bit misinformed there.  But I would still make the point that the Windows solution is a bit simpler - maybe 10-15 more lines of code in the batch file.

highb:
Can we get a rebase on this?

MikaelSmith:
I'm starting to think the Windows approach of running a batch file afterwards is the best option for all platforms. However, rebased.

highb:
@MikaelSmith Did the test runs on this PR go OK?

highb:
Sorry about that accidental close/comment. That button is too close!

MikaelSmith:
I don't recall, it probably hasn't been tested sufficiently.

highb:
@MikaelSmith Does this still need a merge or are we looking at alternative solutions?

MikaelSmith:
More that it probably needs to be run through several acceptance suites, and I haven't had time to do that.

glennsarti:
@MikaelSmith can we close this?

MikaelSmith:
I'm not working on it, so sure.

Issue #146, opened 2016-08-24
Permissions on puppet-agent-x64.msi attempt to update every Puppet run


esalberg:
After using the module to upgrade from 3.8.2 to PE 2016.2.1 am seeing the Windows agents get updated every Puppet run in the PE 2016.2.1 environment (as far as I can tell, the upgrade was successful):

mode changed '0664' to '0644'
Source:	/Stage[main]/Puppet_agent::Prepare::Package/File[C:\ProgramData\Puppetlabs\packages\puppet-agent-x64.msi]/mode
File:	/etc/puppetlabs/code/environments/myenv/modules/puppet_agent/manifests/prepare/package.pp Line: 40

Looking at the file itself, it has:
Everyone: Read
SYSTEM: Full control
Administrators: Special
 -> Administrators have all permissions except for "Full control" and "Traverse folder / execute file"

I tried removing the 6 checkboxes for write permissions by hand on my test node.  Puppet then responded that the permissions needed to change from 0444 to 0644.  Apparently when it did that change, it did not do so successfully, for it then started back with the need to change the file from 0664 to 0644.

Per this ticket: https://tickets.puppetlabs.com/browse/PUP-5491
The reason for this behavior is because when puppet service runs as SYSTEM, the default DACL (analogous to umask) will cause any newly created file/dir to have owner Administrators (note plural) and group SYSTEM. [...] By comparison, when running in an elevated context (Administrator), then the owner will be you, and group will be None.

However, from the behavior I'm seeing, I almost wonder if the owner of the file is Administrators -and- the group is Administrators as well, in which case those values need to be set to the same thing (e.g. 0664 or 0444, but not 0644)?  Or should Puppet be updating the Windows file permissions differently?

From that same ticket I linked:
"I found a reference in Mechanics of User Identification and Authentication that explains this. SYSTEM is an implicit / hidden member of the Administrators group [...] an explicit grant to Administrators should also cover SYSTEM in theory as well."

FYI, I do get the same permissions change every run when running "puppet agent -t" versus running puppet as the service.

MikaelSmith:
Fix merged, and filed https://tickets.puppetlabs.com/browse/MODULES-3834 to track further.

PR #147, opened 2016-08-24
Set perms for Windows package differently than for Linux

esalberg:
See issue:
https://github.com/puppetlabs/puppetlabs-puppet_agent/issues/146

All I did was set the Windows permissions on that file to be 0664 and leave everything else 0644.  I don't know if this is the correct way to patch what I was seeing or not, but I figured I'd put in the PR anyway.  I'll have more testing done in the next couple of days as I move all my agents to PE 2016.

### DIFF
--- a/manifests/prepare/package.pp
+++ b/manifests/prepare/package.pp
@@ -33,15 +33,17 @@ class puppet_agent::prepare::package(
 
     if $::osfamily =~ /windows/ {
       $local_package_file_path = windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}")
+      $mode = undef
     } else {
       $local_package_file_path = "${::puppet_agent::params::local_packages_dir}/${package_file_name}"
+      $mode = '0644'
     }
 
     file { $local_package_file_path:
       ensure  => present,
       owner   => $::puppet_agent::params::user,
       group   => $::puppet_agent::params::group,
-      mode    => '0644',
+      mode    => $mode,
       source  => $source,
       require => File[$::puppet_agent::params::local_packages_dir],
     }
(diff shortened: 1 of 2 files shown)


highb:
@esalberg Thanks for opening a PR! Apparently our linter settings are doing something funky, and that is why travis is failing on this PR. We're going to fix those up soon. I'll re-run the specs then and then we can run this through review.

esalberg:
I went ahead and implemented the lint update in my PR too (hopefully not an issue).

MikaelSmith:
@Iristyle any context on this?

Iristyle:
@MikaelSmith I haven't seen anything like this before (and it certainly doesn't happen in our tests).

However, if the file has picked up some borked permissions via inheritance from its parent, I could see where that might be a problem.

manifests/prepare/package.pp
```diff
@@ -33,15 +33,17 @@
 
     if $::osfamily =~ /windows/ {
       $local_package_file_path = windows_native_path("${::puppet_agent::params::local_packages_dir}/${package_file_name}")
+      $mode = '0664'
```
Iristyle: Is there a reason we even need to set a mode on Windows?

Would `undef` be better? (I have not verified that as viable)

MikaelSmith: I'm not aware of a reason why we set mode. @jpartlow wrote this originally.

jpartlow: So, the original manifest was only dealing with the SLES/OSX cases where we needed to grab the package manually due to a lack of package management support; I think the fact that there was a POSIX mode was simply overlooked when it was reworked to include windows.  @Iristyle, @MikaelSmith I assume undef would be fine for the Windows case so long as the catalog is subsequently able to install the msi it has downloaded.

MikaelSmith: I concur with `undef` being sufficient on Windows.

esalberg:
Per the discussion, I updated the windows mode to be undef instead of 0664.  Please let me know if I did this hastily, and I can put a mode back.

MikaelSmith:
Thanks for the fix!

PR #149, opened 2016-08-25
(PE-17157) Do not attempt to unzip tarball when upgrade is unneeded

demophoon:
Before this commit if you were on solaris with puppet aio and ran puppet
agent -t after classifying the node with the puppet agent module the
manifest would fail to apply because there was no logic which gated us
from unzipping the agent tarball if an upgrade was not needed.

This commit verifies that the version of puppet agent on the master is
the same on the agent before attempting an upgrade.

### DIFF
--- a/manifests/prepare.pp
+++ b/manifests/prepare.pp
@@ -94,7 +94,7 @@ class puppet_agent::prepare(
       contain $_osfamily_class
     }
     'solaris': {
-      if $::aio_agent_version != $package_version {
+      if $::puppet_agent::aio_upgrade_required {
         class { $_osfamily_class:
           package_file_name => $package_file_name,
         }
--- a/manifests/prepare.pp
+++ b/manifests/prepare.pp
@@ -87,11 +87,19 @@ class puppet_agent::prepare(
   # the osfamily of the client being configured.
 
   case $::osfamily {
-    'redhat', 'debian', 'windows', 'solaris', 'aix', 'suse', 'darwin': {
+    'redhat', 'debian', 'windows', 'aix', 'suse', 'darwin': {
       class { $_osfamily_class:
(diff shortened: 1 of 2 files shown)


demophoon:
I'm running this through some manual tests now. Will post results shortly.

MikaelSmith:
Rebase to pick up a Travis CI fix.

manifests/prepare.pp
```diff
@@ -87,12 +87,20 @@
...
   case $::osfamily {
-    'redhat', 'debian', 'windows', 'solaris', 'aix', 'suse', 'darwin': {
+    'redhat', 'debian', 'windows', 'aix', 'suse', 'darwin': {
       class { $_osfamily_class:
         package_file_name => $package_file_name,
       }
       contain $_osfamily_class
     }
+    'solaris': {
+      if $::aio_agent_version != $package_version {
```
MikaelSmith: I'm curious why we shouldn't just do this for all platforms.

demophoon: We could, it was mainly to avoid rerunning the exec statements which are not idempotent with solaris so I figured I'd limit the impact of the change to just solaris.

We were specifically having issues with this exec statement (https://github.com/demophoon/puppetlabs-puppet_agent/blob/0eb56e366c84ebc9d0a720e8a1227cf9f259d7ca/manifests/osfamily/solaris.pp#L22-L28) always trying to untar the agent tarball downloaded from the master. And in the case that the solaris agent is already AIO then we never download a tarball and the exec command fails to run since there isn't anything to untar.

MikaelSmith: Ok; then I'd like the logic in prepare/package.pp to match.

gimmyxd:
Closing this at as ticket was resolved.

PR #150, opened 2016-09-02
AIX fixes for PowerPC_POWER8 architecture and operatingsystem case fix

guesswho911:


guesswho911:
The spec testing for AIX is incorrect as well since the test looks at the architecture of the PowerPC chipset to determine the OS version, which is not how that works.  Spec test will for AIX will need to be rewritten to appropriately test AIX versions.

MikaelSmith:
I seem to remember some more work needed for this. The architecture is tied to package version (which may not be correct, and shouldn't find a valid package). Are you overriding the source parameter to make this work?

guesswho911:
The changes I made to the detected PowerPC architecture allow use of the module.  I don't know if AIX 7.2 has seen agent release for it as we use the 7.1 agent currently.

I'm trying to see if I can get the module and spec tests to agree.

While the OS versions take advantage of the latest in the chipset (AIX 7.2 for PowerPC8, AIX 7.1 for PowerPC7, AIX 6.1 for PowerPC6), it is not tied to the PowerPC architecture.  PowerPC8 can be used all the way back to AIX 5.3 and AIX 7.1 can use any chip back to PowerPC4 (7.2 requires PowerPC7 or higher).

highb:
@guesswho911 Thank you for your contribution, but this PR has been subsumed by #162 .

PR #152, opened 2016-09-09
(MODULES-3872) Updated to provide the ability to disable configuration of PE/FOSS repositories

exodusftw:
This is intended to enable users to utilize existing systems management
tools and internally hosted repositories to ensure agent upgrades

### DIFF
--- a/manifests/osfamily/redhat.pp
+++ b/manifests/osfamily/redhat.pp
@@ -88,14 +88,16 @@ class puppet_agent::osfamily::redhat(
     logoutput => 'on_failure',
   }
 
-  yumrepo { 'pc_repo':
-    baseurl       => $source,
-    descr         => "Puppet Labs ${::puppet_agent::collection} Repository",
-    enabled       => true,
-    gpgcheck      => '1',
-    gpgkey        => "${gpg_keys}",
-    sslcacert     => $_sslcacert_path,
-    sslclientcert => $_sslclientcert_path,
-    sslclientkey  => $_sslclientkey_path,
+  if $::puppet_agent::manage_repo {
+    yumrepo { 'pc_repo':
+      baseurl       => $source,
+      descr         => "Puppet Labs ${::puppet_agent::collection} Repository",
+      enabled       => true,
+      gpgcheck      => '1',
+      gpgkey        => "${gpg_keys}",
+      sslcacert     => $_sslcacert_path,
+      sslclientcert => $_sslclientcert_path,
+      sslclientkey  => $_sslclientkey_path,
+    }
   }
 }
--- a/manifests/osfamily/solaris.pp
+++ b/manifests/osfamily/solaris.pp
@@ -37,61 +37,63 @@ class puppet_agent::osfamily::solaris(
       }
     }
     '11': {
-      class { 'puppet_agent::prepare::package':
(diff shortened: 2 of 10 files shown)


MikaelSmith:
Seems very useful. 👍

manifests/osfamily/redhat.pp
```diff
@@ -88,14 +88,16 @@
...
-  yumrepo { 'pc_repo':
-    baseurl       => $source,
-    descr         => "Puppet Labs ${::puppet_agent::collection} Repository",
-    enabled       => true,
-    gpgcheck      => '1',
-    gpgkey        => "${gpg_keys}",
-    sslcacert     => $_sslcacert_path,
-    sslclientcert => $_sslclientcert_path,
-    sslclientkey  => $_sslclientkey_path,
+  if $::puppet_agent::manage_repo {
```
MikaelSmith: If you're not managing the repo, do you really want all the GPG key setup?

manifests/osfamily/solaris.pp
```diff
@@ -37,61 +37,63 @@
       }
     }
     '11': {
-      class { 'puppet_agent::prepare::package':
```
MikaelSmith: No support for Solaris 10 in the managed repositories?

MikaelSmith:
@highb have any other thoughts on this?

manifests/osfamily/redhat.pp
exodusftw: For my particular use case, I thought it was beneficial to still have puppet_agent install the gpg keys so there's no need to distribute them through another mechanism and the rpms could be loaded in internal repos without being re-signed.

That said, I could see the argument being made that if the ability to disable repo creation is there, then the same ability should exist for the gpg key.

MikaelSmith: Works for me, putting the gpg keys there shouldn't prevent any other configuration.

manifests/osfamily/solaris.pp
exodusftw: That's a good question, I didn't add the bool there as there was no repo being created for Solaris 10...I guess it's possible that there could be externally managed repos for Solaris 10, so that block should have enable/disable functionality. Thoughts?

exodusftw:
Did you have any issues with the spec tests?

There's probably a more elegant way to handle testing with and without the repos being managed, but I thought the tests were more straightforward if just split into separate with/without contexts

MikaelSmith:
Yeah, spec tests seemed fine.

manifests/osfamily/solaris.pp
highb: Makes sense to me.

highb:
@MikaelSmith @exodusftw This PR is excellent and I'm 👍 from a code perspective after my cursory glance. Have we had a chance to run it through the acceptance tests yet?

MikaelSmith:
I have not.

MikaelSmith:
Filed as https://tickets.puppetlabs.com/browse/MODULES-3872.

MikaelSmith:
@highb passed an upgrade from PE 2016.1.2 to 2016.2.1 on centos6-64mcda-centos7-64a.

exodusftw:
Thanks @MikaelSmith @highb

PR #153, opened 2016-09-15
(MODULES-4236) Disable proxy for yum repo as it will not pass through to Puppet Server and will not hand over the certs.

cyberious:
- We have system wide proxy for all repos and this is critical for us so we can use the module

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -46,6 +46,7 @@ class puppet_agent (
   $service_names   = $::puppet_agent::params::service_names,
   $source          = $::puppet_agent::params::_source,
   $install_dir     = $::puppet_agent::params::install_dir,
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Sometimes sourcepath points to a public url. We should probably only disable proxy when source points to a puppet master.

cyberious:
So what do you think is a viable check for internet? not using port `8140`? We override the agent source b/c we use an haproxy and in case one gets overloaded.

MikaelSmith:
That's tricky isn't it. Not using port `8140` might be the best we can do. Alternatively make disabling the proxy a param that can be overridden.

cyberious:
Updated to have as a parameter defaulting to false to maintain the current behavior. Only currently affects redhat

PR #154, opened 2016-09-16
(PE-17508) Fix puppet-agent suffix on fedora

johnduarte:
This commit modifies the install logic when `osfamily` is `RedHat`
to accommodate for the `fedoraf` package suffix when the OS is
fedora rather than an `el` derivative.

Prior to this commit the `puppet-agent` package would be suffixed
with `el` which does not match the package naming convention for
fedora.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -107,8 +107,13 @@ class puppet_agent::install(
     }
   } elsif ($::osfamily == 'RedHat') and ($package_version != 'present') {
     # Workaround PUP-5802/PUP-5025
+    if ($::operatingsystem == 'Fedora') {
+      $pkg_os_suffix = 'fedoraf'
+    } else {
+      $pkg_os_suffix = 'el'
+    }
     package { $::puppet_agent::package_name:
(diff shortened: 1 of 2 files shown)


manifests/install.pp
```diff
@@ -107,8 +107,13 @@
...
   } elsif ($::osfamily == 'RedHat') and ($package_version != 'present') {
     # Workaround PUP-5802/PUP-5025
+    if ($::operatingsystem == 'Fedora') {
+      $pkg_os_suffix = 'fedoraf'
+    } else {
+      $pkg_os_suffix = 'el'
+    }
     package { $::puppet_agent::package_name:
-      ensure => "${package_version}-1.el${::operatingsystemmajrelease}",
+      ensure => "${package_version}-1.${pkg_os_suffix}${::operatingsystemmajrelease}",
```
highb: This indicates that I have never tested this module on Fedora. Fail.

highb: I guess it's a new platform, so that makes some sort of sense.

johnduarte:
@highb specs updated and :green_apple: in Travis.

PR #158, opened 2016-09-23
updated unless condition on Suse/Redhat GPG key imports to ensure downcasing

exodusftw:
the current code on Suse simply converts all letters to the character 'a' and causes each run to re-import keys - so just for consistency sake, setting redhat/suse to both use the [:upper:] [:lower:] syntax

* edit - I should clarify - this doesn't happen on all systems,  but the behavior appears to occur on older versions of coreutils

### DIFF
--- a/manifests/osfamily/redhat.pp
+++ b/manifests/osfamily/redhat.pp
@@ -66,7 +66,7 @@ class puppet_agent::osfamily::redhat(
   exec {  "import-${legacy_keyname}":
     path      => '/bin:/usr/bin:/sbin:/usr/sbin',
     command   => "rpm --import ${legacy_gpg_path}",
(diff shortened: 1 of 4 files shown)


exodusftw:
Just to show an example:

```
root@examplehost:/tmp $ echo "ABCDEFG" | tr [A-Z] [a-z]
aaaaaaa

root@examplehost:/tmp $ echo "ABCDEFG" | tr A-Z a-z
abcdefg

root@examplehost:/tmp $ echo "ABCDEFG" | tr [:upper:] [:lower:]
abcdefg
```

MikaelSmith:
@highb any concerns?

MikaelSmith:
Created https://tickets.puppetlabs.com/browse/MODULES-3894 to track this.

highb:
👍  from me

PR #162, opened 2016-09-27
(FM-4989) Remove POWER version constraint for AIX

geoffnichols:
Prior to this commit, puppet-agent upgrades on AIX were constrained to
POWER 5-7 architectures.

However, if Puppet is already running on a supported version of AIX, it
should be safe to assume we're on a supported POWER version.

This commit removes the POWER version constraint, allowing upgrade on
POWER8 and higher architectures.


geoffnichols:
Ping @MikaelSmith and @highb for review.

geoffnichols:
@bh Yes, this successfully upgrades an AIX box from the PE 3.8.6 agent to puppet-agent-1.5.3.

PR #167, opened 2016-10-06
(maint) Use the forge for test fixtures

glennsarti:
Previously the test fixtures used pointed to the head of git repos however there
is no certainty that the code at HEAD is working.  This commit changes the
fixtures to use the forge which uses publicly released modules.


glennsarti:
@MikaelSmith @highb For review.

highb:
👍  Sane to me, and it looks like the tests still run OK in Travis.

PR #169, opened 2016-10-07
(maint) add task for generating nodesets for testing

MosesMendoza:
This commit adds a rake task (copied from puppetlabs-apt) that uses
beaker-hostgenerator to generate a nodeset compatible with our vmpooler. This
enables local testing with the vmpooler, in addition to the default testing
using vagrant. Note that we still need to tell beaker to look for our nodeset
using BEAKER_setfile=<nodeset>. Credit to @DavidS for the original contribution
to puppetlabs-apt.

Signed-off-by: Moses Mendoza <moses@puppet.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,4 +15,5 @@ end
 group :system_tests do
   gem 'beaker', '~> 2.16'
   gem 'beaker-rspec', '~> 5.1'
+  gem 'beaker-hostgenerator'
 end
--- a/Rakefile
+++ b/Rakefile
@@ -50,3 +50,36 @@ task :test => [
   :spec,
   :metadata,
 ]
+
+desc 'Generate pooler nodesets'
+task :gen_nodeset do
+  require 'beaker-hostgenerator'
+  require 'securerandom'
+  require 'fileutils'
+
+  agent_target = ENV['TEST_TARGET']
+  if ! agent_target
+    STDERR.puts 'TEST_TARGET environment variable is not set'
+    STDERR.puts 'setting to default value of "redhat7-64default.a"'
+    agent_target = 'redhat7-64default.a'
+  end
+
+  master_target = ENV['MASTER_TEST_TARGET']
+  if ! master_target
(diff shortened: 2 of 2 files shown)


glennsarti:
This may be superseded by my PR to bring this module under modsync control. https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/168

DavidS:
@glennsarti @MosesMendoza to note: that code was originally by justin stoller, is not actually used by anyone I know, and is slated for removeal once we can integrate BHG better into the module workflow.

MosesMendoza:
@glennsarti excellent! looking at that PR I think these may be complimentary? I'd want to remove my addition of beaker-hostgenerator to the Gemfile since that will be managed by PR #168, but perhaps the Rakefile addition will still be useful? I found it quite helpful personally.

glennsarti:
@MosesMendoza Bad news-ish.  I'm having MAJOR problems with my PR due to testing against ruby 1.8.7.  Most everything no longer supports it.  I'm +1 to getting this merged in the meantime

DavidS:
where are we still using/supporting ruby 1.8.7?

glennsarti:
This is a bit of a special case.  This module is used to upgrade from older PE and FOSS agents to a more modern agent.  Therefore we need to be able to "support"/test on old agent versions (The module supports down to 3.7.0).  Thus the need for ruby 1.8.7 testing.  The documentation states Puppet dropped 1.8.7 support in Pup 4.x.

DavidS:
Good luck then. We gave up 1.8.7 testing, even for stdlib, since everyone
else has jumped already.

Your best bet (in terms of functionality and sanity) is probably doing a
big `if RUBY_VERSION` to completely change your Gemfile when running on 1.8.

MosesMendoza:
Just to confirm - this PR is separate from issues surrounding 1.8.7 support and just allows us to create a nodeset for use with beaker acceptance via `rake beaker`

highb:
👍  sounds like you're all 👍  on this as well so I'm going to get it merged

PR #170, opened 2016-10-07
(MODULES-3951) Add explicit check for stringify_facts

binford2k:
Without this check, you get incomprehensible results. The is_pe fact is
evaluated as true, and so the class bails out. In some cases the
info() function isn't loud enough and so it appears to do nothing.

This just fails hard and makes it obvious why.

### DIFF
--- a/lib/facter/settings.rb
+++ b/lib/facter/settings.rb
@@ -12,6 +12,12 @@ Facter.add('puppet_config') do
   end
 end
 
+Facter.add('puppet_stringify_facts') do
+  setcode do
+    Puppet.settings['stringify_facts'] || false
+  end
+end
+
 Facter.add('puppet_sslpaths') do
   setcode do
     result = {}
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,6 +5,10 @@
 #
 class puppet_agent::params {
 
+  if $::puppet_stringify_facts {
+    fail('The puppet_agent class requires stringify_facts to be disabled')
+  }
+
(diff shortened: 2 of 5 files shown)


glennsarti:
Would be nice to get a spec test for this.

binford2k:
@glennsarti I'd be happy to, but I couldn't think of a non-trivial test, and I haven't spent time figuring out how to set up a beaker test with config settings. :-/

glennsarti:
A simple spec test should be sufficient.  Mock the facts and expect a fail.

branan:
This was already done (albeit less-cleanly) here: https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/prepare/ssl.pp#L48-L50

This PR should probably also remove that check, since `stringify_facts` will now explicitly be checked

glennsarti:
Is it possible for that code path to be triggered by something other than stringify = true?

branan:
@glennsarti Only if a user also has their own custom fact called `puppet_sslpaths` that isn't a hash, or if pluginsync has somehow failed - both cases that I'm comfortable not worrying too much about.

https://github.com/puppetlabs/puppetlabs-puppet_agent/commit/6fb63a0b6eaba445a69d62fbe9c4fd0870324504 is the actual commit - the entire `case` is just to check if the fact is a hash.

glennsarti:
Excellent...I can revert those.  Thanks

glennsarti:
@branan travis has completed with the changes you requested.  Can you please review once more and approve once you're happy.  Thanks.

PR #172, opened 2016-10-10
(MODULES-3912) Remove POWER8 hardware constraint from README

geoffnichols:
Starting with the 1.3.0 release of the puppet_agent module, use of the
module is no longer constrained by the PowerPC architecture version of
the target system's underlying hardware.

Here we update README.markdown to reflect this change.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -169,7 +169,6 @@ In addition, there are several known issues with Windows:
  
 Specifically in the 1.2.0 Release:
 * For Windows, you must trigger an agent run after upgrading so that Puppet can create the necessary directory structures.
-* AIX package names are based on PowerPC architecture version. PowerPC 8 is not yet supported.
 
 ## Development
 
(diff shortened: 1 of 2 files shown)


geoffnichols:
@glennsarti and @MosesMendoza

README.markdown
```diff
@@ -167,9 +167,8 @@ In addition, there are several known issues with Windows:
 * On Windows Server 2003, only x86 is supported, and the `arch` parameter is ignored. If you try to force an upgrade to x64, Puppet installs the x86 version with no error message.
 * On Windows Server 2003 with Puppet Enterprise, the default download location is unreachable. You can work around this issue by specifying an alternate download URL in the `source` parameter.
  
-Specifically in the 1.2.0 Release:
+Specifically in the 1.3.0 Release:
```
glennsarti: Not sure if it's worth changing this

geoffnichols: Ah, I wasn't sure when that bit should be updated. Just pushed an update that leaves that alone for now.

geoffnichols:
Added a commit that updates the AIX spec test to use `7.1` instead of `7.2`, since we have not yet announced puppet-agent support for AIX 7.2.

PR #173, opened 2016-10-11
(MODULES-3872) add acceptance test for manage_repo param

MosesMendoza:
This commit adds tests to the acceptance test suite in `class_spec.rb` for
confirming that the manage_repo parameter does, in fact, do what it says.
Specifically, we test that puppet created the repo by applying the
'puppet_agent' class with no params (implying default value of true for
`manage_repo`) and then specifically ensuring the repo present with puppet and
catching (failing) on any resultant changes. Then we remove the repo
configuration and apply the puppet agent class again but with `manage_repo =>
false`, and ensure that the repo configs are not re-created (again by catching
changes).

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -39,6 +39,45 @@ describe 'puppet_agent class' do
         is_expected.to_not match /parser[ ]*=[ ]*future/
       }
     end
+
+    describe 'manage_repo parameter' do
+      context 'when true (default)' do
+        it 'should create repo config' do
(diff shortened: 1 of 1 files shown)


MosesMendoza:
Relevant output from a run against a rhel agent: https://gist.github.com/MosesMendoza/c35cc6d5b3b9fcdb21924161c88b214c

MosesMendoza:
Relevant output from a run against an ubuntu agent: https://gist.github.com/MosesMendoza/5a81d5a2842598a5f195c5db99831ccb

PR #176, opened 2016-10-11
(maint) Pin stdlib module to 4.12

Magisus:
The 4.13 release of the puppetlabs-stdlib module deprecates many
functions, breaking all of the tests in the puppet_agent module. Until
we can update the tests to handle the deprecations, we need to stay on
4.12.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -120,7 +120,7 @@
     }
   ],
   "dependencies": [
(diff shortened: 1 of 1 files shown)


Magisus:
This needs to be fixed in the spec helper first, because we are unsure if using the new stdlib actually breaks functionality, rather than just tests.

PR #179, opened 2016-10-13
(FM-5317) Prep for 1.3.0 release

glennsarti:
Prepare the module for a 1.3.0 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,49 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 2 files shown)


Magisus:
Looks good, very thorough 👍

MikaelSmith:
Should we update supported platforms for those that have agent packages? Like Fedora 23, Ubuntu 16.04.

glennsarti:
@MikaelSmith No commits since 1.2.0 have modified the metadata.json file with additional versions and I didn't see any obvious commits that explain that these platforms are now supported.  If you want to quickly add an additional PR for it I can then update the changelog.

PR #182, opened 2016-10-21
(MODULES-3994) Fix incorrect condition so services are created

seanhoughton:
This fixes a typo that prevents the service resources from being created

### DIFF
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -109,7 +109,7 @@ describe 'puppet_agent' do
           end
         end
 
-        [{}, {:service_names => []}].each do |params|
+        [{}, {}].each do |params|
           context "puppet_agent class without any parameters" do
             let(:params) { params.merge(global_params) }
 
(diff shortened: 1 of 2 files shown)


seanhoughton:
Hello, I'm reasonably sure that the fix is valid - but I've exhausted my time budget for working out the unit tests today. If I can get some time next week I'll give it another try, but this seems like a fairly serious bug because it results in services not getting installed so maybe someone else can have a look.

MosesMendoza:
@seanhoughton thank you for filing this! I created a ticket, https://tickets.puppetlabs.com/browse/MODULES-3994, to track resolution of this. Agreed this bug needs fixed

seanhoughton:
I'm not sure why the 4.0.0 tests are failing but I'm suspicious of the version comparisons. All the tests pass for me locally running inside a container...

`docker  run --rm -ti -v .:/agent ruby:2.1.10 bash`

spec/classes/puppet_agent_spec.rb
```diff
@@ -109,7 +109,7 @@ def global_facts(facts, os)
           end
         end
 
-        [{}, {:service_names => []}].each do |params|
+        [{}, {}].each do |params|
```
MikaelSmith: Why run this twice now? This was also meant to test with services managed and unmanaged.

MikaelSmith:
Tests are failing. Is that something you're looking into?

MikaelSmith:
Fixed up in https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/209.

PR #183, opened 2016-10-24
(maint) convert install_puppet.bat to CRLF

ferventcoder:
Use CRLF instead of just LF for the template. It is not
known if using LF will cause issues with Batch files in
any versions of Windows. It is thought that running a
batch with LF should work, but it is possible that there
may be some issues somewhere. It's typically better to be
on the safe side and ensure the file is CRLF. Newer versions
of Puppet should leave the EOL format the same.


ferventcoder:
@glennsarti @MikaelSmith let me know if there is already a reason we went with LF.

MikaelSmith:
I'm not aware of one.

ferventcoder:
@joshcooper ^^

ferventcoder:
@Iristyle ^^

joshcooper:
I don't think there are any issues running with LF on windows. Changing to CRLF will make it easier to load the resulting `bat` file using windows notepad. That said, there could be other parts of the CI pipelines that don't preserve CRLF when doing a `git clone`. IOW, even if you change this file, the resulting file on disk after MSI install may have lost the `CR` along the way, depending on the git config of the builder host that runs vanagon.

ferventcoder:
Fair enough. I'm happy with this though. 👍

Iristyle:
Yeah, I'm not sure this will have an effectual change at all... unless you were also to add `.gitattributes` containing:

```
install_puppet.bat	text eol=crlf
```

ferventcoder:
Well as far as the commit having changes, you can see it did based that there are changes. However it may show differently to you if you don't have commit as is. That is a local configuration though, the PR showed the differences. :D

ferventcoder:
Unless I'm missing some context @Iristyle :)

PR #187, opened 2016-11-11
(MODULES-4092) Install solaris 10 package per-zone

branan:
Solaris 10 packaging, when combined with zones, is super weird. The
Puppet Enterprise frictionless install has historically made sure that
the agent is installed per-zone. However, the `puppet_agent` module
did not. This leads to conflicts when upgrading agents - once the
Global zone is updated, trying to upgrade the other zones will fail,
since the `puppet-agent` package is "already installed".

This changes solaris 10 packaging to be per-zone, which matches how we
have historically done that deployment.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -41,9 +41,10 @@ class puppet_agent::install(
 
     $_unzipped_package_name = regsubst($package_file_name, '\.gz$', '')
     $_package_options = {
-      adminfile => '/opt/puppetlabs/packages/solaris-noask',
-      source    => "/opt/puppetlabs/packages/${_unzipped_package_name}",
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This seems sane. Whatever the frictionless installer does probably makes sense. I'm not aware of anyone who wants to install it for all zones, but that's the scenario where I could see this surprising someone.

MikaelSmith:
@highb have any concerns to add on this? Otherwise I think we can merge it?

highb:
@MikaelSmith I'm not super familiar with Solaris zones, but making the module consistent with the frictionless installer seems like the right thing to do so 👍  from me.

PR #188, opened 2016-11-12
(maint) Fix config for Jenkins

MikaelSmith:
Module config got out-of-sync with expectations in our Jenkins setup.
Fix it to use PUPPET_GEM_VERSION to select Puppet version for testing,
and change the default to a more recent version of Puppet.


highb:
👍

PR #198, opened 2016-12-12
(MODULES-4207) Optionally move puppetres.dll on Windows upgrade

glennsarti:
During investigation into network stack restarts on Windows (https://tickets.puppetlabs.com/browse/PA-663),
it was found that a file lock on puppetres.dll was causing MSI to trigger an
unforseen service restart. This commit adds an installation option to the
puppet_agent module called `msi_move_locked_files` which premptively moves files
that can cause locks, prior to the upgrade, so that the upgrade can complete
without a service restart.

- [X] Add spec tests
- [X] ~~Add acceptance tests~~
- [X] Update README

Acceptance tests will be moved to another ticket due to complexity.

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -40,6 +40,34 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
+<% if @msi_move_locked_files %>
+REM Move puppetres.dll to avoid file locks and service restarts (MODULES-4207)
+
+REM Puppet-Agent, less than 1.6.0, location
+SET PUPPETRES=<%= @env_windows_installdir %>\misc\puppetres.dll
+SET RANDFILE=%TEMP%\puppetres-%RANDOM%-1.DLL
+IF EXIST "%PUPPETRES%" (
+  MOVE /Y "%PUPPETRES%" "%RANDFILE%"
+  DEL /Q "%RANDFILE%"
+
+  IF EXIST "%PUPPETRES%" (
+    ECHO *** WARNING - Unable to move %PUPPETRES%
+  )
+)
+
+REM Puppet-Agent, 1.6.0 and above, location
+SET "PUPPETRES=<%= @env_windows_installdir %>\puppet\bin\puppetres.dll"
+SET RANDFILE=%TEMP%\puppetres-%RANDOM%-2.DLL
+IF EXIST "%PUPPETRES%" (
+  MOVE /Y "%PUPPETRES%" "%RANDFILE%"
+  DEL /Q "%RANDFILE%"
+
+  IF EXIST "%PUPPETRES%" (
+    ECHO *** WARNING - Unable to move %PUPPETRES%
+  )
+)
+<% end %>
+
 start /wait msiexec.exe /qn /norestart /i "<%= @_msi_location %>" /l*vx "<%= @_logfile %>" PUPPET_MASTER_SERVER="<%= @_puppet_master %>" PUPPET_AGENT_ENVIRONMENT="%environment%" <% unless @install_dir.to_s.empty? -%>INSTALLDIR="<%= @install_dir %>"<% end -%>
 
 :End
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ spec/fixtures
 vendor
 log
 tmp
+Gemfile.local
+Gemfile.lock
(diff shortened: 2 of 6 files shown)


glennsarti:
Rationale to why this implementation method was chosen:

*Allow a user to specify arbitrary batch code*

One way would be to add a parameter to the puppet_agent install class which would pass through arbitrary batch commands into the `puppet_agent_windows_install_spec.tb`  e.g.

``` puppet
class { '::puppet_agent':
...
  batch_command => 'DEL C:\windows\somefile.txt',
...
}
```

This would give the user an incredible amount of flexibility however adding multiple lines, text encoding and line endings would be difficult to diagnose.  Given that this PR is only being introduced to fix an uncommon bug, it seemed far simpler to add the required functionality behind a simple boolean flag.

*Use puppet catalog to force an Exec prior to upgrade*

e.g. add an Exec into the manifest

``` puppet
exec { 'preflight-windows-upgrade':
    command => <bunch on commands>,
    ...
    before  => Exec['install_puppet.bat'],
  }
```

This was tried and mostly worked however because `Exec['install_puppet.bat']` is not included in the catalog if no upgrade is required, it means that the introduced exec must also be guarded.  This can be a simple `if` statement (e.g. `if ($::puppetversion != '4.5.2') and ($::operatingsystem == 'windows') {`) however they tend to be fragile and can cause a lot of false compilation errors

glennsarti:
@puppetlabs/windows Raised an early PR for review.  Thoughts?

MosesMendoza:
Couple questions - 

Is there any case when the prerequisite criteria are met (ie puppet version/os) where we would _not_ want to move the files (ie where we would not set the flag to true)?

Is this logic that could just live in the MSI itself? pros/cons?

glennsarti:
@MosesMendoza 

1.  Given that we're doing a MSI install immediately after I can't think of any case where moving that file is a "bad thing".  Worst case it will populate the `%TEMP%` directory with files that need to be cleaned up.  In _theory_ this could fill up a hard disk over time, but given it's a < 4K file and puppet runs every 30mins by default, that's a LONG time to fill the disk.

2. Unfortunately this PR is not the best place to discuss all the things I've tried RE MSIs.  https://tickets.puppetlabs.com/browse/PA-663 contains all of the things I tried, including merging this into the MSI itself.  I've also raised https://tickets.puppetlabs.com/browse/PA-768 to mark this file as permanent which "removes" this issue.

templates/install_puppet.bat.erb
```diff
@@ -40,6 +40,34 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
+<% if @msi_move_locked_files %>
+REM Move puppetres.dll to avoid file locks and service restarts (MODULES-4207
+)
```
MikaelSmith: Is this parenthesis meant to be on the following line?

templates/install_puppet.bat.erb
```diff
@@ -40,6 +40,34 @@ REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
 
+<% if @msi_move_locked_files %>
+REM Move puppetres.dll to avoid file locks and service restarts (MODULES-4207
+)
+REM Puppet-Agent, less than 1.6.0, location
+SET PUPPETRES=<%= @env_windows_installdir %>\misc\puppetres.dll
+SET RANDFILE=%TEMP%\puppetres-%RANDOM%-1.DLL
```
MikaelSmith: Why use a random name, rather than just always overwrite the same name?

templates/install_puppet.bat.erb
glennsarti: Typo. Fixed.

templates/install_puppet.bat.erb
glennsarti: Because if we can't move a file to a name that already exists.  If the file we're moving to exists and cannot be deleted we're in a bad situation.  In this way we mostly guarantee that there will be no existing file with that name.

MikaelSmith:
I don't really have objections to this. I wish we could fix it in the MSI, but comfortable if we can't. Does Windows clean up TEMP on a regular schedule (pretty sure it doesn't, but it'd be a nice surprise)?

glennsarti:
@MikaelSmith PA-768 has been raised to modify the MSI but it will take two Puppet Agent releases for it to become effective.  But that will not solve any upgrades from previous versions.

Unfortunately, TEMP is not purged however this is a 1KB file so it's not that bad, and the temp filename does indicate which application put it there.

glennsarti:
Rebased.

PR #199, opened 2016-12-21
Use getvar for facts that doesn't exist in my environment

TomaszUrugOlszewski:
After fixing those 3 facts puppet module started to work with strict_variables set to true


MikaelSmith:
@highb we can use 'squash and merge' to fix up the commit title.

highb:
@MikaelSmith 👍 on that plan from me

PR #200, opened 2016-12-21
(MODULES-4241) Add custom fact puppet_agent_appdata

MikaelSmith:
Module params depended on `::common_appdata` being defined, which exists
in PE but not necessarily other environments. Add a new custom fact - so
names don't conflict - to cover this case.


highb:
👍  specs pass and I approve of reducing the dependence on PE variables.

PR #202, opened 2016-12-22
(maint) Only update server.cfg if not already managed

### LINKED ISSUE #193: does not work with PE modules
There is a direct duplicate resource collision that is happening because the puppet_enterprise module is already managing `Service['mcollective']` and `File['/etc/puppetlabs/mcollective/server.cfg']` which this module tries to manage as well.  Anybody use PE and the modules it comes with will most likely see this error.

While you can override the service_names to only include `[puppet]` there is no way to stop managing the mco config files.

I think to resolve this issue one of the following should be done:

1.  use the defined function to determine if the resource is already managed  `unless defined(File['/etc/puppetlabs/mcollective/server.cfg'])`  

2.  The module is called puppet_agent and should not have anything to do with mcollective, remove the mcollective portion alltogether

3. Add a feature flag to allow the user to manage mcollective  (services, files)

MikaelSmith:
PE modules also manage server.cfg, so only manage it if not already done
so by other modules. Fixes #193.

### DIFF
--- a/manifests/prepare/mco_server_config.pp
+++ b/manifests/prepare/mco_server_config.pp
@@ -8,48 +8,50 @@ class puppet_agent::prepare::mco_server_config {
   assert_private()
 
   $mco_server = $::puppet_agent::params::mco_server
-  file { $mco_server:
-    ensure => file,
-    source => $::mco_server_config,
-  }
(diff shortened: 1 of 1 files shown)


highb:
👍

PR #203, opened 2017-01-04
Ensure all variables are populated

rnelson0:
Prevent failures when STRICT_VARIABLES='yes'.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -23,6 +23,8 @@ class puppet_agent::params {
       true    => "https://${::servername}:8140/packages",
       default => undef,
     }
+  } else {
+    $_source = undef
   }
 
   $package_name = 'puppet-agent'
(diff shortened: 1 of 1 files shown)


hunner:
👍

hunner:
Why is travis not testing strict variables?

hunner:
Actually, strict variables may be on for 4.x of puppets in rspec-puppet, but they won't test on windows anyway...

manifests/params.pp
```diff
@@ -24,6 +24,9 @@
       default => undef,
     }
   }
+  else {
```
hunner: This should be on one line: `} else {`

rnelson0:
Thanks @MikaelSmith! Any idea on when a new release might be cut? Just planning for work :)

MikaelSmith:
I don't off the top of my head, but given the number of items stacking up under 1.3.2 sometime in the next month seems reasonable.

rnelson0:
Thanks! I just remembered the RSS feeds and will have to figure out a way
to pipe that to email. Cheers!

PR #204, opened 2017-01-09
(MODULES-4214) Add additional installation parameters during upgrade

glennsarti:
Previously it was not possible to pass additional installation parameters during
an upgrade process.  In the case of MSI files this could be sensitive passwords
or other public MSI properties.  In the case of package resources, additional
options could not be added to the package resource.

This commit adds a parameter called `install_options` which mirrors the
`install_options` parameter from the package resource and then passes this down
so that it is invoked on installation (either by a package resource or in the
batch file to install the MSI)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -151,6 +151,14 @@ The directory the puppet agent should be installed to. This is only applicable f
   install_dir => 'D:\Program Files\Puppet Labs'
 ```
 
+##### `install_options`
+
+An array of additional options to pass when installing puppet-agent. Each option in the array can either be a string or a hash. Each option will automatically be quoted when passed to the install command. With Windows packages, note that file paths in an install option must use backslashes. (Since install options are passed directly to the installation command, forward slashes won't be automatically converted like they are in `file` resources.) Note also that backslashes in double-quoted strings _must_ be escaped and backslashes in single-quoted strings _can_ be escaped.
+
+``` puppet
+  install_options => ['PUPPET_AGENT_ACCOUNT_DOMAIN=ExampleCorp','PUPPET_AGENT_ACCOUNT_USER=bob','PUPPET_AGENT_ACCOUNT_PASSWORD=password']
+```
+
 ##### `msi_move_locked_files`
 
 This is only applicable for Windows operating systems. There may be instances where file locks cause unncessary service restarts.  By setting to true, the module will move files prior to installation that are known to cause file locks. By default this is set to false.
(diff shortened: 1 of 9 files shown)


spec/classes/puppet_agent_spec.rb
```diff
@@ -110,6 +110,21 @@ def global_facts(facts, os)
         end
 
         [{}, {:service_names => []}].each do |params|
+          context "puppet_agent class with install_options" do
+            let(:params) { global_params.merge(
+              {:install_options => ['OPTION1=value1','OPTION2=value2'],})
+            }
+
+            it { is_expected.to compile.with_all_deps }
```
MikaelSmith: This test doesn't seem to be working right.

glennsarti: Still a work in progress.  Running spec test on Windows is difficult for this module.

glennsarti:
Rebased all the bad commit messages.  Still need to update the Readme

MikaelSmith:
Is this finished? Didn't realize it'd been updated.

README.markdown
```diff
@@ -151,6 +151,14 @@ The directory the puppet agent should be installed to. This is only applicable f
   install_dir => 'D:\Program Files\Puppet Labs'
 ```
 
+##### `install_options`
+
+An array of additional options to pass when installing a package. These options are package-specific, and should be documented by the software vendor. One commonly implemented option is `INSTALLDIR` Each option in the array can either be a string or a hash, where each key and value pair are interpreted in a provider specific way. Each option will automatically be quoted when passed to the install command. With Windows packages, note that file paths in an install option must use backslashes. (Since install options are passed directly to the installation command, forward slashes won't be automatically converted like they are in `file` resources.) Note also that backslashes in double-quoted strings _must_ be escaped and backslashes in single-quoted strings _can_ be escaped.
```
MikaelSmith: Since we control the package, shouldn't we just point to where these options are defined (rather than say they "documented by the software vendor")?

glennsarti: Whoops.  Copy-pasta fail from the package provider

glennsarti:
@MikaelSmith Documentation updated.

manifests/init.pp
```diff
@@ -35,6 +35,18 @@
 #   The directory the puppet agent should be installed to. This is only applicable for
 #   windows operating systems. This only applies when upgrading the agent to a new
 #   version; it will not cause re-installation of the same version to a new location.
+# [install_options]
+#   An array of additional options to pass when installing a package. These
```
MikaelSmith: Here too?

glennsarti: Nice pickup.  Thanks.  Fixed.

PR #205, opened 2017-01-13
(MODULES-4424) Add skip_if_unavailable to yumrepo resource

walkamongus:
The skip_if_unavailable yum option allows unrelated yum package
management to continue in the event that the agent repository
is unavailable for any reason. It is defaulted to 'absent', which means
the option is not set in the repository file.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -60,6 +60,7 @@ class puppet_agent (
   $install_dir           = $::puppet_agent::params::install_dir,
   $disable_proxy         = false,
   $install_options       = $::puppet_agent::params::install_options,
+  $skip_if_unavailable   = 'absent',
   $msi_move_locked_files = false,
 ) inherits ::puppet_agent::params {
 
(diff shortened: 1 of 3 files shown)


walkamongus:
rebased onto latest

walkamongus:
Issue opened at: https://tickets.puppetlabs.com/browse/MODULES-4424

manifests/init.pp
```diff
@@ -60,6 +60,7 @@
   $install_dir           = $::puppet_agent::params::install_dir,
   $disable_proxy         = false,
   $install_options       = $::puppet_agent::params::install_options,
+  $skip_if_unavailable   = 'absent',
```
MikaelSmith: Maybe we should have a yumrepo hash, that lets you override any particular parameters to setting up the repo... although that also sounds terrible.

walkamongus: Seems like this would be easy to do with the splat operator, but we'd have to wait until Puppet 3 support is dropped.

CLA signed.

caseywilliams:
Thanks, @walkamongus! This looks good. We'll rebase this change against the 1.x branch and merge.

speedofdark:
This is rebased against 1.x and replaced by #296.

PR #209, opened 2017-01-31
(MODULES-3994) Manage services on Puppet 4

MikaelSmith:
Previously a bug in the logic including the puppet_agent::service class
prevented it from being applied on non-PE installs using Puppet 4. Fix
it so that services are now managed for non-PE installs under Puppet 4.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -147,7 +147,7 @@ class puppet_agent (
 
     # On windows, our MSI handles the services
     # On PE AIO nodes, PE Agent nodegroup is managing the services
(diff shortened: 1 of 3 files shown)


MikaelSmith:
What does the "in progress" label mean?

highb:
Oh sorry, that's an artifact of the [waffle.io](https://waffle.io/) board I'm getting Integration to use to track PRs across multiple repos. I removed that part of the workflow, so hopefully it won't show up any more. :)

highb:
Feel free to remove those if I miss any.

PR #207, opened 2017-01-30
{WIP}(maint)(MODULES-3710) Apply module sync configs to puppet_agent module and fix strict variable tests

glennsarti:
Previously the puppetlabs-puppet_agent module was not under the control of the module sync (modsync) configuration tool used by the Modules team.  This Pull Request starts by bringing the module under the control of modsync and then applies a series of commits to fix up any configuration or test errors.

In particular there were many strict variable test failures when running spec tests.

PR #211, opened 2017-02-01
(maint) Remove require to vendored gem

demophoon:
PUP-7114 moved the semantic vendored library which broke this require.
We shouldn't be referencing the vendored libraries at all since their
API's can change without warning (as has happened in the 4.9.0 version
of puppet).

Since nothing important uses this library in the Rakefile it is safe to
remove this line.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,7 +6,6 @@ if Puppet.version.to_f < 4.0
 end
 
 require 'puppetlabs_spec_helper/rake_tasks'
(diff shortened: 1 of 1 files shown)


MikaelSmith:
That's crazy. @glennsarti want to take a look at whether this is handled in your PR?

MosesMendoza:
related to https://tickets.puppetlabs.com/browse/PUP-7156

glennsarti:
@MikaelSmith Yes PR #207 removes that Gem

demophoon:
Closing in favor of #207

Issue #132, opened 2016-07-13
Forcing apt-update on Debian causes hosts to be considered 'Changed'


gfokkema:
When using this module in conjunction with Puppetboard, using `notify` to enforce an apt-update causes a hosts to consider itself to have changes.
See [code](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/osfamily/debian.pp#L93).

The enforced apt-update seems to apply as well when the puppet agent is already at the target version.
Therefore, when using this module, every host in my environment is reporting to have applied changes every puppet run.

MikaelSmith:
May be able to change that to a subscribe. It should happen only if the repo config changes, or if a newer version is specified in the class parameters.

Hakon:
Wouldn't it suffice to let the package resource require => Apt::Source['pc_repo'] on debian?

MikaelSmith:
I think there was previously an issue in how containment worked that made the refresh happen too late. I'll have to retest it.

jghward:
I worked around this in the above commit by changing the Notify resource to an Exec, allowing me to use the refreshonly param. Happy to submit as a pull request if it's an acceptable solution.

MikaelSmith:
That seems interesting, I'd love to see it.

MikaelSmith:
Also filed as [MODULES-4171](https://tickets.puppetlabs.com/browse/MODULES-4171).

sigv:
[apt::source](https://forge.puppet.com/puppetlabs/apt/readme#defined-type-aptsource) appears to have `notify_update` since 2.3.0 - that is the latest release, released in September 2016. Why not use that instead of this custom notify/exec flow?

MikaelSmith:
I wasn't aware of it, I'll take a look. Thanks!

sigv:
Behind the scenes, it seems to just add a `notify => Class['apt::update']` to the sources.list file and I haven't tested out how will fare here, but it seems to be made just for the use case here - running `apt update` after having added the source..

PR #217, opened 2017-03-11
(Issue #132) Change Notify resource to Exec

jghward:
This resource is Notified, so change it to a resource type that is refreshable. Set refreshonly to true.


MikaelSmith:
This seems straight-forward, thanks! Going to double check system tests still work before merging it.

PR #220, opened 2017-03-15
(MODULES-4547) setting the package provider to sun for Solaris 10 hosts in case the …

sirinek:
…default was changed

setting the package provider to sun for Solaris 10 hosts in case the default was changed.
our site default for the package provider on Solaris 10 hosts is 'pkgutil' and that causes this module to not work. since the sun provider works correctly for the puppet-agent package, i forced the package resource to set the provider to 'sun' even though it's the default for most sites.


MikaelSmith:
This seems to make sense.

PR #224, opened 2017-04-01
Don't include apt

johnf:
Using ```include ::apt``` breaks anyone specifying ```Class { 'apt':```


MikaelSmith:
Right now you can use `::puppet_agent::manage_repo` to avoid this problem. What I'd probably like to do is move to using the deb packages hosted by apt.puppet.com instead of the `apt` class.

Also, I really wish Puppet would allow these kinds of multiple inclusion.

MikaelSmith:
Another common way to work around this is to use `include ::apt`, then use Hiera to set default class parameters.

johnf:
Assuming this won't be merged

PR #233, opened 2017-05-23
(DO NOT MERGE)[#puppethack] Ticketnr MODULES-4839

lbayerlein:


glennsarti:
As per comment on the ticket:
---
The puppet_agent module is specifically designed to move people from older agents to newer ones. In this case I don't think the version requirement makes sense as it defeats the purpose of the module; to help people update their puppet agents.

MikaelSmith:
Agreed.

PR #234, opened 2017-06-06
(PA-1160) Add support for AIX 7.2

branan:
We did not add new packages for AIX 7.2. Instead, we use the AIX
7.1 packages. This adds the necessary logic to point AIX 7.2 at
the older packages.


branan:
This is wrong. I'm fixing it up (I hope)

PR #236, opened 2017-07-08
puppet_stringify_facts fixes for Puppet 4

lukebigum:
Remove default value for Fact ::puppet_stringify_facts and allow Puppet version being tested to provide it (thus simulating it's non-existence in Puppet 4).

Fixing the logic around testing for $::puppet_stringify_facts in Puppet 4.


MikaelSmith:
That looks great, thanks!

PR #237, opened 2017-07-17
(maint) Include spec_helper so parallel specs succeed

MikaelSmith:
All spec files need to include spec_helper, or parallel specs runs will
fail.

### DIFF
--- a/spec/classes/coverage_spec.rb
+++ b/spec/classes/coverage_spec.rb
@@ -1,1 +1,3 @@
+require 'spec_helper'
+
 at_exit { RSpec::Puppet::Coverage.report! }


glennsarti:
Are we able to run this through adhoc pipelines? Travis doesn't run specs in parallel.  So I can't be sure this actually fixes the problem

MikaelSmith:
Hmm, not clear from a local run. I guess I'll try to figure that out.

glennsarti:
Found the adhoc pipelines, running it through now

glennsarti:
Adhoc shows this fix works.  The error in the Travis CI job is a known failure and should not stop merging this fix.

PR #239, opened 2017-07-17
(maint) Use http for RedHat repos

MikaelSmith:
RedHat 5 uses an old OpenSSL version that can't connect to
www.puppetlabs.com using HTTPS. Use HTTP - which is what the repos setup
by Puppet's repo packages use - instead. That should still be secure
since we use signed packages and the GPG keys are included with this
module.


glennsarti:
Merging with failed Travis on ~>4.0.0.  Expected.

PR #240, opened 2017-07-18
(maint) bump inifile dependency

eputnam:
inifile 2.0.0 is on its way so this modules dependency needs to be updated to use the new version


MikaelSmith:
Thanks!

PR #241, opened 2017-07-27
(MODULES-3787) Fix upgrades on Solaris 10 when initiated from service

MikaelSmith:
Due to PUP-7110, Solaris 10 upgrades fail when triggered by the service
when starting with Puppet prior to 5.0.0. For those upgrades, use a
script to upgrade after the Puppet run has completed.

Add Puppet 5 to Travis CI runs. Also update to latest patched version of
Ruby 2.3.

Uses the env_temp_variable fact - instead of hard-coding `/tmp` - for
the solaris_start_puppet script. Also redirect output to a log file for
the backgrounded script execution. Ensure these steps are only taken
when we just upgraded, to avoid unnecessary "changes".

Rearranged the service management logic for Solaris to be closer to
other relevant operations. In particular, moved stopping old PE service
names to package removal for Solaris 10.

Move package installation into a single conditional chain. Previous two
chains made it difficult to handle new cases and separating out options
and extra steps from install didn't simplify code.


ScottGarman:
Other than the linting issues Travis is complaining about, LGTM so long as it blends. :+1:

MikaelSmith:
Re-opened from a different branch that's easier to test against: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/242

PR #242, opened 2017-07-27
(MODULES-3787) Fix upgrades on Solaris 10 when initiated from service

MikaelSmith:
Due to PUP-7110, Solaris 10 upgrades fail when triggered by the service
when starting with Puppet prior to 5.0.0. For those upgrades, use a
script to upgrade after the Puppet run has completed.

Add Puppet 5 to Travis CI runs. Also update to latest patched version of
Ruby 2.3.

Uses the env_temp_variable fact - instead of hard-coding /tmp - for
the solaris_start_puppet script. Also redirect output to a log file for
the backgrounded script execution. Ensure these steps are only taken
when we just upgraded, to avoid unnecessary "changes".

Rearranged the service management logic for Solaris to be closer to
other relevant operations. In particular, moved stopping old PE service
names to package removal for Solaris 10.

Move package installation into a single conditional chain. Previous two
chains made it difficult to handle new cases and separating out options
and extra steps from install didn't simplify code.


manifests/install.pp
```diff
@@ -37,21 +36,73 @@
...
+          ensure  => file,
+          mode    => '0755',
+          content => epp('puppet_agent/solaris_install.sh.epp', {
+                          'old_package_names' => $old_package_names,
+                          'adminfile'         => $adminfile,
+                          'sourcefile'        => $sourcefile,
+                          'install_options'   => $install_options})
+        }
+        -> exec { 'solaris_install script':
+          command => "${_installsh} ${::puppet_agent_pid} 2>&1 > ${_logfile} &",
```
MikaelSmith: Need to use the helper to run in a separate contract.

MikaelSmith:
This now passes upgrade integration tests.

manifests/install.pp
```diff
@@ -64,32 +115,30 @@
...
       path        => '/bin:/usr/bin:/sbin:/usr/sbin',
       refreshonly => true,
     }
-
-    $_package_options = {
-      require => Exec['puppet_agent restore /etc/puppetlabs'],
-      notify  => Exec['puppet_agent post-install restore /etc/puppetlabs'],
-      install_options => $install_options,
-    }
   } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /10\.[9,10,11]/ {
```
mruzicka: this regexp looks totally bogus to me, IMO it should read:
```
/10\.(9|10|11)/
```

mruzicka: Perhaps it should use some anchors too:
```
/^10\.(9|10|11)/
```

manifests/install/remove_packages.pp
```diff
@@ -36,6 +36,15 @@
...
       if $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '10' {
+        # Tidy up pe-puppet and pe-mcollective services that would be left running
+        # after package removal.
+        service { 'pe-mcollective':
+          ensure  => stopped,
+        }
+        service { 'pe-puppet':
+          ensure => stopped,
+        }
+
```
mruzicka: Does stopping of services work after the respective packages have been removed?

MikaelSmith: That's an excellent question. It's happening before now, but used to "happen" after. Not sure if it ever succeeded.

MikaelSmith:
This is waiting on an integration test change to ensure it doesn't break existing tests.

manifests/install.pp
```diff
@@ -20,8 +20,7 @@
 
   $old_packages = (versioncmp("${::clientversion}", '4.0.0') < 0)
 
-  if ($::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10') or
-        $::operatingsystem == 'AIX' {
+  if ($::operatingsystem == 'SLES' and $::operatingsystemmajrelease == '10') or $::operatingsystem == 'AIX' {
```
mruzicka: Is the closing of the parentheses before the `or` intentional. Note that it does not change the condition.

MikaelSmith: It's intentional (and was there before) just to clarify precedence.

manifests/install.pp
```diff
@@ -37,21 +36,73 @@
...
-    $_package_options = {
+
+    # package provider does not provide 'versionable'
+    package { $::puppet_agent::package_name:
+      ensure          => 'present',
       provider        => 'rpm',
       source          => "/opt/puppetlabs/packages/${package_file_name}",
       install_options => $_install_options,
     }
   } elsif $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '10' {
```
mruzicka: Why is this only needed on Solaris 10?
I'd assume the same approach much be taken on Solaris 11 as it uses contracts too.

manifests/install.pp
mruzicka: Ah, ok.

manifests/install.pp
MikaelSmith: There are major differences in how we do the upgrade between 10 and 11. I believe 11 uses upgrade, instead of uninstall/reinstall. It's quite possible our upgrade process doesn't restart the service until after upgrade.

MikaelSmith:
This is unblocked.

manifests/install.pp
MikaelSmith: There is a possibility the upgrade is succeeding but the Puppet process is killed immediately after. I'll try to look into that.

MikaelSmith: So doing the upgrade does TERM the process that triggered the upgrade. Since Puppet then runs again and succeeds, and this is fixed in a future Puppet version, I'm tempted to ignore it (or at least treat it as a separate ticket).

branan:
Looks like this got caught up in the Travis precise -> trusty transition. Would be nice to get that fixed and then rebase this so we get coverage on those older puppet versions

MikaelSmith:
@branan I'm not sure I follow. Can you expand on what you said?

branan:
@MikaelSmith Travis is migrating their test VMs from Ubuntu Precise to Ubuntu Trusty. In this particular case, it looks like there aren't Ruby 2.1 builds for Trusty. That means if we want to keep testing that ruby version for coverage of older puppet releases in the spec tests, we'll need to pin our Travis configuration back to Precise.

Once we do that, we should be able to rebase this PR and see Travis go green

MikaelSmith:
Rebased on CI fixes.

PR #244, opened 2017-08-17
(maint) Fix Travis CI

MikaelSmith:
Travis CI switched from Precise to Trusty, but don't have Ruby 2.1
builds for Trusty. Pin to Precise so tests against Ruby 2.1 continue to
work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 ---
 language: ruby
+dist: precise
 sudo: false
 bundler_args: --without system_tests --full-index
 before_script: cat Gemfile.lock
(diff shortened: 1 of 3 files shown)


branan:
Woah, that failure is kinda funky. I've re-kicked it to see if Travis can sort itself out

MikaelSmith:
Maybe something happened in Puppet 4.8.2?

branan:
@MikaelSmith It looks like this may be some sort of issue related to Puppet 4.8 with Facter 2.5 :(

MikaelSmith:
Tweaked, got it to pass.

branan:
:+1:

PR #245, opened 2017-08-25
Fix puppet 5 repo URL for redhat

ayush12030:


branan:
I'm closing this in favor of my work in #247, which retains a bit more flexibility around how we use the `collection` parameter, and also includes updated tests.

PR #246, opened 2017-09-15
(maint) Add pid file to Solaris upgrade

MikaelSmith:
This makes it easier to test when the upgrade is still in-progress, and
mirrors how the Windows upgrade is handled. It's either this or we grep
ps for running upgrades. Currently used in upgrade acceptance tests.

This pattern - laying down a file in /tmp/puppet_agent_upgrade.pid -
should be followed for any scripts used to perform async upgrades to
maintain testability with our current test suite.

### DIFF
--- a/templates/solaris_install.sh.epp
+++ b/templates/solaris_install.sh.epp
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+# Mark install process starting
+pid_path="$(dirname ${BASH_SOURCE[0]})/puppet_agent_upgrade.pid"
+
+if [[ -f $pid_path ]]; then
+  rm -f $pid_path
+fi
+echo $$ > $pid_path
+
 # Wait for Puppet to exit
 puppet_pid=$1
 while $(kill -0 ${puppet_pid:?}); do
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Testing this out with another change to use it in testing.

MikaelSmith:
I think failures are caused by a bump in rspec-puppet that doesn't work with Puppet 3. Going to ignore for now.

branan:
What's the actual goal of this file existing? As-written, the script will happily blows away a stale pid file, so no behavior of the module is changed.

MikaelSmith:
Almost purely for testing/debug-ability. It's either this or we grep ps for running upgrades, and this is the approach we've previously used for Windows. See https://github.com/puppetlabs/pe_acceptance_tests/pull/1593 for how it's used.

branan:
That makes sense.

Do you mind filing a ticket with that information to attach to this PR, or at least adding it to the commit message? As-written this commit feels like an easy "I have no idea why we do that, let's get rid of it" target in the future.

MikaelSmith:
Expanded the commit message a bit.

PR #247, opened 2017-09-18
(MODULES-5633) Add support for Puppet 5 on Redhat osfamily

branan:
We made a couple of changes to how Yum repo paths are constructed
in puppet 5:

1) We now put the repo name (puppet5) before the OS name/version
2) We no longer put an extra 'f' in fedora repos

This updates the module to only use the old repo path
construction if the collection is 'PC1'. Otherwise, we use
the newer construction (under the assumption that we won't change
again for puppet 6)

### DIFF
--- a/manifests/osfamily/redhat.pp
+++ b/manifests/osfamily/redhat.pp
@@ -3,8 +3,14 @@ class puppet_agent::osfamily::redhat(
 ) {
   assert_private()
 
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Travis failure is a known issue with Puppet 3.8 that really needs to be dealt with. Sometime.

PR #249, opened 2017-09-19
fix dist tag on Amazon Linux (avoids issue when ensuring package version)

wyardley:
While the EL6 packages work for Amazon Linux (https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/osfamily/redhat.pp#L15-L16), I get this error when ensuring a package version.

This PR fixes that.

```
Error: Could not update: Failed to update to version 5.1.0-1.el2017, got version 5.1.0-1.el6 instead
Error: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]/ensure: change from '5.1.0-1.el6' to '5.1.0-1.el2017' failed: Could not update: Failed to update to version 5.1.0-1.el2017, got version 5.1.0-1.el6 instead
```

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -157,12 +157,14 @@ class puppet_agent::install(
   } elsif ($::osfamily == 'RedHat') and ($package_version != 'present') {
     # Workaround PUP-5802/PUP-5025
     if ($::operatingsystem == 'Fedora') {
-      $pkg_os_suffix = 'fedoraf'
+      $dist_tag = "fedoraf${::operatingsystemmajrelease}"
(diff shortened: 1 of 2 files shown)


wyardley:
ps - The test failures on 3.8 seem to be failing in other recent builds, so I don't think they're related to this change.

wyardley:
btw: https://tickets.puppetlabs.com/browse/MODULES-5637
I suspect there may be an additional issue once I update to the version that supports the `puppet5` repo source, but will let you know.

wyardley:
@branan @MikaelSmith BTW, tests are passing after rebasing.

PR #252, opened 2017-10-24
Make management of /etc/pki directory optional

danpatdav:
Managing the /etc/pki directory inside the puppet_agent module can be problematic for organizations that manage gpg keys and settings in other modules.  Adding a manage_pki_dir parameter that defaults to true allows some flexibility to the end users.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -52,6 +52,7 @@ class puppet_agent (
   $arch                  = $::architecture,
   $collection            = 'PC1',
   $is_pe                 = $::puppet_agent::params::_is_pe,
+  $manage_pki_dir        = true,
   $manage_repo           = true,
   $package_name          = $::puppet_agent::params::package_name,
   $package_version       = $::puppet_agent::params::package_version,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -13,6 +13,10 @@
 #   The Puppet Collection to track. Defaults to 'PC1'.
 # [is_pe]
 #   Install from Puppet Enterprise repos. Enabled if communicating with a PE master.
+# [manage_pki_dir]
+#   Whether or not to manage the /etc/pki directory.  Defaults to true.
+#   Managing the /etc/pki directory inside the puppet_agent module can be problematic for 
+#   organizations that manage gpg keys and settings in other modules.
 # [manage_repo]
 #   Boolean to determine whether to configure repositories
 #   This is intended to provide the ability to disable configuring a local repo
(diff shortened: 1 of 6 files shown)


danpatdav:
@MikaelSmith Can you take a look at this PR?  Not clear on why it failed the build checks.  (this is my first PR to this project, so I may be overlooking something)

manifests/osfamily/debian.pp
```diff
@@ -84,8 +84,10 @@
     $keyname = 'GPG-KEY-puppet'
     $gpg_path = "/etc/pki/deb-gpg/${keyname}"
 
-    file { ['/etc/pki', '/etc/pki/deb-gpg']:
-      ensure => directory,
+    if getvar('::puppet_agent::manage_pki_dir') == true {
+      file { ['/etc/pki', '/etc/pki/deb-gpg']:
```
MikaelSmith: Is this something you manage elsewhere? This is something I really wish we'd fix in Puppet.

manifests/init.pp
```diff
@@ -52,6 +52,7 @@
   $arch                  = $::architecture,
   $collection            = 'PC1',
   $is_pe                 = $::puppet_agent::params::_is_pe,
+  $manage_pki_dir        = true,
```
MikaelSmith: Can you add a comment above describing this option?

MikaelSmith:
The Travis CI failure was an issue with our configuration of that service. If you rebase this PR against master, you should pickup the .travis.yml change that fixes it.

It would also be nice to see tests added to assert that disabling management works. They would be somewhere around https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/spec/classes/puppet_agent_osfamily_suse_spec.rb#L146-L150 for each of the OS you changed.

manifests/osfamily/debian.pp
danpatdav: Yes, some of our profiles manage this resource.  And in general, since this module is about managing the puppet agent, rather than gpg keys, I would avoid managing this particular directory within this module.

manifests/init.pp
danpatdav: Done.

danpatdav:
@MikaelSmith I am not experienced with rspec-puppet tests, unfortunately.  I have taken a few cracks at it, but can't seem to get the right syntax.  Can you tell what I am doing wrong?

MikaelSmith:
Sorry, I've been busy. I'll try to look at this again today or tomorrow.

MikaelSmith:
Oh, thanks for working through the tests!

MikaelSmith:
Created https://tickets.puppetlabs.com/browse/MODULES-6068 to track this issue.

PR #258, opened 2017-11-07
(maint) Switch to parallel_tests

petems:
* There’s a large amount of specs in this module
* Switching to parallel spec to speed up Travis


petems:
Yeah, and it seems to have only speed up the 3.8 tests. Killing for now.

PR #259, opened 2017-11-08
(MODULES-5953) Adds ability to set stringify_facts

petems:
* `stringify_facts` is currently not configurable 
in the module
* Lets add a class to configure it
* AND add docs on using the `puppet_conf` task to
Do it

### DIFF
--- a/manifests/prepare/stringify_facts.pp
+++ b/manifests/prepare/stringify_facts.pp
@@ -1,0 +1,21 @@
+# == Class puppet_agent::prepare::stringify_facts
+#
+# A preperation class to configure the stringify_facts setting to false
+#
+class puppet_agent::prepare::stringify_facts {
+
+  if (versioncmp($::clientversion, '4.0.0') < 0) {
+
+    ini_setting { 'puppet stringify_facts':
+      ensure  => present,
+      path    => $::puppet_config,
+      section => 'main',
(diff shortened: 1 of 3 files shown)


manifests/prepare/stringify_facts.pp
```diff
@@ -0,0 +1,21 @@
...
+    ini_setting { 'puppet stringify_facts':
+      ensure  => present,
+      path    => $::puppet_config,
+      section => 'main',
+      setting => 'stringify_facts',
+      value   => false,
+    }
+
+  } else {
+    fail('The puppet_agent::prepare::stringify_facts class should only be run on Puppet < 4')
```
MikaelSmith: This approach is slightly awkward, because it means you have to apply this class, then remove it and apply puppet_agent to upgrade. Letting it be a warning might be slightly friendlier.

petems: Yeah, I thought the stringify_facts setting would throw an error with Puppet 4, but it's just ignored, so I'll switch this to a warning 👍

petems:
@MikaelSmith Fixed to a warning 👍

MikaelSmith:
@branan care if we include this in the next release? We could hold off while we do a bugfix release, then plan to add it in a later feature release.

PR #260, opened 2017-11-09
(PE-22810) Removed _client_cert_verification

barleyj-puppet:
The private key is not world readable and apt now runs as an
 unpriveledged user. This means apt can no longer read the private
 key. We don't validate the client so we're removing the usage of the
 private key to communicate with master.

### DIFF
--- a/manifests/osfamily/debian.pp
+++ b/manifests/osfamily/debian.pp
@@ -32,19 +32,10 @@ class puppet_agent::osfamily::debian(
         "Acquire::http::proxy::${source_host} DIRECT;",
       ]
 
-      # Xenial has some sort of change that seems to have broke client cert
-      # verification in APT. While it is nice to have client cert verification,
-      # it is not strictly necessary since really all that we want to verify is
-      # that there isn't a MITM on the route to the master.
-      if ($::operatingsystem == 'Ubuntu' and $::lsbdistcodename == 'xenial') {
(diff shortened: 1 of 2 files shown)


barleyj-puppet:
I tested this by first installing puppet with:

./frankenbuilder 2017.3.0 --install --agent=debian-9-amd64 --mono --preserve-hosts=always --keyfile=~/.ssh/id_rsa-acceptance --vmpooler --workdir /tmp/frankenbuilder1 --puppet_agent=../puppetlabs-puppet_agent

I verified that the /etc/apt/apt.conf.d/90pc_repo file was as expected. I then upgraded master by using:

./frankenbuilder 2017.3.2 --upgrade --master bhtcweoy0m8ah0l.delivery.puppetlabs.net    

Then from the debian agent I rancurl --tlsv1 -O -k https://bhtcweoy0m8ah0l.delivery.puppetlabs.net:8140/packages/current/install.bash
bash install.bash

This ran without error. 

The same behavior prior to the change would result in an error that it was not able to find the packages.

manifests/osfamily/debian.pp
```diff
@@ -47,7 +43,6 @@
       } else {
```
MikaelSmith: This whole if/else block can be collapsed to the single case. I'd reduce the comment to explain why we don't use the cert/key.

PR #261, opened 2017-11-10
(MODULES-5979) Use rpm upgrade for puppet-agent upgrades

MikaelSmith:
When upgrading from a puppet-agent package, use rpm to do an inplace
upgrade rather than the previous method of uninstalling the old package
before installing the new one. This avoids an issue when upgrading
Puppet 5 installs where the upgrade run tries to use files that have
been uninstalled.

### DIFF
--- a/spec/classes/puppet_agent_osfamily_aix_spec.rb
+++ b/spec/classes/puppet_agent_osfamily_aix_spec.rb
@@ -13,7 +13,7 @@ describe 'puppet_agent' do
   end
 
   package_version = '1.2.5'
-  package_ensure = 'present'
+  package_ensure = Puppet.version < "4.0.0" ? 'present' : package_version
   let(:params) do
     {
       :package_version => package_version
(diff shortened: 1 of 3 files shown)


MikaelSmith:
This worked at least once, need to try it in a few scenarios.

MikaelSmith:
Worked for 2017.3.0 to 2017.3.2 and 2015.2.3 to 2017.3.2 upgrades.

spec/classes/puppet_agent_osfamily_aix_spec.rb
```diff
@@ -13,7 +13,7 @@
   end
 
   package_version = '1.2.5'
-  package_ensure = 'present'
+  package_ensure = Puppet.version < "4.0.0" ? 'present' : '1.2.5'
```
highb: Couldn't this '1.2.5' be set to `package_version`?

MikaelSmith: Yes, done.

highb:
👍

PR #274, opened 2018-02-26
 (PE-23563) Stop MCO and Puppet prior to MSI install 

glennsarti:
Previously it was possible to get into a strange race condition where the MSI
was able to propertly stop the MCollective and Puppet service prior to upgrade.  This
PR attempts to stop the mco and puppet services prior to the MSI being invoked, giving
the service more time to stop correctly.

This PR also adds some helpful information into the installation log for use during debugging failed installations.

### DIFF
--- a/templates/install_puppet.bat.erb
+++ b/templates/install_puppet.bat.erb
@@ -39,8 +39,9 @@ REM sets its startup type, which prevents installs from proceeding.
 REM This may fail on agents without pxp-agent, but since this is not
 REM run interactively and the next command sets ERRORLEVEL, it's OK.
 net stop pxp-agent
-REM Same for the Marionette Collective Service
+REM Same for the Marionette Collective Service and Puppet Agent service
 net stop mcollective
+net stop puppet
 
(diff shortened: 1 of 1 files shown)


glennsarti:
This PR is based on the work by @Iristyle on https://github.com/iristyle/puppetlabs-puppet_agent/tree/PE-23563-tests

MikaelSmith:
Is CI fixed with the other PR?

MikaelSmith:
This PR seems fine.

glennsarti:
Rerunning Travis CI...

glennsarti:
Travis is now green but Jenkins is Red due to some other issue.

Iristyle:
I'm a bit surprised that we were able to merge a PR using a PE ticket number, since those tickets are private / not searchable.... we have taken lengths in other projects to making sure we use public tickets for public / open source projects like this module.

That said, I see some prior art in the PR merge history... but we should probably be a bit more careful about this in the future.

glennsarti:
@Iristyle Whoops true.  I shall remember that for next time.

PR #276, opened 2018-02-27
(MODULES-6688) Use travis for CI testing, instead of Jenkins

glennsarti:
Previously the module was tested in both Travis and private Jenkins CI systems
however this is no longer required.  This commit moves the spec tests that
are tested in Jenkins and moves them to Travis;
* Only test puppet agent and ruby combinations that are installed in AIO installs
* Split out validate and lint checking to only happen once per job not for all
  matrix cells
* Now tests on Puppet latest 3.8, first 4.0, latest 4.7.x, 4.10.x and 5.x gems
* The Hiera Gem needs to be set on Puppet 4.7.1 as it has incompatibilites with
  later hiera gem versions (4.4.2)
* Brings the Travis CI config file into a similar format as that used by the PDK
  to make migration easier later.
* The Travis file needs to remain unmanaged until it supports setting the
  bundler_args setting

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -8,6 +8,40 @@
 .travis.yml:
   unmanaged: true
 
+# Requires the bundler_args to be supported.
+# .travis.yml:
+#   dist: trusty
+#   script: 'bundle exec rake $CHECK'
+#   before_install:
+#     - bundle -v
+#     - rm Gemfile.lock || true
+#     - gem update --system
+#     - gem --version
+#     - bundle -v
+#   docker_sets:
+#   includes:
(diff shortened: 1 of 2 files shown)


glennsarti:
This is in preparation for removing spec tests in Jenkins.

.sync.yml
```diff
@@ -8,6 +8,40 @@
 .travis.yml:
   unmanaged: true
 
+# Requires the bundler_args to be supported.
```
Iristyle: Why add this now if it can't be used?

glennsarti: Saving myself (or someone else) future work.  That's the only thing missing.  PDK adds that by default.

PR #278, opened 2018-02-28
(PA-1887) Fix osfamily::darwin OSX version regex

ekinanp:
The previous regex used a character class instead of a union to
match the major Mac OSX product version, which would match values
such as "10.14", "10.15". This meant that the puppet_agent module
would think these are supported platforms, and proceed further with
the installation (failing at a different point, which would produce
a harder-to-read error message). This commit fixes things so that
non-supported Mac OSX platforms correctly fail in the osfamily::darwin
class with a readable error message.

### DIFF
--- a/manifests/osfamily/darwin.pp
+++ b/manifests/osfamily/darwin.pp
@@ -3,7 +3,7 @@ class puppet_agent::osfamily::darwin(
 ) {
   assert_private()
 
-  if $::macosx_productversion_major !~ '10\.[9,10,11]' {
+  if $::macosx_productversion_major !~ /^10\.(9|10|11|12|13)/ {
     fail("${::macosx_productname} ${::maxosx_productversion_major} not supported")
   }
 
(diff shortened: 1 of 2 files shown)


ekinanp:
A quick sanity run (https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/job/experimental_test_playground_2017.3.2_2017.3.6-rc0-5-gcd13eb0_automate_run_pe_acceptance_tests/8/) indicates that this PR does not break anything, so this should be good for a merge.

MikaelSmith:
We should also update the spec test at https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/spec/classes/puppet_agent_osfamily_darwin_spec.rb#L53

ekinanp:
@MikaelSmith Oh forgot to tell you that I updated the spec test

MikaelSmith:
This should probably have a ticket, since it will fix upgrades for 10.12 and 10.13. I assume those wouldn't work before.

ekinanp:
What's interesting is those worked and still work fine -- the reason is b/c the old regex was wrong, it was using character classes. I talk about it a bit more in the commit message. I can file the ticket though.

MikaelSmith:
Oh gotcha. Probably still worth ensuring we have a release note about now enforcing macOS version checks.

ekinanp:
@MikaelSmith Ok, I filed PA-1887 for this work and it seems that the CI checks have passed.

PR #280, opened 2018-03-05
(MODULES-6708) Pin inifile module to <= 2.1.0

Iristyle:
- To support older platforms with older Ruby versions that are being
   upgraded from, pin back to an older version of inifile that doesn't
   introduce incompatibilities with older Ruby runtime versions.

   In version 2.2.0 of the inifile module, Ruby 2.0+ constructs for
   named arguments were introduced which will break this modules
   behavior.

   Guard against that so that tests may pass.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -4,7 +4,9 @@ fixtures:
       repo: "puppetlabs/stdlib"
       ref: "4.12.0"
     transition: "puppetlabs/transition"
-    inifile: "puppetlabs/inifile"
+    inifile: 
+      repo: "puppetlabs/inifile"
+      ref: "2.1.0"
     apt:
       repo: "puppetlabs/apt"
       ref: "2.3.0"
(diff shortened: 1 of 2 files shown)


Iristyle:
To add:

* Previously Jenkins pipelines passed against 2.0.0 of inifile
* 2.1.1 introduces at least one change that is incompatible with JRuby 1.7 - https://github.com/puppetlabs/puppetlabs-inifile/compare/2.1.0...2.1.1#diff-6b9f1e8fde3f18501b209c68609a51b0R5
* Based on diff from 2.0.0, 2.1.0 appears to be latest version that doesn't introduce Ruby language compatibility issues - https://github.com/puppetlabs/puppetlabs-inifile/compare/2.0.0...2.1.0

Iristyle:
@glennsarti has a similar branch going at https://github.com/glennsarti/puppetlabs-puppet_agent/tree/MODULES-6708-fix-smoke-tests

glennsarti:
@Iristyle  Can I close this?

Iristyle:
@glennsarti I don't think so... `inifile` hasn't cut another release yet, nor has the work landed to fix its language compatibility...

/cc @eputnam

glennsarti:
I should've explained myself better.  This was cherrypicked into the 1.x branch and merged in https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/281 .  Once the release for 1.6.0 is done this will be merged back into master.

I guess what I'm say is this PR is superceded by #281

PR #285, opened 2018-04-02
(MODULES-6708) fix tests for windows agent to agent upgrades

speedofdark:
* remove tests around configuration migration as they are not managed by the upgrade
* changed base version from 3.x to 1.7.0 and associated changes
* specified upgrade target as 1.10.0

Three Windows tests are currently red. Two appear to be legitimate failues related to
idempotency. The third is related to service startup.

Tests have only been updated to make things green, with no other refactoring applied
as yet.

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -2,13 +2,13 @@ require 'spec_helper_acceptance'
 
 describe 'puppet_agent class' do
 
-  context 'default parameters' do
+  context 'default parameters in apply' do
     before(:all) { setup_puppet_on default }
     after (:all) { teardown_puppet_on default }
 
     it 'should work idempotently with no errors' do
       pp = <<-EOS
-      class { 'puppet_agent': }
+      class { 'puppet_agent': package_version => '1.10.0' }
       EOS
 
       # Run it twice and test for idempotency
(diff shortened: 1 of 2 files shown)


spec/acceptance/class_spec.rb
```diff
@@ -23,9 +23,17 @@ class { 'puppet_agent': }
...
     end
 
-    describe service('puppet') do
-      it { is_expected.to be_enabled }
-      it { is_expected.to be_running }
+    if default['platform'] =~ /windows/i
+      # puppet service is not started automagically on windows
+      describe service('puppet') do
+        it { is_expected.to be_enabled }
+        it { is_expected.to_not be_running }
```
glennsarti: This appears very odd....Not sure we should be expecting this.

speedofdark:
We reverted the change that Glenn commented on above; this results in a total of 3 red tests for now. This will be addressed separately.

speedofdark:
1.7.0 was the version shipped in the 2016.4.0 (the initial LTS release), and the module doesn't work well for 5 yet on every OS we support so we use 1.10.0 as the last of the 4 series. We could pick 1.10.12 instead. It doesn't really matter (shrug).

spec/spec_helper_acceptance.rb
```diff
@@ -135,24 +134,19 @@ def setup_puppet_on(host, opts = {})
 
   puts "Setup foss puppet on #{host}"
   configure_defaults_on host, 'foss'
-  install_puppet_on host, :version => ENV['PUPPET_CLIENT_VERSION'] || '3.8.6'
+  install_puppet_agent_on host, :version => ENV['PUPPET_CLIENT_VERSION'] || '1.7.0'
```
glennsarti: 👍

PR #286, opened 2018-04-03
(MODULES-5230) Do not manage PA version on PE infra nodes

npwalker:
Prior to this commit, we would manage the puppet-agent version
on any node that was classified with the puppet_agent class.

After this commit, we no check to see if the node is a PE infra
node and if so print a message saying we're not going to do anything
because PE infra nodes have a specific version of puppet-agent
on them for a reason.

You would not, for instance, want your compile master to get its
agent downgraded right after you completed an upgrade of your PE
installation.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -78,6 +78,8 @@ class puppet_agent (
     info('puppet_agent performs no actions if a package_version is not specified on Puppet 4')
   } elsif $package_version == undef and $is_pe {
     info("puppet_agent performs no actions if the master's agent version cannot be determed on PE 3.x")
+  } elsif $facts['pe_server_version'] != undef {
+    info('puppet_agent performs no actions on PE infrastructure nodes to prevent a mismatch between agent and PE components')
   } else {
     if $package_version != undef and $package_version !~ /^\d+\.\d+\.\d+([.-]?\d*|\.\d+\.g[0-9a-f]+)$/ {
       fail("invalid version ${package_version} requested")
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -76,6 +76,14 @@ describe 'puppet_agent' do
             it { is_expected.to contain_class('puppet_agent').with_package_version(nil) }
           end
         end
+
+        context 'On a PE infrastructure node puppet_agent does nothing' do
+          before(:each) do
+            facts['pe_server_version'] = '2016.2.2'
+          end
+          it { is_expected.to_not contain_class('puppet_agent::prepare') }
+          it { is_expected.to_not contain_class('puppet_agent::install') }
+        end
       end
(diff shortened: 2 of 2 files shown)


jpartlow:
:+1: You've made the linter unhappy though.
How about a simple spec verifying that nothing is included if the fact is set?
nit: Og find typo in commit message 'we no check' :)

npwalker:
@jpartlow changes made.

MikaelSmith:
That makes sense.

manifests/init.pp
```diff
@@ -76,6 +76,8 @@
     info('puppet_agent performs no actions if a package_version is not specified on Puppet 4')
   } elsif $package_version == undef and $is_pe {
     info("puppet_agent performs no actions if the master's agent version cannot be determed on PE 3.x")
+  } elsif $facts['pe_server_version'] != undef {
+    info('puppet_agent performs no actions on PE infrastructure nodes to prevent a mismatch between agent and PE components')
```
npwalker: @MikaelSmith thoughts on why these are logging at info?  I was thinking of upgrading them to [warning](https://docs.puppet.com/puppet/3.8/function.html#warning).

MikaelSmith: I forget if that's something we got complaints about. The 1st one was info so that you wouldn't get warnings immediately after upgrading from Puppet 3. The 2nd should probably be a warning.

branan: I don't think it should be a warning. We should only use warnings for things that a user is doing incorrectly that we believe are actionable. In this case, it's a totally reasonable behavior to classify "all nodes" with this module, and the module knows how to handle that case correctly (now). Warning here and forcing users to create a new "everything-but-pe-infrastructure" group in order to silence that warning is just annoying busywork.

npwalker: I think that makes sense the message I'm adding isn't actionable it's informational.

branan:
Should this be targeted at the 1.x branch now?

branan:
Also, I /generally/ have a preference that tickets against our FOSS repos have a publicly-visible ticket so that non-puppet contributors can see the tickets behind a `git blame`. It looks like the comments which mention customers in that ticket are already marked for developers only, so could we migrate it to MODULES and update the commit message before we land this?

npwalker:
@branan There's an existing MODULES ticket https://tickets.puppetlabs.com/browse/MODULES-5230 I can switch over to using.  

Do you want me to retarget at the 1.x branch?  idk what the difference between branches is.

branan:
@npwalker master is the branch that's going to drop support for Puppet 3. Since that will likely involve lots of cleanups and refactoring, I'm unsure of when we'll release from it. a Relaese from 1.x is pretty straightforward for fixes to the existing code.

glennsarti:
This change has broken the Acceptance tests since April :-(

PR #289, opened 2018-05-04
(MODULES-6915) Remove check for PE when calculating whether to upgrade

smcclellan:
On FOSS Windows specifically, the agent would continue to upgrade on
every puppet run. This removes a code path that would behave differently
for PE vs. FOSS, and actually perform a version check.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -90,9 +90,10 @@ class puppet_agent (
       $_expected_package_version = $package_version
     }
 
-    $aio_upgrade_required = ($is_pe == false and $_expected_package_version != undef) or
-      (getvar('::aio_agent_version') != undef and $_expected_package_version != undef and
-        versioncmp("${::aio_agent_version}", "${_expected_package_version}") < 0)
+    $aio_upgrade_required = (getvar('::aio_agent_version') == undef) or
+      (getvar('::aio_agent_version') != undef and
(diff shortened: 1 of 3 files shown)


smcclellan:
Updating spec tests now, but wanted review on whether this is the right approach in the meantime.

geoffnichols:
Wondering if spec tests will fail until https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/285 is merged.

manifests/init.pp
```diff
@@ -90,9 +90,11 @@
...
     }
 
-    $aio_upgrade_required = ($is_pe == false and $_expected_package_version != undef) or
-      (getvar('::aio_agent_version') != undef and $_expected_package_version != undef and
-        versioncmp("${::aio_agent_version}", "${_expected_package_version}") < 0)
+    $aio_upgrade_required = (getvar('::aio_agent_version') == undef) or
+      (getvar('::aio_agent_version') != undef and
+      $_expected_package_version != undef and
+      versioncmp("${::aio_agent_version}", "${_expected_package_version}") < 0)
+    # $aio_upgrade_required = true
```
speedofdark: Really needed? Thanks.

geoffnichols:
@smcclellan just to confirm, is this good to merge now?

PR #290, opened 2018-05-15
(MODULES-7167) Prepare for 1.6.1 release

glennsarti:
This commit prepares the module for a version 1.6.1 release.

Note that this is targetted directly on the `1.x` branch not `release` as the 1.x branch is end-of-life


MikaelSmith:
Do we expect 3.8 CI to pass?

glennsarti:
The first 3.8x failures happened during a release prep so it looks like an external dependency issue which I haven't bothered to track down.  So it's expected right now

PR #291, opened 2018-05-23
Enable support for inifile 2.2.0

robbyt:
inifile 2.2 is backwards compatible with 2.1, and is working fine in my environments


fduranti:
It would be nice to have the possibility to update inifile in the next version... 1.5.0 has inifile <3.0 while 1.6.0 has inifile < 2.2.0

GabrielNagy:
Closing this since we now support inifile < 5.0.0

PR #292, opened 2018-06-05
[MODULES-4195] Install AIX 6.1 RPMs on all AIX versions for puppet6

caseywilliams:
For puppet 6, we intend to provide AIX 6.1 packages for all supported
AIX platforms; Prior releases should keep the same behavior for AIX
platforms.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -130,9 +130,15 @@ class puppet_agent (
       $_package_file_name = "${puppet_agent::package_name}-${package_version}-1.osx${$::macosx_productversion_major}.dmg"
     } elsif $::operatingsystem == 'AIX' {
       $_aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
-      $aix_ver_number = $_aix_ver_number ? {
-        /^7\.2$/ => '7.1',
-        default  => $_aix_ver_number,
+      if $_aix_ver_number {
+        if $collection =~ /(PC1|puppet5)/ {
+          $aix_ver_number = $_aix_ver_number ? {
+            /^7\.2$/ => '7.1',
+            default  => $_aix_ver_number,
+          }
+        } else {
+          $aix_ver_number = '6.1'
+        }
       }
       $_package_file_name = "${puppet_agent::package_name}-${package_version}-1.aix${aix_ver_number}.ppc.rpm"
     } elsif $::osfamily == 'windows' {
--- a/manifests/prepare/package.pp
+++ b/manifests/prepare/package.pp
@@ -24,10 +24,16 @@ class puppet_agent::prepare::package(
       }
       $source = "puppet:///pe_packages/${pe_server_version}/${tag}/${package_file_name}"
     } elsif $::operatingsystem == 'AIX' {
-      $tag = $::platform_tag ? {
-        'aix-7.2-power' => 'aix-7.1-power',
-        default         => $::platform_tag,
+      if $::puppet_agent::collection =~ /(PC1|puppet5)/ {
+        $tag = $::platform_tag ? {
+          'aix-7.2-power' => 'aix-7.1-power',
+          default         => $::platform_tag,
+        }
+      } else {
+        # From puppet6 onward, AIX 6.1 packages are used for all AIX platforms
+        $tag = 'aix-6.1-power'
       }
+
       $source = "puppet:///pe_packages/${pe_server_version}/${tag}/${package_file_name}"
     } else {
       $source = "puppet:///pe_packages/${pe_server_version}/${::platform_tag}/${package_file_name}"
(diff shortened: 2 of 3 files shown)


manifests/prepare/package.pp
```diff
@@ -24,10 +24,15 @@
       }
       $source = "puppet:///pe_packages/${pe_server_version}/${tag}/${package_file_name}"
     } elsif $::operatingsystem == 'AIX' {
-      $tag = $::platform_tag ? {
-        'aix-7.2-power' => 'aix-7.1-power',
-        default         => $::platform_tag,
+      if $package_file_name =~ /^puppet-agent-6.*/ {
```
caseywilliams: It seems like there's a better way to check this here -- is there a way I could check the collection as opposed to the agent version?

spec/classes/puppet_agent_osfamily_aix_spec.rb
```diff
@@ -1,108 +1,132 @@
...
+
+    it {
+      is_expected.to contain_package('puppet-agent').with({
+        'source'    => "/opt/puppetlabs/packages/#{rpmname}",
+        'ensure'    => Puppet.version < '4.0.0' ? 'present' : params[:package_version],
+        'provider'  => 'rpm',
+      })
+    }
 
+    if Puppet.version < "4.0.0"
```
caseywilliams: Could we remove these < 4.0.0 tests now?

manifests/prepare/package.pp
caseywilliams: come to think of it, I don't expect this will even work until the reported agent version is updated

McdonaldSeanp: I _think_ you can use `$::puppet_agent::collection`

smcclellan:
Good to merge once tests complete.

ekinanp:
You should be able to have https://github.com/puppetlabs/pe_acceptance_tests/blob/2018.2.x/lib/puppet_enterprise_acceptance/setup_helpers.rb#L355-L356 point to your fork and branch of the puppetlabs-puppet_agent module.

@cthorn42 would know more about running those tests, including what branch's tests you should use. I believe your situation is testing if an agent upgrade to puppet6 (Johnson?) successfully installs the AIX 6.1 package on all the AIX platforms right?

cthorn42:
If you're looking for the job that is testing non-standard agents upgrading to PE Johnson, this is the job you want: https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/pe-integration/view/pe-2018.2.x/job/enterprise_pe-acceptance-tests_integration-system_pe_non-standard-full-agent-upgrade_non-standard_2018.2.x/
That is testing upgrades from PE 2018.1.0, so should be 5.5.1 agents upgrading to 6.0 agents, and tests aix 61, 71, and 72.

caseywilliams:
I've updated this to rely on the puppet-agent package version number instead of the puppet collection, since internal builds do not consistently ship to paths that reflect the actual collection -- this will require that the version number in puppet-agent#master is updated to 5.99.z (as was the strategy for pre-release puppet5), so I'm blocking this for now

manifests/init.pp
```diff
@@ -130,9 +130,20 @@
...
+            $aix_ver_number = $_aix_ver_number ? {
+              /^7\.2$/ => '7.1',
+              default  => $_aix_ver_number,
+            }
+          }
+          else {
+            $aix_ver_number = '6.1'
+          }
+        } else {
+          $aix_ver_number = $_aix_ver_number
```
caseywilliams: This code should never be reached, but thought I'd allow for it just in case -- if anyone has ideas around cleaning this up, they're appreciated

ekinanp: Might be cleaner to create a custom function that's named something like
```
is_puppet6?(agent_version)
```

Then you can rewrite this as:
```
$actual_aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
$aix_package_ver_number = is_puppet6?(agent_version) ? {
  true => '6.1'
  false => $actual_aix_ver_number ? {
    /7.2/ => 7.1
    default => $actual_aix_ver_number
  }
}
```

and probably do something similar for the other case

caseywilliams:
I've updated this with a little cleanup, and puppet-agent has been tagged as 5.99 in master, so I'm unblocking this now

caseywilliams:
(The Travis failures for 3.8/future parser look similar to those in https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/290 to me, so I'm thinking they may not be related)

caseywilliams:
Instead of using a version-based workaround, we've decided to just update the collections used for internal builds of puppet5 and puppet6 so that they no longer use PC1 everywhere. This PR is blocked until that switchover happens.

caseywilliams:
Repo paths have been updated to use `puppet5` instead of `PC1` for agent 5.5.x [here](https://github.com/puppetlabs/puppet-agent/pull/1450) -- unblocking this now.

PR #293, opened 2018-06-05
Added facts_upload task

dylanratcliffe:
Adds the ability to run `puppet facts upload` via a task.

Tested on RHEL 7 but as it uses the face it should work on all OSes, would appreciate help testing on others.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -15,6 +15,7 @@
     * [Public classes](#public-classes)
     * [Private classes](#private-classes)
     * [Parameters](#parameters)
+    * [Tasks](#tasks)
 6. [Limitations - OS compatibility, etc.](#limitations)
     * [Known issues](#known-issues)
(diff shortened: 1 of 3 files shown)


glennsarti:
nit: Was this feature added in the 1.10 series too? or only 5.5

dylanratcliffe:
I think it's only 5.5. Certainly no mention in the ticket of any other versions but you will probably have a better understanding than me on how that works

adreyer:
I'm not sure the puppet agent module is the correct place for tasks that "run" the agent. It's primary job now is managing the agent installation.

The puppetlabs-facts module may be a better place for this but it's primary concern now is collecting facts for plans.

I also don't think we should be merging tasks in to a high profile module without someone committed to writing acceptance tests for them

dylanratcliffe:
Will do in a separate module.

PR #302, opened 2018-07-24
(BOLT-701) Add task to install agent package on linux

adreyer:


tasks/shell.sh
```diff
@@ -0,0 +1,555 @@
...
+  machine="ppc"
+fi
+
+if test "x$platform" = "x"; then
+  critical "Unable to determine platform version!"
+  report_bug
+  exit 1
+fi
+
+# TODO: we shold either get a specific version or latest right?
```
adreyer: Can I remove all of this logic and force the user specify a specific puppet agent package version?

tasks/shell.sh
```diff
@@ -0,0 +1,555 @@
...
+if test "x$version" = "x"; then
+  version="latest";
+  info "Version parameter not defined, assuming latest";
+else
+  info "Version parameter defined: $version";
+  info "Matching Puppet version to puppet-agent package version (See http://docs.puppetlabs.com/puppet/latest/reference/about_agent.html for more details)"
+  case "$version" in
+    5.0.*)
```
MikaelSmith: This section is weird, why? I would expect them to specify a specific version of the puppet-agent package, not a version of Puppet that we map to a package.

tasks/shell.sh
MikaelSmith: We specifically wanted that use case because it seemed like the most common request would be: I want a Puppet 5 agent package, give me the latest. The logic was specifically
- if I don't specify a version and nothing's installed, install latest
- if I don't specify a version and something's installed, leave it
- if I specify 'latest', install or update to the latest version
- if I specify a version, install or update to that version

tasks/shell.sh
```diff
@@ -0,0 +1,555 @@
...
+          "12.04") deb_codename="precise";;
+          "12.10") deb_codename="quantal";;
+          "13.04") deb_codename="raring";;
+          "13.10") deb_codename="saucy";;
+          "14.04") deb_codename="trusty";;
+          "15.04") deb_codename="vivid";;
+          "15.10") deb_codename="wily";;
+          "16.04") deb_codename="xenial";;
+          "16.10") deb_codename="yakkety";;
+          "17.04") deb_codename="zesty";;
```
donoghuc: I have been using this for testing `apply` on ubuntu 18.04 by adding `"18.04") deb_codename="bionic";;`

tasks/shell.sh
adreyer: this is the logic I was hoping to remove

tasks/shell.sh
adreyer: I'll go ahead and add that

tasks/install.json
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "Install the puppet 5 agent package",
+  "parameters": {
+    "version": {
+      "description": "The version of puppet to install",
+      "type": "Optional[String]"
+    }
+  },
+  "implementations": [
+    {"name": "shell.sh", "requirements": ["shell"]},
```
nicklewis: This trailing comma breaks the metadata

tasks/shell.sh
```diff
@@ -0,0 +1,539 @@
...
+#      usage: $0 [-v version] [-f filename | -d download_dir]"
+#      exit 1;;
+#  esac
+#done
+shift `expr $OPTIND - 1`
+
+machine=`uname -m`
+os=`uname -s`
+
+# CODEREVIEW: What are we doing with this?
```
MikaelSmith: Apparently unused?

tasks/shell.sh
```diff
@@ -0,0 +1,539 @@
...
+  fi
+# Apple OS X
+elif test -f "/usr/bin/sw_vers"; then
+  platform="mac_os_x"
+  # Matching the tab-space with sed is error-prone
+  platform_version=`sw_vers | awk '/^ProductVersion:/ { print $2 }'`
+
+  major_version=`echo $platform_version | cut -d. -f1,2`
+  case $major_version in
+    "10.6") platform_version="10.6" ;;
```
MikaelSmith: These are ancient, I don't think we even support them anymore. But I guess macOS work is a separate ticket.

MikaelSmith:
I think we should add something to the README, similar to what https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/293 has.

MikaelSmith:
We can add the README as we get closer to completion.

PR #303, opened 2018-07-26
(MODULES-7535) Prep for 1.6.2 release

branan:


branan:
Red travis cell is due to a transitive dependency that needs to be pinned in our gemfile for old ruby support

PR #306, opened 2018-08-01
(BOLT-641) Add version task

MikaelSmith:
Adds a task to retrieve the installed version of the puppet-agent
package. Returns empty results if the package is not present.


tasks/version_powershell.ps1
```diff
@@ -0,0 +1,8 @@
+if (Test-Path 'HKLM:\SOFTWARE\Puppet Labs\Puppet') {
+  $reg = $(Get-ItemProperty -Path 'HKLM:\SOFTWARE\Puppet Labs\Puppet')
+  if (Test-Path $reg.RememberedInstallDir64) {
+    type $($reg.RememberedInstallDir64+"VERSION")
+  } elseif (Test-Path $reg.RememberedInstallDir) {
+    type $($reg.RememberedInstallDir+"VERSION")
+  }
+}
```
MikaelSmith: Should we also try `puppet --version` in case Puppet is present not from our packages? For Bolt's use-case, apply wouldn't work without `/opt/puppetlabs/puppet/bin/ruby` or `C:/Program Files/Puppet Labs/Puppet/sys/ruby/bin/ruby`.

MikaelSmith:
In a plan, that would be `$result.message`.

nicklewis:
Does this consistently return puppet-agent version vs. puppet version, between the various implementations? Related, is there a benefit to having the .rb implementation if the sh and ps1 versions do the same thing but faster?

MikaelSmith:
Oops, the Ruby version won't. And we've historically included the Ruby version for Bolt over Orchestrator, where we don't have features enabled. But it probably doesn't actually help (you'd then have to choose shell or powershell version), so I can remove it.

tasks/version_shell.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+if test -f /opt/puppetlabs/puppet/VERSION; then
+  cat /opt/puppetlabs/puppet/VERSION
```
adreyer: If we're writing a task in a supported module I think it should return JSON

tasks/version_powershell.ps1
adreyer: I think we can include some info about how we found it as well.

tasks/version_shell.sh
MikaelSmith: Do you think it needs to be extensible, or some other reason? And what info specifically about how we found it?

MikaelSmith: Made it `{"version": "<ver>", "source": "/opt/puppetlabs/puppet/VERSION"}`.

MikaelSmith:
Added tests.

tasks/version_powershell.ps1
```diff
@@ -0,0 +1,14 @@
...
+    $loc = $reg.RememberedInstallDir64+'VERSION'
+  } elseif (Test-Path $reg.RememberedInstallDir) {
+    $loc = $reg.RememberedInstallDir+'VERSION'
+  }
+}
+
+if ($loc -ne $null) {
+  Write-Output "{`"version`":`"$(type $loc)`",`"source`":`"$($loc.replace('\', '/'))`"}"
+} else {
+  Write-Output '{"version":null}'
```
adreyer: I think this should be an _error or have source: null

MikaelSmith: If one of the use cases is detecting whether Puppet is installed, making it an error could be a little harder to handle. Not sure if we'd want to try install on any error, or check the error type and only try if it was absent.

nicklewis: This doesn't seem like an error to me. The task correctly determined that there is no puppet agent version.

If there were any *other* sort of error, we should definitely not try to install, because something *really really* weird must be going on.

PR #307, opened 2018-08-01
(BOLT-702) Install puppet agent on osx

donoghuc:
Support puppet-agent install on osx 11,12,13. Logic is based on https://github.com/puppetlabs/puppetlabs-pe_repo/blob/hoyt/templates/osx.bash.erb#L5

PR #308, opened 2018-08-02
(BOLT-703) Install puppet agent on windows

donoghuc:
Add powershell task implementation to install puppet agent on windows platforms. Logic based on https://github.com/puppetlabs/puppetlabs-pe_repo/blob/hoyt/templates/install.ps1.erb

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -27,38 +27,51 @@ describe 'install task' do
     end
 
     nodes = hosts.map do |host|
-      ssh = host[:ssh]
-
-      if ssh
-        config = { transport: 'ssh' , ssh: {} }
-        [:password, :user, :port].each { |k| config[:ssh][k] = ssh[k] if ssh[k] }
-
-        # This is required to find the key that may be loaded from config
-        # The vagrant provisioner passes the vagrant ssh config instead of
-        # specific options.
-        key = nil
-        keys = host.connection.instance_variable_get(:@ssh).options[:keys]
-        key = keys.first if keys
-        config[:ssh][:"private-key"] = key if key
-
-        # If the hypevisor stores IP as part of its core config then hostname
-        # is unlikely to resolve in beakers world, but there may be many hosts
-        # on the same IP. Use a query argument to generate a unique hostname.
-        # BOLT-510
-        hostname = host.host_hash[:ip] ? "#{host.host_hash[:ip]}?n=#{host.hostname}" : host.hostname
-        node_name = "ssh://#{hostname}"
-
+      if host[:platform] =~ /windows/
+        config = {
+          transport: 'winrm',
+          winrm: {
+            user: host[:user],
+            password: ENV['BEAKER_password'],
+            ssl: false
+          }
+        }
         node = {
-          name: node_name,
+          name: host.hostname,
           config: config
         }
       else
-        raise 'non-ssh targets not yet implemented'
-      end
+        ssh = host[:ssh]
+        if ssh
+          config = { transport: 'ssh' , ssh: {} }
+          [:password, :user, :port].each { |k| config[:ssh][k] = ssh[k] if ssh[k] }
 
+          # This is required to find the key that may be loaded from config
+          # The vagrant provisioner passes the vagrant ssh config instead of
+          # specific options.
+          key = nil
+          keys = host.connection.instance_variable_get(:@ssh).options[:keys]
+          key = keys.first if keys
+          config[:ssh][:"private-key"] = key if key
+
+          # If the hypevisor stores IP as part of its core config then hostname
+          # is unlikely to resolve in beakers world, but there may be many hosts
+          # on the same IP. Use a query argument to generate a unique hostname.
+          # BOLT-510
+          hostname = host.host_hash[:ip] ? "#{host.host_hash[:ip]}?n=#{host.hostname}" : host.hostname
+          node_name = "ssh://#{hostname}"
+
+          node = {
+            name: node_name,
+            config: config
+          }
+        else
+          raise 'non-ssh targets not yet implemented'
+        end
+      end
       # Make alias groups for each role
       host[:roles].each do |role|
-        add_node(node_name, role, groups)
+        add_node(node[:name], role, groups)
       end
       node
     end
@@ -174,11 +187,6 @@ describe 'install task' do
 
   it 'runs and installs the agent' do
     results = run_task_on('target', "puppet_agent::install")
-    results.each do |res|
-      expect(res["status"]).to eq("success")
-    end
-    # Some packages don't add this path
-    results = run_command_on('target', '/opt/puppetlabs/bin/puppet --version')
     results.each do |res|
       expect(res["status"]).to eq("success")
     end
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -1,0 +1,63 @@
+[CmdletBinding()]
+Param(
+	[String]$version
+)
+# If an error is encountered, the script will stop instead of the default of "Continue"
+$ErrorActionPreference = "Stop"
+
+if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -match '^32') {
+    $arch = "x86"
+} else {
+    $arch = "x64"
+}
+
+if ($version) {
+    $msi_source = "http://downloads.puppetlabs.com/windows/puppet5/puppet-agent-${version}-${arch}.msi"
+}
+else {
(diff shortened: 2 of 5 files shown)


donoghuc:
Similar test setup as https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/307
Note the addition of the environment variable `WINRM_PWD` required for winrm transport.
Example test invocation:
```
puppetlabs-puppet_agent/task_spec$ WINRM_PWD=****** GEM_BOLT=true BEAKER_debug=true BEAKER_set=windows10ent-32 bundle exec rake task_acceptance
```

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -176,10 +180,10 @@ def bolt_on(target, type, action, object,
       expect(res["status"]).to eq("success")
     end
     # Some packages don't add this path
-    results = run_command_on('target', '/opt/puppetlabs/bin/puppet --version')
-    results.each do |res|
-      expect(res["status"]).to eq("success")
-    end
+    # results = run_command_on('target', '/opt/puppetlabs/bin/puppet --version')
```
donoghuc: This test will fail on windows. Can we rely on the version task for verification of install?

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -52,6 +52,10 @@ def add_node(node, group_name, groups)
       else
         # This is very unlikely to work
         node_name, node = host.hostname
+        node = {
```
donoghuc: Will make this more dynamic as we develop the test further (example the hard coded "Administrator" user...). For now focusing on adding logic to BOLT-702.

Iristyle:
It's unclear to me why this lands in the puppet-agent module and not the bootstrap module?

https://github.com/puppetlabs/puppetlabs-bootstrap/blob/master/tasks/windows.ps1


Do we have a plan to merge the modules together at some point?

tasks/install_powershell.ps1
```diff
@@ -0,0 +1,50 @@
...
+Param(
+	[String]$version
+)
+# If an error is encountered, the script will stop instead of the default of "Continue"
+$ErrorActionPreference = "Stop"
+
+$arch = if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -match '^32') {"x86"} else {"x64"}
+
+if ($version) {
+    $msi_source = "http://downloads.puppetlabs.com/windows/puppet5/puppet-agent-${version}-${arch}.msi"
```
glennsarti: Hardcoded to `puppet5`. This should be parameterised

tasks/install_powershell.ps1
```diff
@@ -0,0 +1,50 @@
+[CmdletBinding()]
+Param(
+	[String]$version
+)
+# If an error is encountered, the script will stop instead of the default of "Continue"
+$ErrorActionPreference = "Stop"
+
+$arch = if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -match '^32') {"x86"} else {"x64"}
```
glennsarti: nit: This is a little ruby-like.  Normally Powershell would follow
```
If 'something' {
  $arch = '1'
} else {
  $arch = '2'
}
```

tasks/install_powershell.ps1
```diff
@@ -0,0 +1,50 @@
...
+    if($_.Exception.Response.StatusCode -eq [system.net.httpstatuscode]::NotFound) {
+        Throw "Failed to download the Puppet Agent installer: $msi_source"
+    }
+    # Throw all other WebExceptions
+    Throw $_
+  }
+}
+
+function InstallPuppet {
+  $msiexec_args = "/qn /i $msi_dest"
```
glennsarti: Additional Install arguments should parameterised.  e.g. Install logs for debug. 

* Also you're missing `RESTART=` otherwise you may get an unexpected reboot during a bolt run.

tasks/install_powershell.ps1
```diff
@@ -0,0 +1,50 @@
...
+  $msiexec_proc.WaitForExit()
+  if (@(0, 1641, 3010) -NotContains $msiexec_proc.ExitCode) {
+    Throw "Something went wrong with the installation on $env:COMPUTERNAME. Exit code: " + $msiexec_proc.ExitCode
+  }
+}
+
+DownloadPuppet
+InstallPuppet
+
+Write-Output "Puppet Agent installed on $env:COMPUTERNAME"
```
glennsarti: Missing newline at end of file

tasks/install_powershell.ps1
```diff
@@ -0,0 +1,50 @@
...
+  Write-Output "Installing the Puppet Agent on $env:COMPUTERNAME..."
+  $msiexec_proc = [System.Diagnostics.Process]::Start('msiexec', $msiexec_args)
+  $msiexec_proc.WaitForExit()
+  if (@(0, 1641, 3010) -NotContains $msiexec_proc.ExitCode) {
+    Throw "Something went wrong with the installation on $env:COMPUTERNAME. Exit code: " + $msiexec_proc.ExitCode
+  }
+}
+
+DownloadPuppet
+InstallPuppet
```
glennsarti: Should this script cleanup the downloaded file afterwards?

tasks/install_powershell.ps1
```diff
@@ -0,0 +1,50 @@
...
+
+$msi_dest = Join-Path ([System.IO.Path]::GetTempPath()) "puppet-agent-$arch.msi"
+
+function DownloadPuppet {
+  Write-Verbose "Downloading the Puppet Agent for Puppet Enterprise on $env:COMPUTERNAME..."
+
+  $webclient = New-Object system.net.webclient
+
+  try {
+    $webclient.DownloadFile($msi_source,$msi_dest)
```
glennsarti: Have you tested to see what happens if `$msi_dest ` already exists?

task_spec/spec/acceptance/init_spec.rb
adreyer: I think just wrapping it in conditional that checks if the host is windows. We don't want to test ssh against windows for now and beaker doesn't handle winrm connections(for the most part).

Is there an existing PASSWORD environment variable that beaker uses we can reuse here?

tasks/install_powershell.ps1
MikaelSmith: Yes, eventually it should be.

tasks/install_powershell.ps1
MikaelSmith: Where do the MSI logs go if you don't specify anything?

tasks/install_powershell.ps1
MikaelSmith: I think we could add that as a `collection` parameter that takes `Enum[PC1, puppet5, puppet6]`. This will increase our testing matrix a bit.

tasks/install_powershell.ps1
glennsarti: They don't log to a file.  You just the usual minimal logs in the eventlog.

tasks/install_powershell.ps1
glennsarti: If we had pester tests, we could easily test at the unit level and avoid acceptance tests.

adreyer:
@Iristyle the bootstrap module only works with PE, manages agent configuration and service, and is not super discoverable.
Long term I expect the puppet_agent module to support installing from a PE master and a plan with an apply block to handle configuring the agent.

tasks/install_powershell.ps1
donoghuc: Added a `Cleanup` function to remove MSI and Install Log upon successful installation.

tasks/install_powershell.ps1
donoghuc: It appears that `$msi_dest` will be over-written. I think this is the desired behavior.

tasks/install_powershell.ps1
donoghuc: Current implementation will log the install to temp location. If the installation fails, the file (and MSI) are persisted. Upon successful installation both the log and the MSI are deleted.

tasks/install_powershell.ps1
donoghuc: @MikaelSmith Would you like the `collection` metaparameter added in this ticket? `puppet5` is currently hard coded for the OSX support in the `install_shell.sh` implementation.

MikaelSmith: Let's ticket as an extra subtask. Also, it's just a parameter, not a metaparameter.

MikaelSmith:
Needs a rebase.

MikaelSmith:
Ran through osx, windows 32, windows 64. 👍

PR #309, opened 2018-08-03
(MODULES-7580) Fix PowerShell task for puppet version on 32bit

glennsarti:
Previously the powershell based version task was failing on 32bit windows
operating systems due to passing in a null path to Test-Path.  This commit
modifies the detection to be far more tolerant of errors and instead detects if
the registry entries are null instead of using Test-Path.

This commit also cleans up the code as per PS Script Analyzer rules;
* Remove use of aliases
* Prefer `$null -eq ...` instead of `... -eq $null


glennsarti:
Tested manually on Windows 10 32bit, 64bit and 2008R2 64bit.

tasks/version_powershell.ps1
```diff
@@ -1,14 +1,16 @@
...
-  } elseif (Test-Path $reg.RememberedInstallDir) {
+  } elseif ($null -ne $reg.RememberedInstallDir) {
     $loc = $reg.RememberedInstallDir+'VERSION'
   }
 }
 
-if ($loc -ne $null) {
-  Write-Output "{`"version`":`"$(type $loc)`",`"source`":`"$($loc.replace('\', '/'))`"}"
+if ($null -ne $loc) {
+  Write-Output "{`"version`":`"$(Get-Content -Path $loc -ErrorAction SilentlyContinue)`",`"source`":`"$($loc.replace('\', '/'))`"}"
```
MikaelSmith: `-ErrorAction SilentlyContinue` will result in an empty version if reading the file fails for some reason (permissions?). I'm not sure that's what we want... if the file is present but we can't read it, we may want to fail the task.

tasks/version_powershell.ps1
```diff
@@ -1,14 +1,16 @@
-if (Test-Path 'HKLM:\SOFTWARE\Puppet Labs\Puppet') {
-  $reg = $(Get-ItemProperty -Path 'HKLM:\SOFTWARE\Puppet Labs\Puppet')
-  if (Test-Path $reg.RememberedInstallDir64) {
+$rootPath = 'HKLM:\SOFTWARE\Puppet Labs\Puppet'
+
+$reg = Get-ItemProperty -Path $rootPath -ErrorAction SilentlyContinue
+if ($null -ne $reg) {
+  if ($null -ne $reg.RememberedInstallDir64) {
```
MikaelSmith: I think we only want to set `$loc` if the path exists, can we also `Test-Path`? If the registry entry is there but the file isn't, I interpret that as the package is not installed (but registry cleanup didn't succeed for some reason).

tasks/version_powershell.ps1
glennsarti: Fair enough.  Can just read it first before `write-Output`

tasks/version_powershell.ps1
glennsarti: We could add that in Line 12

``` powershell
if ( ($null -ne $loc) -and (Test-Path -Path $loc) ) {
```

MikaelSmith: Are you going to follow up on these changes, or would you like me to take over the PR?

glennsarti: Feel free to take over.  Just busy with other tasks first.

MikaelSmith:
Ready for a rebase.

PR #310, opened 2018-08-06
(maint) Restore correct dependencies with Beaker 4

MikaelSmith:
Add back dependencies we need that were removed in Beaker 4.


MikaelSmith:
Rebased.

PR #311, opened 2018-08-07
(BOLT-229) update to use new test helpers in BoltSpec and task_helpers

adreyer:
This depends on the follow PRs

https://github.com/puppetlabs/bolt/pull/549
https://github.com/puppetlabs/beaker-task_helper/pull/30

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -1,239 +1,60 @@
 # frozen_string_literal: true
 
 require 'spec_helper_acceptance'
+require 'beaker-task_helper/inventory'
+require 'bolt_spec/run'
 
 describe 'install task' do
-
-  # TODO: These tests should use beaker-task_helpers with BOLT-229
-  def exec(command)
-    stdout_str, _stderr, _status = Open3.capture3(*command)
-    stdout_str
-  end
-
-  def make_inventory
-    groups = []
-
-    def add_node(node, group_name, groups)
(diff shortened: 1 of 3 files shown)


task_spec/spec/acceptance/init_spec.rb
```diff
@@ -1,203 +1,45 @@
...
-        expect(res['status']).to eq('success')
-      end
-    end
-
-  end
+  include Beaker::TaskHelper::Inventory
+  include BoltSpec::Run
 
   def module_path
     RSpec.configuration.module_path
```
MikaelSmith: This feels like something we should maybe put in a helper?

task_spec/spec/spec_helper_acceptance.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
-require 'puppet'
```
MikaelSmith: We'll be able to go back to this when we unvendor puppet, right?

task_spec/spec/acceptance/init_spec.rb
adreyer: This is rspec specific so it doesn't seem like it belongs in beaker-task_helpers. I'm not sure it's the best way to track this either.

task_spec/spec/spec_helper_acceptance.rb
adreyer: yes

MikaelSmith: I think a comment reminding us of that would be nice.

PR #315, opened 2018-08-13
(maint) Revert to beaker 3.x.x

donoghuc:
This commit pins beaker-rspec at 6.2.3. This forces beaker to be pre 4.0.0. The combination of beaker 4.0.0 and beaker-puppet causes failures in acceptance tests tracked with https://tickets.puppetlabs.com/browse/BKR-1516.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,8 +54,7 @@ end
 group :system_tests do
   gem "puppet-module-posix-system-r#{minor_version}",                            :require => false, :platforms => "ruby"
   gem "puppet-module-win-system-r#{minor_version}",                              :require => false, :platforms => ["mswin", "mingw", "x64_mingw"]
-  gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '>= 3')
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -55,13 +55,12 @@ group :system_tests do
   gem "puppet-module-posix-system-r#{minor_version}",                            :require => false, :platforms => "ruby"
   gem "puppet-module-win-system-r#{minor_version}",                              :require => false, :platforms => ["mswin", "mingw", "x64_mingw"]
   gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '>= 3')
```
MikaelSmith: Could also do this by changing `>= 3` to `~> 3.37`.

donoghuc: My original thought process was that explicitly pinning back the dependency that blocked beaker 4 (beaker-rspec) would reflect the incremental progress to beaker 4 compatibility. For example specifically requiring the beaker-[hypervisor] libraries. Pinning beaker seems to be more straight forward though. Updated.

PR #320, opened 2018-08-23
(MODULES-7698) Fix OSX agent upgrades

ekinanp:


ekinanp:
I've tested this out on both Solaris 10 and OSX 10.12. Upgrades still work properly on Solaris 10, while on OSX 10.12, the race condition is no longer there and upgrade paths of the form "V1 => V2 => V3" work. My test set-up was a PE master with three PE versions installed:
# 2018.1.3
# 2018.1.4
# PE 2019.0 dev. build

Upgrades were tested by installing an older agent (e.g. 2018.1.3) and upgrading to a newer one (e.g. 2018.1.4). The "V1 => V2 => V3" path was done by upgrading "2018.1.3 => 2018.1.4 => PE 2019.0 dev. build". Note that I used my version of the puppet-agent module and placed it in the `/etc/puppetlabs/code/environments/production/modules` directory. Nodes were pointed to this manifest in the corresponding `site.pp` file.

manifests/install.pp
```diff
@@ -124,15 +121,24 @@
...
-      ensure          => 'present',
-      provider        => 'pkgdmg',
-      source          => "/opt/puppetlabs/packages/${package_file_name}",
-      require         => Class['puppet_agent::install::remove_packages'],
-      install_options => $install_options,
+    if $old_packages or $puppet_agent::aio_upgrade_required {
+      $install_script = "osx_install.sh.erb"
+  
+      contain puppet_agent::install::remove_packages
+
```
ekinanp: Dunno if it'd be useful to refactor this and the Sol 10. stuff into a defined type? Something like `perform_agent_upgrade { <platform> }`? I chose not to do it here since the code seems simple enough, and b/c our Sol 10 agents use ctrun to execute the script (here: https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/install.pp#L99) -- I don't want to introduce a weird param. like $script_wrapper when this code is only used twice.. I don't know why ctrun is used, as from what I understand that code basically runs the script in the background. We don't use any of ctrun's event reporting features. Maybe that could be refactored to just execute the script in the background? Or we could use a Mac-OS' equivalent of ctrun? I couldn't find any examples of the latter.

ekinanp:
CI passed. This should be ready for merge now.

branan:
Would be nice to have a more detailed description in the first commit about why we need to use ERB instead of EPP

ekinanp:
@branan Updated the PR with a more detailed description in the first commit.

templates/osx_install.sh.erb
```diff
@@ -0,0 +1,9 @@
+function random_hexdump() {
+  hexdump -n 2 -e '/2 "%u"' /dev/urandom
+}
+
+mountpoint="$(mktemp -d -t $(random_hexdump))"
```
branan: We should be able to rely on mktemp to generate a securely random filename here - `$(mktemp -d)` is probably just fine on its own?

ekinanp: Sure. For context, I ported this code over from https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/tasks/install_shell.sh#L427

ekinanp:
Updated the PR to remove the use of the `random_hexdump` function, instead using only `mktemp -d` in the OSX install code.

manifests/install.pp
branan: I don't think it's worth factoring out yet.

As for ctrun, I have no idea why we use that instead of just backgrounding the script.

ekinanp:
Squashed the commit. This should be ready for merge now.

PR #325, opened 2018-09-24
(BOLT-834) Allow upgrade from puppet5 to puppet6 for install task

donoghuc:
When the install_shell.sh implementation is used to install an agent from the puppet5 collection package managment for debian based systems (notibly ubuntu) fail to install puppet-agent from puppet6 collection. This commit allows upgrade from puppet5 to puppet6.

### DIFF
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -6,10 +6,19 @@ Param(
 # If an error is encountered, the script will stop instead of the default of "Continue"
 $ErrorActionPreference = "Stop"
 
-if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -match '^32') {
-    $arch = "x86"
-} else {
+try {
+  if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -match '^32') {
+      $arch = "x86"
+  } else {
+      $arch = "x64"
+  }
+}
+catch [System.Management.Automation.CommandNotFoundException] {
+  if (((Get-CimInstance -ClassName win32_OperatingSystem).OSArchitecture) -eq '64-bit') {
     $arch = "x64"
+  } else {
+    $arch = "x86"
+  }
 }
 
 if ($version) {
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -98,7 +98,6 @@ shift `expr $OPTIND - 1`
 machine=`uname -m`
 os=`uname -s`
 
-# CODEREVIEW: What are we doing with this?
 if [ -f /opt/puppetlabs/puppet/VERSION ]; then
   installed_version=`cat /opt/puppetlabs/puppet/VERSION`
 else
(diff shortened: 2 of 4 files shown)


tasks/install_shell.sh
```diff
@@ -407,8 +407,9 @@ install_file() {
       ;;
     "deb")
       info "installing puppetlabs apt repo with dpkg..."
-      dpkg -i "$2"
+      dpkg -i --force-conflicts "$2"
       apt-get update -y
+      apt-get install --fix-broken -y
```
nicklewis: I'm not sure whether we want to affect *other* packages that might be installed, with a broad `--fix-broken`...

tasks/install_shell.sh
```diff
@@ -407,8 +407,9 @@ install_file() {
       ;;
     "deb")
       info "installing puppetlabs apt repo with dpkg..."
-      dpkg -i "$2"
+      dpkg -i --force-conflicts "$2"
```
MikaelSmith: This is because the rpm writes to the same repo.d file? Since we know the contents, this seems fairly safe. It forces only one Puppet collection at a time.

tasks/install_shell.sh
MikaelSmith: What is `--fix-broken` addressing?

donoghuc: I can update with dpkg using the `--force-conflicts` flag. However when `apt-get` is used to install this results in a conflict between puppet5 and 6. 
```
The following packages have unmet dependencies:
 puppet6-release : Conflicts: puppet5-release but 5.0.0-2bionic is to be installed
E: Unmet dependencies. Try 'apt --fix-broken install' with no packages (or specify a solution).
```
The error message indicates running the `--fix-broken` install without listing packages. I am looking in to how to `specify a solution`.

donoghuc: Totally not the way to do it... I updated with some logic to handle this differently.

donoghuc:
Detect if a new collection is requested and delete the old release package if so. Currently implemented for  `.deb` and `.rpm`  install types. Still need figure out `.noarch.rpm` and `.dmg`.

tasks/install_shell.sh
```diff
@@ -398,6 +411,20 @@ install_file() {
       ;;
     "noarch.rpm")
       info "installing puppetlabs yum repo with zypper..."
+
+      if test "x$installed_version" != "xuninstalled"; then
```
donoghuc: Currently if the puppet5-release repo is installed dpkg, zypper, or rpm is used to remove it. This allows upgrade to puppet6. However going the "other way" seems like it would also require an apt-get, yum or zypper to also uninstall puppet-agent. Should we include this logic?

MikaelSmith: I think downgrades through collections can error.

tasks/install_shell.sh
```diff
@@ -389,6 +388,20 @@ install_file() {
         info "existing puppetlabs yum repo found, moving to old location"
         mv /etc/yum.repos.d/puppetlabs-pc1.repo /etc/yum.repos.d/puppetlabs-pc1.repo.old
       fi
+
+      if test "x$installed_version" != "xuninstalled"; then
```
donoghuc: I guess having the `x` prepended is unnecessary. Will remove before merge (not going to force push right away so I don't change the ref in the bolt PR)

tasks/install_powershell.ps1
```diff
@@ -6,10 +6,19 @@ Param(
...
-    $arch = "x86"
-} else {
+try {
+  if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -match '^32') {
+      $arch = "x86"
+  } else {
+      $arch = "x64"
+  }
+}
+catch [System.Management.Automation.CommandNotFoundException] {
```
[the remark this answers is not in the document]
donoghuc: When I was looking in to adding integration tests for puppet5/6. 

I was trying to run against our vagrant setup (mwrock/WindowsNano). I started debugging the install_powershell.ps1 puppet_agent::install implementation. I was getting the deprecation error for using `GetWmiObject` (note that part of this work was adding `bolt` user as local admin in the vagrant config).

I also had to use an alternate download method `Invoke-WebRequest`  to get the MSI because `system.net.webclient` was not available (that did not seem worth adding  at the time though I guess it could given that we have `wget`, `curl`, etc in the shell implementation). 

Once I added those workarounds I got to the `msiexec` step and found that windows installer (MSI) is not supported for windows nano (https://docs.microsoft.com/en-us/windows-server/get-started/deploy-nano-server). So I decided that I should focus on testing windows install at the acceptance level.

PR #329, opened 2018-09-30
Support for stdlib 5.x and apt 6.x

hfm:


tnmt:
it seems that resolved in ##356

hfm:
Certainly

PR #330, opened 2018-10-08
 (BOLT-878) Fail fast when run as non-root

adreyer:
There is not a good way to make sure a non-root user is installed on test images which makes this hard to test. I've manually tested on:

centos7
fedora 28
debian8
ubuntu-1804
sles 12
osx 10.13

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -41,19 +41,6 @@ exists() {
   fi
 }
 
-# Helper bug-reporting text
-report_bug() {
-  critical "Please file a bug report at https://github.com/puppetlabs/puppetlabs-puppet_agent"
-  critical ""
-  critical "Version: $version"
-  critical "Platform: $platform"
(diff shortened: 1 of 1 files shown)


faucct:
As `bolt` only collects custom facts in `apply_prep` task, which first makes sure that the agent is installed, failing fast makes it impossible to gather custom facts when running as non-root even if the agent is already installed. Can this be reverted?

MikaelSmith:
My memory was that Bolt only runs this task if it detects the agent isn't present. Is that not how it's behaving?

faucct:
As I understand, it performs some static check which ssh executor fails to pass no matter what, so this script runs anyway. 
Here is the fix: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/475#issuecomment-597725231

MikaelSmith:
Oh, it changed. It used to run the `puppet_agent::version` task first to see if it was installed.

PR #331, opened 2018-10-10
(BOLT-915) Use facts module to query for target platform

donoghuc:
Much of the puppet_agent task is logic to determine the operating systems and version of the target. This is duplicated by the shell implementation of the facts task. The addition of multi-file task support allows using the facts module bash.sh implementation code. The bash.sh implementation has been modified to return platform and version when environment variables PLATFORM=TRUE or RELEASE=TRUE are set. This commit utilizes the code in the facts module to determine platform and version.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,7 +66,7 @@ group :system_tests do
   gem "puppet-blacksmith", '~> 3.4',                                             :require => false
   # Bundler fails on 2.1.9 even though this group is excluded
   if ENV['GEM_BOLT']
-    gem 'bolt', '~> 0.21.8', require: false
+    gem 'bolt', '~> 1.1.0', require: false
     gem 'beaker-task_helper', '~> 1.5.2', require: false
   end
 end
--- a/task_spec/.fixtures.yml
+++ b/task_spec/.fixtures.yml
@@ -1,3 +1,7 @@
 fixtures:
+  forge_modules:
+    facts:
+      repo: "puppetlabs/facts"
+      ref: "0.4.1"
   symlinks:
     puppet_agent: "#{File.absolute_path(File.join(source_dir, '..'))}"
--- a/task_spec/spec/spec_helper_acceptance.rb
+++ b/task_spec/spec/spec_helper_acceptance.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-
 # Make sure bolt's puppet is loaded
 require 'bolt/pal'
 Bolt::PAL.load_puppet
@@ -21,8 +20,4 @@ RSpec.configure do |c|
   c.add_setting :module_path
   c.module_path  = File.join(base_dir, 'fixtures', 'modules')
 
-  # Configure all nodes in nodeset
-  c.before :suite do
-    #run_puppet_access_login(user: 'admin') if pe_install?
-  end
 end
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 # Install puppet-agent as a task
 #
 # From https://github.com/petems/puppet-install-shell/blob/master/install_puppet_5_agent.sh
@@ -25,12 +25,6 @@ critical () {
     log "CRIT: ${1}"
 }
 
-utopic () {
-    warn "There is no utopic release yet, see https://tickets.puppetlabs.com/browse/CPR-92 for progress";
-    warn "We'll use the trusty package for now";
-    deb_codename="trusty";
-}
-
 # Check whether a command exists - returns 0 if it does, 1 if it does not
 exists() {
   if command -v $1 >/dev/null 2>&1
@@ -42,7 +36,6 @@ exists() {
 }
 
 # Get command line arguments
-
 if [ -n "$PT_version" ]; then
   version=$PT_version
 fi
@@ -53,15 +46,13 @@ else
   collection='puppet'
 fi
 
-shift `expr $OPTIND - 1`
-
-os=`uname -s`
-
+# Error if non-root
 if [ `id -u` -ne 0 ]; then
   echo "puppet_agent::install task must be run as root"
   exit 1
 fi
 
+# Track to handle puppet5 to puppet6
 if [ -f /opt/puppetlabs/puppet/VERSION ]; then
   installed_version=`cat /opt/puppetlabs/puppet/VERSION`
 else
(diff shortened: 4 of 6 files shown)


donoghuc:
Depends on https://github.com/puppetlabs/puppetlabs-facts/pull/13

MikaelSmith:
Do we use any facts for the powershell task? I'm guessing no.

donoghuc:
The only fact we care about in powershell is the architecture. In the case where puppet agent is already installed we could use the facts task to find that. In the (probably more prevalent) case that there is no existing agent the facts task only gets:
```
  {
    "os": {
      "name": "windows",
      "release": {
        "full": "2016",
        "major": "2016"
      },
      "family": "windows"
    }
  }
``` 
I think there is a problem with how I am using fixtures.yaml. Tests are passing locally for me against the docker containers in travis as well as all of the targets in task acceptance and bolt acceptance. We will likely want to wait for a bolt release (for multi file task support) anyway.

donoghuc:
I built an experimental pipeline testing these changes.
Changes Include
1. Bolt is tracking master branch to pick up multi-file support 
2. puppetlabs-facts is tracking a commit to my fork with the supporting changes to the facts::bash implementation
Once the next bolt release goes out and puppetlabs-facts is release I can change the references and re-test against the experimental pipeline.

task_spec/Boltdir/Puppetfile
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+forge "http://forge.puppetlabs.com"
+
+mod 'puppetlabs-facts',
```
donoghuc: I ran in to multiple problems trying to support the `repositories` key in `.fixtures.yml` for Travis CI and Jenkins CI. I found that just installing facts with bolt and then using fixtures.yml to build a symlink to be a simpler solution.

adreyer: I'm concerned we're still using fixtures.yaml to "install" the puppet_agent module for testing but a pupeptfile for the facts module. I think we need to manage all code with a single tool

donoghuc: I think that as soon as we release fact and do not need to point to the `donoghuc` fork of the repository within fixtures.yml we can use only `fixtures.yml` instead of `Puppetfile`. I just tested in Jenkins and it works there. 

When the commit is merged to puppetlabs-facts I will see if I can clone the repo in Travis from puppetlabs fork. 

The message for Travis is: 
```
2.14s$ cd task_spec && bundle exec rake task_acceptance
Cloning into 'spec/fixtures/modules/facts'...
Warning: Permanently added the RSA host key for IP address '192.30.253.112' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
rake aborted!
Failed to clone git repository git@github.com:donoghuc/puppetlabs-facts.git into spec/fixtures/modules/facts
/home/travis/.rvm/gems/ruby-2.3.1/gems/puppetlabs_spec_helper-2.11.0/lib/puppetlabs_spec_helper/tasks/fixtures.rb:136:in `clone_repo'
/home/travis/.rvm/gems/ruby-2.3.1/gems/puppetlabs_spec_helper-2.11.0/lib/puppetlabs_spec_helper/tasks/fixtures.rb:300:in `block (3 levels) in <top (required)>'
Tasks: TOP => task_acceptance => spec_prep
(See full trace by running task with --trace)
```

Gemfile
```diff
@@ -66,7 +66,7 @@ group :system_tests do
   gem "puppet-blacksmith", '~> 3.4',                                             :require => false
   # Bundler fails on 2.1.9 even though this group is excluded
   if ENV['GEM_BOLT']
-    gem 'bolt', '~> 0.21.8', require: false
+    gem 'bolt', git: "https://github.com/puppetlabs/bolt.git"
```
MikaelSmith: Can we pin to a specific release, like `~> 1.1`?

task_spec/.fixtures.yml
```diff
@@ -1,3 +1,8 @@
 fixtures:
+  repositories:
+    facts:
+      repo: "git@github.com:donoghuc/puppetlabs-facts.git"
```
MikaelSmith: The README should definitely note the dependency on puppetlabs-facts. I'm leaning against putting it in metadata.json yet because it's only required for tasks.

donoghuc:
The facts module has been released to forge and changes are available in bolt. 

Experimental pipeline PR pointing at this PR is at: https://github.com/puppetlabs/ci-job-configs/pull/5191

Github seems to temporarily be having issues with web hooks so PR testing in Jenkins is not working yet.

task_spec/spec/spec_helper_acceptance.rb
```diff
@@ -13,6 +13,10 @@
 
 base_dir = File.dirname(File.expand_path(__FILE__))
 
+# Download required facts module
```
lucywyman: Why the comments?

donoghuc:
Tested with https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental%20automatic/job/experimental_auto_donoghuc-puppet-agent-module_intn-run_task_acceptance-BOLT-915/

tasks/install_shell.sh
```diff
@@ -53,91 +46,68 @@ else
...
+    case $major_version in
+      "10.11") platform_version="10.11";;
+      "10.12") platform_version="10.12";;
+      "10.13") platform_version="10.13";;
+      *) echo "No builds for platform: $major_version"
+         exit 1
+         ;;
+    esac
   fi
-elif test -f "/etc/arch-release"; then
```
lucywyman: I assume we can remove these since they're covered by the facts module now?

tasks/install_shell.sh
```diff
@@ -191,12 +155,6 @@ if test "x$platform_version" = "x"; then
   exit 1
 fi
 
-if test "x$platform" = "xsolaris2"; then
```
lucywyman: Are we no longer supporting solaris?

tasks/install_shell.sh
```diff
@@ -444,95 +399,84 @@ install_file() {
   fi
 }
 
-#Platforms that do not need downloads are in *, the rest get their own entry.
+info "Downloading Puppet $version for ${platform}..."
 case $platform in
-  "archlinux")
```
lucywyman: I'm not seeing where these got replaced / moved to?

tasks/install_shell.sh
```diff
@@ -444,95 +399,84 @@ install_file() {
...
+    filetype="dmg"
+    if test "$version" = "latest"; then
+      filename="puppet-agent-latest.dmg"
+    else
+      filename="puppet-agent-${version}-1.osx${platform_version}.dmg"
+    fi
+    download_url="http://downloads.puppetlabs.com/mac/${collection}/${platform_version}/x86_64/${filename}"
+    ;;
+  *)
+    critical "Sorry $platform is not supported yet!"
```
lucywyman: Ah, just get caught here for all other platforms?

lucywyman:
Been a while since I read  bash, but besides the supported OS stuff looks good to me

donoghuc:
Never worked for Solaris or Arch. Not supported for the puppet_agent::install task. I think supporting those would be separate ticket.

PR #332, opened 2018-10-11
(MODULES-7791) Remove deprecated source_pemissions

caseywilliams:
`source_permissions => use` is deprecated; remove it to quiet warnings.

The suggested alternative to `source_permissions => use` for puppet 6 is to explicitly set user, group, and mode. I'm unsure how to best approach doing that for all the files in this part of the module, but curious to see whether the defaults are adequate.


caseywilliams:
Alternatively, would we want to just let the warnings happen for 1.x only?

PR #333, opened 2018-10-11
Allow "present" or "latest" as a package version

nabertrand:
It appears that the module supported setting the package version to `present` or `latest` prior to c2206a7e9b882041ccb569332fc872f9524a34cd. This commit restores that functionality.


GabrielNagy:
Should be superseded by https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/565

PR #334, opened 2018-10-12
(MODULES-8086) Puppet 5 and 6: wrong urls for windows msi

fculpo:
Correct URLs for windows msis (Puppet 5 and 6)

### DIFF
--- a/manifests/windows/install.pp
+++ b/manifests/windows/install.pp
@@ -19,7 +19,12 @@ class puppet_agent::windows::install(
     $_https_source = "https://pm.puppetlabs.com/puppet-agent/${_pe_server_version}/${_agent_version}/repos/windows/${package_file_name}"
   }
   else {
-    $_https_source = "https://downloads.puppetlabs.com/windows/${package_file_name}"
+    $dir = $package_file_name ? {
+      /.*-5\..*/ => 'puppet5/',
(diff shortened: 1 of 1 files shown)


manifests/windows/install.pp
```diff
@@ -19,7 +19,12 @@
     $_https_source = "https://pm.puppetlabs.com/puppet-agent/${_pe_server_version}/${_agent_version}/repos/windows/${package_file_name}"
   }
   else {
-    $_https_source = "https://downloads.puppetlabs.com/windows/${package_file_name}"
+    $dir = $package_file_name ? {
+      /.*-5\..*/ => 'puppet5/',
```
smcclellan: Can this regex be a bit more specialized, to ensure it only matches what we expect? E.g. `puppet-agent-5\..*` might be better.

caseywilliams: I'm curious whether this `$dir` value could just be derived from `$::puppet_agent::collection`, which will be set to PC1 (for puppet 4), puppet5, or puppet6 -- I will take a look and see if that could work.

caseywilliams: This could work, but the collection will still be the default (PC1) at this point unless someone has specified a value for it explicitly -- using a regex on the package name does seem to be the correct way to get this done

PR #335, opened 2018-10-15
 (MODULES-7760) Remove dist_tag in install.pp for RHEL platforms

ekinanp:
The dist_tag variable was introduced as a workaround for
PUP-5802/PUP-5025. These tickets were resolved two years
ago, so the workaround is no longer necessary.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -201,7 +201,6 @@ class puppet_agent::install(
       install_options => $install_options,
     }
   } elsif ($::osfamily == 'Debian') and ($package_version != 'present') {
-    # Workaround PUP-5802/PUP-5025
     package { $::puppet_agent::package_name:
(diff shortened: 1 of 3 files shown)


ekinanp:
Actually, wouldn't we still need this workaround in case the package_version did contain an epoch (at least for RHEL 7)? @branan Your comment in https://tickets.puppetlabs.com/browse/PUP-5025 seems to indicate that it'd fail on RHEL 7 for a nonzero epoch unless the architecture's explicitly specified.

ekinanp:
Should this also be targetted at 1.x?

ekinanp:
Don't know what to do about the Travis failure -- https://travis-ci.org/puppetlabs/puppetlabs-puppet_agent/jobs/441586316

It is something unrelated to these changes.

PR #337, opened 2018-10-18
fix platform_tag undefined check so it actually works - making puppet…

KlavsKlavsen:
…_agent working for opensource puppet with strict_variables


GabrielNagy:
Hi,

There's been a major refactor of the module, and the code in question was removed at some point.

We tried to reproduce this issue in MODULES-8080 but we couldn't.

If you encounter this again, feel free to reopen the issue/PR.

Thanks!

PR #338, opened 2018-10-18
Remove source permissions

KlavsKlavsen:
as they are deprecated. They were only used on windows_clients.. I assume when copying files from old puppet installation.. (so source was local windows filesystem). AFAIK it should still work without this though.


gguillotte:
Looks like https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/332 merged 31 October removed the same lines from prepare.pp. CC @branan

gimmyxd:
Closing this at it was fixed in #332

PR #343, opened 2018-10-31
(maint) Use legacy dependencies repo for activemq

geoffnichols:
Since activemq packages are currently only available in our legacy
dependencies repo, install the legacy puppetlabs-release package.


geoffnichols:
Apologies, I probably should have targeted the earlier PR against 1.x instead of master!

PR #347, opened 2018-12-06
(feature) add scientific support to install task

tphoney:


HelenCampbell:
+1 LGTM

MikaelSmith:
We should be able to extend the Jenkins test matrix to cover those platforms.

PR #348, opened 2018-12-07
(MODULES-8318) Update rspec-puppet, add yumrepo_core fixtures for puppet 6

caseywilliams:
There are two commits here, both of which are updates to dependencies:

- Update rspec-puppet (2.6.9 to 2.7.2)
- Add yumrepo_core as a fixture to support running spec tests with puppet 6 gems (which can't manage yumrepo resources without the module)

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -11,5 +11,7 @@ fixtures:
     apt:
       repo: "puppetlabs/apt"
       ref: "2.3.0"
+    yumrepo_core:
+      repo: "puppetlabs/yumrepo_core"
   symlinks:
     puppet_agent: "#{source_dir}"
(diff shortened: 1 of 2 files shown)


.fixtures.yml
```diff
@@ -11,5 +11,7 @@ fixtures:
     apt:
       repo: "puppetlabs/apt"
       ref: "2.3.0"
+    yumrepo_core:
+      repo: "puppetlabs/yumrepo_core"
```
joshcooper: FYI it's possible to make this conditional for 6.0 and up only, see https://github.com/puppetlabs/puppetlabs_spec_helper/pull/258. I don't think this PR needs to change (since the latest version of yumrepo_core supports older puppet versions), but just something to keep in mind in the future.

caseywilliams: Thanks! I didn't know that -- I think I'll make a followup PR to do that, just in case.

PR #350, opened 2018-12-10
(MODULES-8318) Remove spec tests for puppet < 4

caseywilliams:
Upgrades from puppet 3 won't be supported in release 2.0 of this module.
This removes spec tests specific to puppet < 4, and updates a few
stubbed values to be more consistent with versions in puppet 4+.


spec/classes/puppet_agent_prepare_stringify_facts_spec.rb
```diff
@@ -1,49 +0,0 @@
...
-          context 'on Puppet 3 or lower' do
-            it {
-              is_expected.to contain_ini_setting('puppet stringify_facts').with(
-                ensure: 'present',
-                value: false
-              )
-            }
-          end
-        else
-          context 'on Puppet 4 or higher' do
```
ekinanp: Why's this test removed?

ekinanp:
[spec/classes/puppet_agent_windows_install_spec.rb](https://github.com/caseywilliams/puppetlabs-puppet_agent/blob/MODULES-8318/update-specs-master/spec/classes/puppet_agent_windows_install_spec.rb) also has some Puppet 3 stuff that should probably be removed.

spec/classes/puppet_agent_prepare_stringify_facts_spec.rb
caseywilliams: I can't actually find stringify_facts at all anymore in the puppet 4+ docs. [These deprecation notes for 3](https://puppet.com/docs/puppet/3.8/deprecated_settings.html#stringifyfacts--true) suggest that it's still possible to set it to true there, though, so I will work on restoring this unless someone else has more info.

spec/classes/puppet_agent_osfamily_darwin_spec.rb
```diff
@@ -75,40 +66,17 @@
...
-                facts.merge({
-                  :aio_agent_version => '1.0.0'
-                })
-              end
-
-              it { is_expected.not_to contain_exec('forget puppet-agent') }
+          context 'aio_agent_version is out of date' do
+            let(:facts) do
+              facts.merge({
+                :aio_agent_version => '0.0.1'
```
speedofdark: Should this match the 1.0.0 in line 103 in the original source?

PR #351, opened 2018-12-11
(PE-25228) Add Fedora 29 (x86_64) to the puppet_agent module

gimmyxd:
This is dependent on: https://github.com/puppetlabs/puppetlabs-pe_repo/pull/533


speedofdark:
I'm not certain if the intent of this ticket is to add the release to things, or just check if the release works. When I did Fedora 28, I ran some tests, but didn't change any code...?

gimmyxd:
@speedofdark this enables running a subset of tests on Fedora 29. Taking into consideration that Fedora 27 is reaching end of support I think is a good idea to add this kinds of tests to the major version.

speedofdark:
Makes sense to me. Thanks.

mihaibuzgau:
@smcclellan can you please review this PR.

mihaibuzgau:
thanks @smcclellan !

PR #352, opened 2018-12-19
(PA-2282) Add Docker workflow for iterative development

smcclellan:
This commit decreases the feedback cycle time to enable iterative
development. The gist is that a developer can use Docker to test out
upgrades on a machine.

The `docker/bin/upgrade.sh [<PLATFORM>] [<BEFORE>] [<AFTER>]`
command builds a container on the `<PLATFORM>` platform that has the
`<BEFORE>` version of puppet on it and adds the `puppet_agent`
module. It then runs `puppet apply` in the container such that the
`<AFTER>` version of puppet exists afterward. This script ends with
a shell inside the container.

Available platforms are Ubuntu Xenial (`ubuntu`) and CentOS 7
(`centos`).

Also available is `docker/bin/versions.sh [<PLATFORM>]`, which will
output which versions of puppet are available on a given
`<PLATFORM>`.

For full details, see docker/ubuntu/Dockerfile or
docker/centos/Dockerfile.

### DIFF
--- a/docker/centos/Dockerfile
+++ b/docker/centos/Dockerfile
@@ -1,0 +1,107 @@
+# This Dockerfile enables an iterative development workflow where you can make
+# a change and test it out quickly. The majority of commands in this file will
+# be cached, making the feedback loop typically quite short. The workflow is
+# as follows:
+#   1. Set up pre-conditions for the system in puppet code using `deploy.pp`.
+#   2. Make a change to the module.
+#   3. Run `docker build -f docker/Dockerfile .` or
+#      `./docker/bin/upgrade.sh centos` from the project directory. If you would
+#      like to test specific version upgrades, you can add run this like so:
+#        `docker build -f docker/centos/Dockerfile . \
+#           -t pa-dev:centos --build-arg before=1.10.14`
+#   4. Upgrade the container by running the image:
+#        `docker run -it pa-dev:centos`
+#      Specify your upgrade TO version as an argument to the `docker run`
+#      command.
+#   5. Review the output. Repeat steps 2-5 as needed.
+#
+# At the end of execution, you will see a line like:
+#
+# Notice: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]/ensure: ensure changed '1.10.14-1.el7' to '6.2.0'
+#
+# This specifies the versions that were used for upgrade.
+#
+# Arguments:
+# - before: The version to do upgrade FROM. Default: "1.10.14"
+
+FROM centos:7
+
+# Use this to force a cache reset (e.g. for output purposes)
+#COPY $0 /tmp/Dockerfile
+
+# Install some other dependencies for ease of life.
+RUN  yum update -y \
+  && yum install -y wget git \
+  && yum clean all
+
+ARG before=1.10.14
+LABEL before=${before}
+
+# Install proper FROM repo: PC1 (puppet 4), puppet 5, or puppet 6.
+RUN if [[ ${before} == 1.* ]]; then \
+        echo Installing PC1 repo; \
+        wget -O puppet-pc1.rpm http://yum.puppetlabs.com/puppetlabs-release-pc1-el-7.noarch.rpm && \
+        rpm -i puppet-pc1.rpm; \
+    elif [[ ${before} == 5.* ]]; then \
+        echo Installing PC1 repo; \
+        wget -O puppet5.rpm http://yum.puppetlabs.com/puppet5/puppet5-release-el-7.noarch.rpm && \
+        rpm -i puppet5.rpm; \
+    elif [[ ${before} == 6.* ]]; then \
+        echo Installing PC1 repo; \
+        wget -O puppet6.rpm http://yum.puppetlabs.com/puppet6/puppet6-release-el-7.noarch.rpm && \
+        rpm -i puppet6.rpm; \
+    else echo no; \
+    fi
+
+# Print out which versions of the puppet-agent package are available (for reference).
+#RUN yum list puppet-agent --showduplicates
+
+# Install FROM version of puppet-agent.
+RUN yum -y update && \
+    yum install -y puppet-agent-${before}-1.el7
+
+# This is also duplicated in the docker/bin/helpers/run-upgrade.sh.
+ENV module_path=/tmp/modules
+WORKDIR "${module_path}/puppet_agent"
+COPY metadata.json ./
+
+# Dependency installation: Forge or source? The former is what the user will
+# have downloaded, but the latter allows testing of version bumps.
+# Install module dependencies from the Forge using Puppet Module Tool (PMT).
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-stdlib
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-inifile
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-apt
+
+# Installing dependencies from source. These versions should be within the range
+# of `dependencies` in metadata.json. `translate` is a dependency for inifile.
+#RUN git clone https://github.com/puppetlabs/puppetlabs-stdlib ../stdlib --branch 5.2.0
+#RUN git clone https://github.com/puppetlabs/puppetlabs-inifile ../inifile --branch 2.5.0
+#RUN git clone https://github.com/puppetlabs/puppetlabs-translate ../translate --branch 1.2.0
+#RUN git clone https://github.com/puppetlabs/puppetlabs-apt ../apt --branch 6.3.0
+
+# Check that all dependencies are installed.
+RUN /opt/puppetlabs/puppet/bin/puppet module --modulepath $module_path list --tree
+COPY docker/deploy.pp /tmp/deploy.pp
+RUN ["sh", "-c", "/opt/puppetlabs/puppet/bin/puppet apply --modulepath $module_path /tmp/deploy.pp"]
+
+# Now move the project directory's files into the image. That way, if these
+# files change, caching will skip everything before this.
+COPY docker/bin/helpers/run-upgrade.sh /tmp/bin/run-upgrade.sh
+COPY files/ ./files/
+COPY locales/ ./locales/
+COPY spec/ ./spec/
+COPY task_spec/  ./task_spec/
+COPY tasks/ ./tasks/
+COPY templates/ ./templates
+COPY types/ ./types/
+COPY Gemfile Gemfile.lock Rakefile ./
+COPY lib/ ./lib/
+COPY manifests/ ./manifests/
+
+COPY docker/upgrade.pp /tmp/upgrade.pp
+
+# Print out which versions of the puppet-agent package are available (for reference).
+#RUN yum list puppet-agent --showduplicates
+
+# Perform the upgrade.
+ENTRYPOINT ["/tmp/bin/run-upgrade.sh"]
--- a/docker/ubuntu/Dockerfile
+++ b/docker/ubuntu/Dockerfile
@@ -1,0 +1,111 @@
+# This Dockerfile enables an iterative development workflow where you can make
+# a change and test it out quickly. The majority of commands in this file will
+# be cached, making the feedback loop typically quite short. The workflow is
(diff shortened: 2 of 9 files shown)


smcclellan:
More comprehensive documentation will be a later part of PA-2282.

smcclellan:
`docker/bin` seems to be in the `.gitignore`.

ekinanp:
If I understand things correctly, the code in the Dockerfile seems to test upgrades on an Ubuntu image and it is also for FOSS upgrades only, right? If so, then we can't test platform-specific upgrades like AIX, Windows, Solaris, OSX and we can't test the more common case of upgrading an agent tied to a PE master (in an iterative way, at least). Do you have any plans on reconciling that?

One idea off the top of my head could be to containerize the PE master (which can be done with our internal builds of PE) and have it contain pe_repo packages for all of our supported platforms. You'd basically Dockerize the workflow I described on HipChat. I don't think this would be a very difficult thing to do at least for PE 2017.3.x+ since the PE console has a reasonably intuitive API that you could use to automatically include those pe_repo classes -- see, for example, https://github.com/ekinanp/scripts/blob/master/bin/pe-utils#L231-L276; it would be tough to do for PE 2016.4.x, however, since that API's not available. Anyways, if we containerize the PE master, then we could have some rake tasks in there that people could use to provision an agent machine that would test their upgrade scenario (or maybe have some simple bash scripts for this).

Just some thoughts.

smcclellan:
@ekinanp This is indeed only for Ubuntu, and only for FOSS. My goal is that we eventually (if needed) add more platforms via separate Dockerfiles.

As for including PE, my understanding is that we want to move away from requiring a full PE setup for developing on this module. Are there sufficient use cases to require development-side testing of upgrades for a PE agent? I thought that given all the AIO agent work we've done, the agent would be independent enough to work the same with both masters. IMO ideally we would avoid bringing PE into the development workflow for both the open-source puppet-agent package and the open-source puppet-agent module. Thoughts? /cc @branan for input.

ekinanp:
@smcclellan That makes sense. Note that not all platforms can be added with Docker, for example we cannot do this for AIX, Solaris, OSX and Windows and these are the platforms where upgrades are a PITA.

There's some PE-specific code in here, but maybe the acceptance tests are enough to cover it (i.e. an iterative workflow wouldn't be of much benefit). I agree that @branan would know more about our plans there.

smcclellan:
Edited the workflow such that the Dockerfile (build stage) ends just before the upgrade is performed, and the upgrade is the first command that's executed (`CMD` as a default, overridden in `run-ubuntu.sh`). That way the final image isn't cached and version changes don't build any new images.

geoffnichols:
@smcclellan this seems pretty great! 

One question - when I execute `./docker/bin/run-ubuntu.sh` I'm seeing the error below - is this expected?

```
Successfully tagged pa-dev:ubuntu
Warning: /etc/puppetlabs/puppet/hiera.yaml: Use of 'hiera.yaml' version 3 is deprecated. It should be converted to version 5
   (in /etc/puppetlabs/puppet/hiera.yaml)
Notice: Compiled catalog for 70085b8bf486.vpn.puppet.net in environment production in 0.57 seconds
Notice: /Stage[main]/Apt/File[preferences]/ensure: created
Notice: /Stage[main]/Apt/Package[dirmngr]/ensure: created
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/File[/etc/pki]/ensure: created
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/File[/etc/pki/deb-gpg]/ensure: created
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/File[/etc/pki/deb-gpg/GPG-KEY-puppetlabs]/ensure: defined content as '{md5}7b4ed31e1028f921b5c965df0a42e508'
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/File[/etc/pki/deb-gpg/GPG-KEY-puppet]/ensure: defined content as '{md5}16e3e148bc861ee66906e475f8342f81'
Notice: /Stage[main]/Apt/Apt::Setting[conf-update-stamp]/File[/etc/apt/apt.conf.d/15update-stamp]/ensure: defined content as '{md5}0962d70c4ec78bbfa6f3544ae0c41974'
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Apt::Source[pc_repo]/Apt::Setting[list-pc_repo]/File[/etc/apt/sources.list.d/pc_repo.list]/ensure: defined content as '{md5}db0d9973a9532f1746820759e3a4ea5f'
Notice: /Stage[main]/Apt::Update/Exec[apt_update]: Triggered 'refresh' from 1 events
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Exec[pc_repo_force]/returns: forcing apt update for pc_repo PC1
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Exec[pc_repo_force]: Triggered 'refresh' from 2 events
Notice: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]/ensure: ensure changed '1.10.14-1xenial' to '6.1.0-1xenial'
Error: Could not find init script for 'puppet'
Error: /Stage[main]/Puppet_agent::Service/Service[puppet]/ensure: change from stopped to running failed: Could not find init script for 'puppet'
Error: Could not find init script for 'mcollective'
Error: /Stage[main]/Puppet_agent::Service/Service[mcollective]/ensure: change from stopped to running failed: Could not find init script for 'mcollective'
Notice: Applied catalog in 21.20 seconds
```

smcclellan:
After hitting issues with Docker + systemd, I have made revisions so that the puppet service is not started up upon running the container. I've also added support for CentOS 7.

docker/bin/run-upgrade.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
speedofdark: Not `!/usr/bin/env bash`?

docker/bin/upgrade-ubuntu.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
speedofdark: These are very similar. Have you considered making the platform a parameter?

docker/bin/upgrade-centos.sh
```diff
@@ -0,0 +1,17 @@
...
+#
+# Builds an upgrade process for the puppet-agent module and tags as
+# "pa-dev:centos".
+#
+# Parameters:
+# - BEFORE: The puppet-agent package version that is installed prior to upgrade.
+# - AFTER: The puppet-agent package version that should exist after upgrade.
+set -e
+
+cd "$(dirname "$0")/../.."
```
speedofdark: I'm surprised parsing this line by the shell works as expected...

docker/centos/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+
+# Dependency installation: Forge or source? The former is what the user will
+# have downloaded, but the latter allows testing of version bumps.
+# Install module dependencies from the Forge using Puppet Module Tool (PMT).
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-stdlib
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-transition
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-inifile
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-apt
+
+# Installing dependencies from source.
```
speedofdark: does this need some docs around the versions being fetched from GitHub?

docker/ubuntu/Dockerfile
```diff
@@ -0,0 +1,101 @@
...
+
+# Dependency installation: Forge or source? The former is what the user will
+# have downloaded, but the latter allows testing of version bumps.
+# Install module dependencies from the Forge using Puppet Module Tool (PMT).
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-stdlib
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-transition
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-inifile
+RUN /opt/puppetlabs/puppet/bin/puppet module install --modulepath $module_path --target-dir .. puppetlabs-apt
+
+# Installing dependencies from source.
```
speedofdark: does this need some docs around the versions being fetched from GitHub?

docker/upgrade.pp
```diff
@@ -0,0 +1,11 @@
+node default {
+  class { '::puppet_agent':
+    package_version => $facts['to_version'],
+    # Upgrades in Docker cannot start the puppet service due to systemd
```
McdonaldSeanp: In the past this has been poorly documented, but `collection` is a required parameter when performing upgrades in FOSS scenarios.

We have hopefully updated the module's README enough to reflect this now.

```
    collection => $facts['to_collection'],
```

docker/centos/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+
+# Install some other dependencies for ease of life.
+RUN  yum update -y \
+  && yum install -y wget git \
+  && yum clean all
+
+# Install several repos: PC1 (puppet 4), puppet 5, and puppet 6.
+RUN wget -O puppet-pc1.rpm http://yum.puppetlabs.com/puppetlabs-release-pc1-el-7.noarch.rpm && \
+    rpm -i puppet-pc1.rpm --force --replacefiles && \
+    wget -O puppet5.rpm http://yum.puppetlabs.com/puppet5/puppet5-release-el-7.noarch.rpm && \
```
McdonaldSeanp: One of the tasks the module performs is installing these repos. Can we parameterize which repo gets installed so the development workflow can also test out the "install the repo" functionality?

docker/centos/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+# Install early version of puppet-agent.
+RUN yum -y update && \
+    yum install -y puppet-agent-${before}-1.el7
+
+# This is also duplicated in the docker/bin/run-ubuntu.sh.
+ENV module_path=/tmp/modules
+WORKDIR "${module_path}/puppet_agent"
+COPY metadata.json ./
+
+# Dependency installation: Forge or source? The former is what the user will
```
McdonaldSeanp: PMT allows us to install specific versions, so I think we should stick with the forge. We don't use any of the functionality of these other modules to a large extent anyway, so I don't expect us to ever require dev versions of one of these modules in the development or testing of the puppet_agent module

docker/bin/upgrade-centos.sh
smcclellan: Yes, it allows execution of the file from any directory.

docker/centos/Dockerfile
smcclellan: I do imagine though that if we ever considered a version bump on these dependencies, running a quick sanity test that an upgrade works might be useful. I'm also seeing this workflow as a prototype for testing other modules, so I'd like to preserve a section (even if it's documented out) enabling custom dependency versions.

docker/bin/upgrade-ubuntu.sh
smcclellan: I ended up making this `upgrade.sh` and `versions.sh`, not a big deal to leave the platform name out of the script name actually.

smcclellan:
Added support for multiple platforms: `./docker/bin/upgrade.sh centos,ubuntu 5.4.0 6.2.0`.

docker/upgrade.pp
```diff
@@ -0,0 +1,12 @@
...
```
caseywilliams: beaker-docker makes systemd work by using `/sbin/init` as the `CMD`. I'm not sure if it's possible to successfully work that in here, but perhaps worth a try. A huge caveat for that, though, is that the container must run in privileged mode for it to work. This can be particularly scary when there's no docker machine in between your laptop and the container (see https://github.com/puppetlabs/beaker-hostgenerator/pull/138 for an example problem that can arise if you're not careful)

smcclellan: I took a look at `/sbin/init` and even using `centos/systemd` as the base container, but `/sbin/init` is mostly for starting services by name and I couldn't find a way to run the upgrade within that context.

PR #353, opened 2018-12-19
(MODULES-8393) Add task required metadata, hide extra implementations

MikaelSmith:
Adds task metadata required to publish to the Forge. Also declare
input_method where it would use the default `both` and hide extra
implementations for task runners that implement private tasks and task
implementations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,23 +16,21 @@ before_install:
 matrix:
   fast_finish: true
   include:
-  - rvm: 2.4.1
-    env: CHECK="validate lint"
-  - rvm: 2.1.9
-    env: PUPPET_GEM_VERSION="4.7.1" HIERA_GEM_VERSION="3.2.2" CHECK=spec
-  - rvm: 2.1.9
-    env: PUPPET_GEM_VERSION="~> 4.10" CHECK=spec
-  - rvm: 2.4.1
+  - rvm: 2.5.3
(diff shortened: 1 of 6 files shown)


branan:
Would be nice to see in the commit message that this is related to *task* metadata, since this module has been published to the forge for years.

MikaelSmith:
Done. Shipping to the forge with the current set of tasks would've failed because the forge requires that every task has a metadata file.

branan:
I'd like someone who actually knows what task metadata looks like to approve this, but nothing jumps out to me as evil :+1:

MikaelSmith:
Looks like Ruby pre-2.3 may be borked.

MikaelSmith:
Should I drop `gem update --system`, or the old Puppet versions?

branan:
We want this module to maintain support for Puppet 4, even in the master branch, so that lagging users can continue to use it to update. I believe the `gem update --system` was added to work around a bug in Travis, so I'm not sure we can drop that.

branan:
In other words, I have NFI what to do here. I'm fine merging this over the red, since it's just a metadata change. We'll have to work out the travis mess separately as part of readying this for a 2.0 release

MikaelSmith:
Trying testing with Puppet 4 + Ruby 2.3. Not ideal, but better than red tests.

PR #354, opened 2018-12-19
(MODULES-8348) Acceptance scaffold with beaker-puppet

caseywilliams:
This commit is a starting point for future work:

It adds an acceptance directory containing a set of scaffold test files and helpers to allow for testing this module with beaker-puppet instead of beaker-puppet_install_helper (which is no longer supported). The README in the acceptance directory describes how to run the new tests. Our plan is to test only FOSS functionality (not PE) in the updated acceptance tests

This leaves the existing tests in spec/acceptance intact for now; in the future these old tests will either be migrated to the acceptance directory, or (if they test puppet 3 functionality) removed.

Take note that this process will also eventually remove beaker-rspec, since it is incompatible with the use of subcommands in beaker 4.

This commit depends on the merge and release of https://github.com/puppetlabs/beaker/pull/1563 and https://github.com/puppetlabs/beaker-puppet/pull/89 -- I will update the Gemfile here to require the new tags once they are made. I'll block this PR until then.

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -1,0 +1,114 @@
+# Acceptance tests for puppetlabs-puppet_agent
+
+## Background
+
+### About Beaker
+
+Beaker is a host provisioning and an acceptance testing framework. If you are
+unfamiliar with beaker, you can start with these documents:
+
+- [The Beaker DSL document](https://github.com/puppetlabs/beaker/blob/master/docs/how_to/the_beaker_dsl.md) will help you understand the test code in the `tests/` and `pre_suite/` subdirectories.
+- [The Beaker Style Guide](https://github.com/puppetlabs/beaker/blob/master/docs/concepts/style_guide.md) will help you write new test code.
+- [Argument Processing](https://github.com/puppetlabs/beaker/blob/master/docs/concepts/argument_processing_and_precedence.md) and [Using Subcommands](https://github.com/puppetlabs/beaker/blob/master/docs/tutorials/subcommands.md) have more information on beaker's command line and environmental options.
+
+### About these tests
+
+This module is responsible for upgrading and downgrading puppet-agent.  Testing
+this behavior necessarily involves repeatedly installing and uninstalling
+puppet-agent. Ideally, the test hosts would be totally destroyed and
+reprovisioned before each fresh install of puppet-agent, but beaker does not
+support workflows like this. Instead:
+
+- Use the `run_setup` helper method to install puppet-agent, plus this module
+  and its dependencies, before the main content of each of your tests.
+- Use the `run_teardown` helper method to uninstall puppet-agent and remove
+  modules in your tests' teardown steps.
+
+See [helpers.rb](./helpers.rb) for the impelementations of these methods.
+
+#### Environment variables affecting test behavior
+
+When the `run_setup` helper installs puppet-agent, it determines the version to
+install as follows:
+
+1. If the `FROM_AGENT_VERSION` environment variable is set, install that version.
+2. Otherwise, install the latest agent from the `FROM_PUPPET_COLLECTION`
+    environment variable (set this to `pc1`, `puppet5`, or `puppet6` for puppet
+    versions 4.x, 5.x, or 6.x respectively).
+3. Otherwise, and by default, install the latest agent from the `pc1`
+    collection (this would be puppet-agent 1.x with puppet 4.x).
+
+During your tests, you may upgrade or downgrade puppet-agent to different
+versions. Unless you're testing a version-specific behavior, any tests that
+upgrade or downgrade the version of puppet-agent should use the
+`TO_AGENT_VERSION` environment variable as the target version by default.
+Similarly, use `TO_PUPPET_COLLECTION` in similar cases where behavior is based
+on the collection instead of the agent version.
+
+## How to run the tests
+
+### Install the dependencies
+
+This directory has its own Gemfile, containing gems required only for these
+acceptance tests. Ensure that you have [bundler](https://bundler.io/) installed,
+and then use it to install the dependencies:
+
+```sh
+bundle install --path .bundle
+```
+
+This will install [`beaker`](https://github.com/puppetlabs/beaker) and
+[`beaker-puppet`](https://github.com/puppetlabs/beaker-puppet) (a beaker
+library for working with puppet specifically), plus several hypervisor gems
+for working with beaker and vagrant, docker, or vsphere.
+
+### Set up the test hosts
+
+Before running any of the acceptance tests in the `tests/` directory, you must
+do the following once:
+
+- configure beaker,
+- provision VMs or containers as test hosts, and
+- run the setup tasks in the `pre_suite/` directory.
+
+Here's how:
+
+```sh
+# Use `beaker-hostgenerator` generate a hosts file that describes the
+# types of hosts you want to test. See beaker-hostgenerator's help for more
+# information on available host types and roles.
+# This example creates a Centos 7 master and a single Debian 9 agent, which will be provisioned with Docker.
+bundle exec beaker-hostgenerator -t docker centos7-64mcda-debian9-64a > ./hosts.yaml
+
+# Now run `beaker init` to generate configuration for this beaker run in
+# `.beaker/`. Pass it the location of your hosts file and the options file:
+bundle exec beaker init -h ./hosts.yaml -o options.rb
+
+# Create the VMs or containers that will act as the test hosts:
+bundle exec beaker provision
+
+# Now run the pre-suite setup tasks. This will install puppetserver and the
+# puppet_agent module on your master host in preparation for running the tests:
+bundle exec beaker exec pre-suite
+```
+
+### Run and re-run the tests
+
+Once you've set up beaker, you can run any number of tests any number of times:
+
+```sh
+# Run all the tests
+bundle exec beaker exec
+# Run all the tests in a specific directory
+bundle exec beaker exec ./tests/subdir
+# Run a commma-separated list of specific tests:
+bundle exec beaker exec ./path/to/test.rb,./another/test.rb
+```
+
+### Clean up
+
+You can destroy existing test hosts like this:
+
+```sh
+bundle exec beaker destroy
+```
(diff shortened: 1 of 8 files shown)


acceptance/helpers.rb
```diff
@@ -0,0 +1,130 @@
+require 'beaker-puppet'
```
caseywilliams: The `run_setup` and `run_teardown` helpers in this file are meant to mimic the setup and teardown functions from the original acceptance test helpers:

- `run_setup` mimics [`setup_puppet_on`](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/spec/spec_helper_acceptance.rb#L132)
	- The `use_master` keyword mimics the `agent: true` behavior from the original helper. I did not include an mcollective option, since release 2.0 of this moudule won't support puppet 3, where this was used before
- `run_teardown` mimcs [`teardown_puppet_on`](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/spec/spec_helper_acceptance.rb#L202)

I've put the minimal amount of functionality here; they will almost certainly need updates in the future. For example, I haven't included (or thoroughly tested) any instances where the `use_master` keyword is set, since I can only find two test cases (I think) in the existing test suite that use the equivalent functionality.

caseywilliams:
I may be wrong, but the travis failures look unrelated to these changes

acceptance/helpers.rb
```diff
@@ -0,0 +1,130 @@
...
+def install_modules_on(host)
+  install_dev_puppet_module_on(host, {
+    source: File.join(File.dirname(__FILE__), '..', ),
+    module_name: 'puppet_agent'
+  })
+
+  on(host, puppet('module', 'install', 'puppetlabs-stdlib',     '--version', '4.16.0'), { acceptable_exit_codes: [0] })
+  on(host, puppet('module', 'install', 'puppetlabs-inifile',    '--version', '2.1.0'),  { acceptable_exit_codes: [0] })
+  on(host, puppet('module', 'install', 'puppetlabs-apt',        '--version', '4.4.0'),  { acceptable_exit_codes: [0] })
+  on(host, puppet('module', 'install', 'puppetlabs-transition', '--version', '0.1.1'),  { acceptable_exit_codes: [0] })
```
caseywilliams: I'd rather read these module versions from ../metadata.json, but I figured I'd save that for later, since we're updating them soon anyway.

acceptance/README.md
```diff
@@ -0,0 +1,113 @@
...
+Beaker is a host provisioning and an acceptance testing framework. If you are
+unfamiliar with beaker, you can start with these documents:
+
+- [The Beaker DSL document](https://github.com/puppetlabs/beaker/blob/master/docs/how_to/the_beaker_dsl.md) will help you understand the test code in the `tests/` and `pre_suite/` subdirectories.
+- [The Beaker Style Guide](https://github.com/puppetlabs/beaker/blob/master/docs/concepts/style_guide.md) will help you write new test code.
+- [Argument Processing](https://github.com/puppetlabs/beaker/blob/master/docs/concepts/argument_processing_and_precedence.md) and [Using Subcommands](https://github.com/puppetlabs/beaker/blob/master/docs/tutorials/subcommands.md) have more information on beaker's command line and environmental options.
+
+### About these tests
+
+Testing this module necessarily involves repeatedly installing and
```
ekinanp: I'd add a note that the reason we setup/remove puppet-agent for each test is b/c we're testing stuff related to upgrades.

acceptance/helpers.rb
ekinanp: Maybe add a TODO then?

acceptance/pre_suite/00_install_agent.rb
```diff
@@ -0,0 +1,20 @@
+require 'beaker-puppet'
```
ekinanp: It might be more useful to have one file for the pre-suite, `setup_master.rb`, so that it is apparent that this setup's only for the master? The individual files are small enough that I think splitting them up into steps under a single test would still give the right amount of granularity in case something fails. As things currently are, I have to dig into these files in order to see that they apply only for the master.

acceptance/helpers.rb
```diff
@@ -0,0 +1,130 @@
...
+    configure_puppet_on(agent, agent_puppetconf)
+  end
+
+  # Either intsall modules or clear SSL/firewall, based on whether this is a masterless run
+  if use_master
+    logger.notify("Setup [master/agent test]: Clean SSL and disable firewall on all hosts")
+    # If using a master, make sure no certificate weirdness happens when the new agent contacts it:
+    hosts.each do |host|
+      on(host, "rm -rf '#{host.puppet['ssldir']}'")
+      stop_firewall_with_puppet_on(host)
```
ekinanp: Could this be done in the pre-suite? I see no code in the teardown that sets the firewall back on for these hosts.

acceptance/README.md
```diff
@@ -0,0 +1,113 @@
...
+support workflows like this. Instead:
+
+- Use the `run_setup` helper method to install puppet-agent, plus this module
+  and its dependencies, before the main content of each of your tests.
+- Use the `run_teardown` helper method to uninstall puppet-agent and remove
+  modules in your tests' teardown steps.
+
+See [helpers.rb](./helpers.rb) for the impelementations of these methods.
+
+#### Environment variables affecting test behavior
```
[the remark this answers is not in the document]
caseywilliams: The list just below this heading explains how the version is selected. If X_AGENT_VERSION is found, it will use the specific version. If that's not set, it will use X_PUPPET_COLLECTION. If that's not set, it will use the PC1 collection.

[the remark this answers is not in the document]
caseywilliams: Oh I think I see what you're saying. If both a version and a collection are supplied, the collection will be ignored. It doesn't make sense to have both control the installation.

acceptance/pre_suite/00_install_agent.rb
caseywilliams: I've merged these into a single file

acceptance/helpers.rb
caseywilliams: I guess? This variant of the method (with use_master set to true) hasn't been exercised yet and I'm not even sure it would work in its current state.

The stop firewall method requires puppet to be installed first -- if I did that in the pre-suite, I'd need to install puppet-agent on the agents in the pre-suite in addition to on the master, which doesn't make much sense to me. I could write a beaker-puppet helper that didn't rely on puppet to stop firewalls, but that seems like a lot of heavy lifting for this initial chunk of work.

acceptance/README.md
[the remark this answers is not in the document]
ekinanp: To clarify, you're saying that if I supply `FROM_AGENT_VERSION` and `FROM_PUPPET_COLLECTION`, then `FROM_AGENT_VERSION` will be used, right?

What about if I supply a version and a collection (in different directions, so e.g. `FROM_AGENT_VERSION` and `TO_PUPPET_COLLECTION`) [unless that's what you were referring to]

acceptance/helpers.rb
ekinanp: Yeah that makes sense. It would be useful to add a comment about why we're stopping the firewall there instead of in the pre-suite.

acceptance/README.md
ekinanp: Still would be useful to mention something about upgrades -- I'd prefer not having to think about why it is necessary for us to repeatedly install and uninstall the puppet-agent when it can be explicitly stated.

acceptance/tests/test_package_version_parameter.rb
```diff
@@ -0,0 +1,42 @@
+require 'beaker-puppet'
+require_relative '../helpers'
+
+# Attempts to find the previous semver-based relase of the given agent version
+def some_prevous_version(version)
```
ekinanp: Is this magic necessary? I think it'd simplify things if people were required to specify the `FROM` and `TO` environment variables.

MikaelSmith:
I thought I had something like this already, but the closest I can find is upgrading from PUPPET_CLIENT_VERSION to latest.

I don't expect to have time to look closely at this until after the new year.

caseywilliams:
I think it may be worth looking at some of the newer tools being used to test supported modules without beaker before proceeding with this PR. It's possible we'll still need beaker, but I hesitate to do much more before we've found out more about what else is available.

acceptance/README.md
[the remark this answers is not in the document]
caseywilliams: It follows the same process to decide on the "to" agent as it does to decide on the "from" agent -- the two sets of variables don't depend on one another. Do you have suggestions about how to improve this wording, or about how the variables should behave or be named?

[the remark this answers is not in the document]
caseywilliams: I should say that that's the _desired_ behavior, though! how puppet collections behave in the "to" context will have to be managed by each test individually, or by other, new helpers, later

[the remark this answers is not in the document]
caseywilliams: For example, in a test of the `package_version` parameter, it wouldn't make much sense to use a puppet collection. Perhaps a test that tests the collection could use a helper from beaker-puppet to find the collection based on the TO_AGENT_VERSION if TO_PUPPET_COLLECTION wasn't set

caseywilliams:
This PR depends on:
- https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/361
- https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/360
- https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/359
- https://github.com/puppetlabs/beaker-puppet/pull/90

acceptance/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+When the `run_setup` helper installs puppet-agent, it determines the version to
+install as follows:
+
+1. If the `FROM_AGENT_VERSION` environment variable is set, install that version.
+2. Otherwise, install the latest agent from the `FROM_PUPPET_COLLECTION`
```
ekinanp: Didn't we agree to remove the `PUPPET_COLLECTION` env variables and just stick with package versions for now? (This was before Christmas). I think that was @branan's suggestion.

caseywilliams:
I'm going to open a new PR with updated changes; this one is pretty stale.

PR #356, opened 2018-12-20
(MODULES-8317) Update module dependencies

caseywilliams:
Updates dependency modules as follows:

- Updates puppetlabs-inifile to 2.4.0+
- Updates puppetlabs-stdlib to 5.1.0+
- Updates puppetalbs-apt to 6.0.0+
- Removes puppetlabs-transition, since its use was mainly in
  installation or removal of puppet-agent before the AIO packaging updates
  in puppet 4; puppet < 4 will not be supported in release 2.0 of this
  module.


speedofdark:
Overall, LGTM. I saw your comment about OSX related fails in HipChat, so those are cool; any commentary on the Solaris fails in Travis? Thanks!

caseywilliams:
Thanks @speedofdark, I _did_ need to update some spec tests - I think everything should be in order now

PR #359, opened 2019-01-11
(MODULES-8446) Improve error messages for PE-only platforms

caseywilliams:
Previously, macOS upgrades (which are supported only under PE) would
fail silently outside of PE. This adds an error message for that case
and updates other PE-only platforms so that they have consistent error
messages for FOSS installations.

### DIFF
--- a/spec/classes/puppet_agent_prepare_spec.rb
+++ b/spec/classes/puppet_agent_prepare_spec.rb
@@ -7,20 +7,10 @@ MCO_LOGFILE = '/var/log/puppetlabs/mcollective.log'
 
 describe 'puppet_agent::prepare' do
   context 'supported operating system families' do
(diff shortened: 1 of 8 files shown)


caseywilliams:
I've updated this to standardize errors and messages across all pe-only platforms

spec/classes/puppet_agent_prepare_spec.rb
```diff
@@ -7,20 +7,10 @@
 
 describe 'puppet_agent::prepare' do
   context 'supported operating system families' do
-    ['Debian', 'RedHat', 'SuSE'].each do |osfamily|
```
caseywilliams: This test fails for all PE-only platforms. It seems like, considering that and what it's actually testing, it wouldn't hurt to remove SLES here.

ScottGarman:
:+1:

PR #360, opened 2019-01-11
(MODULES-8445) Fix collection determination outside PE

caseywilliams:
By default, this module installs the most recent version of puppet-agent
that's compatible with the master's version, using puppet collections to
do so.

Previously, the name of the default puppet collection was determined
based on the installed version of PE, which was technically correct for
PE installations, but did not help outside of PE. For open source users,
it incorrectly selected 'PC1' as the collection name regardless of which
version of the agent was installed on the master.

This derives the collection name from the aio_agent_version fact on the
master instead, as follows:

- Agents in the 1.y.z series (containing puppet 4) still get `PC1`
- Agents in the 5.y.z series (containing puppet 5) now get `puppet5`
- Agents in the 6.y.z series (containing puppet 6) now get `puppet6`
- Later agents will get 'puppet', which always defaults to the latest release (this would be puppet6 right now)
- The default, absent any aio_agent_version information, is still to install from `PC1`

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -112,17 +112,26 @@ class puppet_agent::params {
     true    => pe_build_version(),
     default => undef
   }
-  # Not PE or pe_version < 2018.1.3, use PC1
-  if ($_pe_version == undef or versioncmp("${_pe_version}", '2018.1.3') < 0) {
+
+  # The default behavior is to install the most recent version of the package
+  # that's compatible with the agent on the master. Find the puppet collection
+  # that matches it:
+  if versioncmp("${::aio_agent_version}", '5.0.0') < 0 {
+    # PE versions before 2018.1.3 use the Puppet 4 series of puppet agent
+    # releases. Outside of PE, we refer to this series of releases as 'PC1'. Note
+    # that although they include Puppet 4, the puppet-agent packages in this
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -112,17 +112,25 @@
...
   }
-  # Not PE or pe_version < 2018.1.3, use PC1
-  if ($_pe_version == undef or versioncmp("${_pe_version}", '2018.1.3') < 0) {
+
+  # PE versions before 2018.1.3 use the Puppet 4 series of puppet agent
+  # releases. Outside of PE, we refer to this series of releases as 'PC1'. Note
+  # that although they include Puppet 4, the puppet-agent packages in this
+  # series are numbered '1.y.z' (later series _do_ match the version of puppet
+  # to that of the puppet-agent package, though).
+  if versioncmp("${::aio_agent_version}", '5.0.0') < 0 {
```
ekinanp: Shouldn't this be based on the version we're upgrading to, not the one that's installed?

caseywilliams: The default behavior here is to install the most recent package that's compatible with the master's agent version, so it does make sense to check the existing version. I've added a comment about that.

caseywilliams:
This doesn't work -- I'm closing it.

PR #361, opened 2019-01-11
(MODULES-8443) Don't restart mco for FOSS puppet6 upgrades

caseywilliams:
When attempting to upgrade to or within the puppet 6 series in an open
source context (that is, only outside of PE), the module was still
attempting to restart the mcollective servicte after the upgrade -- but
mcollective was removed in puppet 6.

This updates the agent class params so that they find an accurate list
of the services to restart under FOSS.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -40,6 +40,7 @@ class puppet_agent::install(
     # The following are expected to be available in the solaris_install.sh.erb template:
     $adminfile = '/opt/puppetlabs/packages/solaris-noask'
     $sourcefile = "/opt/puppetlabs/packages/${_unzipped_package_name}"
+    $service_names = $puppet_agent::service_names
 
     # Puppet prior to 5.0 would not use a separate process contract when forking from the Puppet
     # service. That resulted in service-initiated upgrades failing because trying to remove or
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -8,6 +8,18 @@ class puppet_agent::params {
     fail('The puppet_agent class requires stringify_facts to be disabled')
   }
 
+  # Which services should be started after the upgrade process?
+  if ($::osfamily == 'Solaris' and $::operatingsystemmajrelease == '11') {
+    # Solaris 11 is a special case; it uses a custom script.
+    $service_names = []
+  } elsif (versioncmp("${::clientversion}", '5.99.0') < 0) { # (5.99.z indicates pre-release puppet6)
+    # Earlier versions Puppet (4 and 5) rely on mcollective:
+    $service_names = ['puppet', 'mcollective']
+  } else {
+    # Puppet 6+ only needs to manage the puppet service:
+    $service_names = ['puppet']
+  }
+
   # The `is_pe` fact currently works by echoing out the puppet version
   # and greping for "puppet enterprise". With Puppet 4 and PE 2015.2, there
   # is no longer a "PE Puppet", and so that fact will no longer work.
@@ -32,13 +44,6 @@ class puppet_agent::params {
 
   case $::osfamily {
     'RedHat', 'Debian', 'Suse', 'Solaris', 'Darwin', 'AIX': {
-      if !($::osfamily == 'Solaris' and $::operatingsystemmajrelease == '11') {
-        $service_names = ['puppet', 'mcollective']
-      }
-      else {
-        $service_names = []
-      }
-
       $local_puppet_dir = '/opt/puppetlabs'
       $local_packages_dir = "${local_puppet_dir}/packages"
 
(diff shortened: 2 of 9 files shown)


manifests/windows/install.pp
```diff
@@ -10,6 +10,7 @@
   $install_dir           = undef,
   $install_options       = [],
   $msi_move_locked_files = $::puppet_agent::msi_move_locked_files,
+  $service_names         = $::puppet_agent::params::service_names,
```
ekinanp: May be worth adding a comment about why we're not using the `service_names` parameter supplied by the user?

manifests/install.pp
```diff
@@ -40,6 +40,7 @@
     # The following are expected to be available in the solaris_install.sh.erb template:
     $adminfile = '/opt/puppetlabs/packages/solaris-noask'
     $sourcefile = "/opt/puppetlabs/packages/${_unzipped_package_name}"
+    $service_names = $puppet_agent::service_names
```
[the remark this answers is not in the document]
caseywilliams: Like `$adminfile` and `$sourcefile`, this variable is used in the erb template mentioned in the comment above -- If there's a more direct way to get those variables into the template's scope, I'm interested to know about it.

manifests/params.pp
```diff
@@ -8,6 +8,18 @@
     fail('The puppet_agent class requires stringify_facts to be disabled')
   }
 
+  # Which services should be started after the upgrade process?
```
caseywilliams: @ekinanp were your objections here about the comment or the behavior? If it was the comment, I've updated it. If it was behavior, I'm not sure I see how this is changing the behavior from how it used to be (outside of removing mcollective for puppet6), but this is the first time I've made changes to a module that are more than a few lines -- I don't know what I'm missing.

manifests/install.pp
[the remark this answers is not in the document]
caseywilliams: Like `$adminfile` and `$sourcefile`, this variable is used in the erb template mentioned in the comment above -- If there's a more direct way to get those variables into the template's scope, I'm interested to know about it.

manifests/windows/install.pp
```diff
@@ -56,7 +57,7 @@
     default => "${::system32}\\cmd.exe"
   }
 
-  if (member($::puppet_agent::service_names, 'puppet')) {
+  if (member($::puppet_agent::params::service_names, 'puppet')) {
```
caseywilliams: Is this adequate? Given the brittleness of testing in this module, I'm hesitant to make any too-large structural changes, so I'm hoping to do the minimum to fix the bug right now.

manifests/install.pp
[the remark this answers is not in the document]
ekinanp: Oh, you're right. Sorry about that.

manifests/params.pp
ekinanp: Just the comment

manifests/windows/install.pp
ekinanp: You want `$::puppet_agent::service_names` there as-is b/c it's what the user passes in, but you'd want to update [install_puppet.bat.erb](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/templates/install_puppet.bat.erb#L42-L44) so that we don't stop mcollective on a Puppet 6 node (before the upgrade).

manifests/service.pp
```diff
@@ -12,21 +12,22 @@
       ((versioncmp("${::clientversion}", '4.0.0') < 0) or $puppet_agent::aio_upgrade_required) {
     # Only use script if we just performed an upgrade.
     $_logfile = "${::env_temp_variable}/solaris_start_puppet.log"
+    # We'll need to pass the names of the services to start to the script
+    $_service_names_arg = join($::puppet_agent::service_names, ' ')
```
ekinanp: Since the original script hard-coded the services, it may be safer to make this `$::puppet_agent::params::service_names` to avoid breaking things.

manifests/service.pp
```diff
@@ -12,21 +12,22 @@
...
     notice ("Puppet service start log file at ${_logfile}")
     file { "${::env_temp_variable}/solaris_start_puppet.sh":
       ensure => file,
       source => 'puppet:///modules/puppet_agent/solaris_start_puppet.sh',
       mode   => '0755',
     }
     -> exec { 'solaris_start_puppet.sh':
-      command => "${::env_temp_variable}/solaris_start_puppet.sh ${::puppet_agent_pid} 2>&1 > ${_logfile} &",
+      command => "${::env_temp_variable}/solaris_start_puppet.sh ${::puppet_agent_pid} ${_service_names_arg}2>&1 > ${_logfile} &",
```
ekinanp: Needs a space between `${_service_names_arg}` and `2>&1`.

manifests/service.pp
caseywilliams: That makes it sound a lot like the `service_names` param has never worked here -- I will test things out, but I think a 2.0 release would be a good opportunity to fix this if it's broken

caseywilliams: I tested this out today and it seems to work --  I was able to control the services that started after the upgrade with service_names

PR #362, opened 2019-01-11
(MODULES-8431) Update windows installation to use powershell

McdonaldSeanp:
This commit updates the installation method on windows to use a powershell
script rather than a batch file


branan:
I think you forgot to `git rm` the batch file

McdonaldSeanp:
whoops

manifests/windows/install.pp
```diff
@@ -62,24 +56,33 @@
...
+
+  file { "${_installps1}":
     ensure  => file,
-    content => template('puppet_agent/install_puppet.bat.erb')
+    content => template('puppet_agent/install_puppet.ps1.erb')
   }
-  -> exec { 'install_puppet.bat':
-    command => "${::system32}\\cmd.exe /c start /b ${_cmd_location} /c \"${_installbat}\" ${::puppet_agent_pid}",
+  -> exec { 'install_puppet.ps1':
+    command => "${::system32}\\cmd.exe /c start /b ${::system32}\\WindowsPowerShell\\v1.0\\powershell.exe -ExecutionPolicy Bypass -NoProfile -NoLogo -NonInteractive -Command ${_installps1} ${::puppet_agent_pid}",
```
glennsarti: Shouldn't this be `-File` not command?

spec/classes/puppet_agent_windows_install_spec.rb
```diff
@@ -44,26 +44,14 @@
 
         let(:facts) { facts.merge({:is_pe => true}) }
 
-        if Puppet.version < '4.0.0'
```
glennsarti: Not sure why this entire test is removed.  If it's a Puppet 3 thing, might be best to add this as a separate commit

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+.Parameter PuppetStartType
+  Optionally change the default start type of puppet
+.Parameter InstallArgs
+  Provide any extra argmuments to the MSI installation
+.Parameter UseLockedFilesWorkaround
+  Set to $true to enable execution of the puppetres.dll move workaround. See https://tickets.puppetlabs.com/browse/MODULES-4207
+#>
+[CmdletBinding()]
+param(
+  # PuppetPID _must_ come first!, this script needs PuppetPID to be a positional parameter to execute
```
glennsarti: You can force position using advanced functions - https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-6#position-argument

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+  # correctly from the module.
+  [String] $PuppetPID,
+  [String] $Source='<%= @_msi_location %>',
+  [String] $Logfile='<%= @_logfile %>',
+  [String] $InstallDir='<%= @install_dir %>',
+  [String] $PuppetMaster='<%= @_puppet_master %>',
+  [String] $PuppetStartType='<%= @_agent_startup_mode %>',
+  [String] $InstallArgs='<%= @_install_options.map do |option|
```
glennsarti: Personally I'd prefer this "logic" be in the manifest and the $InstallArgs be a du,b string, but I'm not too fussed

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+# fetch_install_dir, move_puppetres_dll and restore_puppetres_dll serve as a workaround for older
+# installations of puppet: we used to need to move puppetres.dll out of the way during puppet
+# upgrades because the file would lock and cause network stack restarts.
+# See https://tickets.puppetlabs.com/browse/MODULES-4207
+<#
+.Synopsis
+  Fetch the location of the puppet installation from the registry
+#>
+function Script:fetch_install_dir {
+  param()
```
glennsarti: Without cmdlet binding, this is not needed.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+    }
+  }
+}
+
+<#
+.Synopsis
+  Move/rename puppetres.dll to a temporary location
+#>
+function Script:move_puppetres_dll {
+  param()
```
glennsarti: Without cmdlet binding, this is not needed.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+    $rand_string = [String[]](Get-Random)
+    $temp_puppetres = "$env:temp\puppetres$rand_string.dll"
+    # _Never_ use the $InstallDir top-level parameter to try and find the InstallDir
+    # for the puppetres workaround. The workaround should _always_ fetch the InstallDir
+    # from the registry. This is so a user can specify a different InstallDir than the
+    # directory where the current package is installed and the workaround will still work
+    # for the already installed package
+    $InstallDir = fetch_install_dir
+    if (Test-Path "$InstallDir\puppet\bin\puppetres.dll") {
+      Out-File -InputObject "Moving puppetres.dll to $temp_puppetres" -FilePath $Logfile -Append
```
glennsarti: The canonical way of doing this is normally;

`"Moving puppetres.dll to $temp_puppetres"  | Out-File -FilePath $Logfile -Append"`

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+  begin {
+    # _Never_ use the $InstallDir top-level parameter to try and find the InstallDir
+    # for the puppetres workaround. The workaround should _always_ fetch the InstallDir
+    # from the registry. This is so a user can specify a different InstallDir than the
+    # directory where the current package is installed and the workaround will still work
+    # for the already installed package
+    $InstallDir = fetch_install_dir
+    if (Test-Path $temp_puppetres) {
+      Out-File -InputObject "Restoring puppetres.dll" -FilePath $Logfile -Append
+      Move-Item -Path $temp_puppetres -Destination "$InstallDir\puppet\bin\puppetres.dll"
```
glennsarti: This should fail gracefully.  Moving the locked DLL will fail on rollback

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+    if ($serv_exists) {
+      Out-File -InputObject "Stopping $($service) before upgrade" -FilePath $Logfile -Append
+      Stop-Service $service
+    }
+  }
+  if ($UseLockedFilesWorkaround) {
+    $temp_puppetres = move_puppetres_dll
+  }
+  Out-File -InputObject "Beginning MSI installation with Arguments: $msi_arguments" -FilePath $Logfile -Append
+  Out-File -InputObject "****************************** Begin msiexec.exe output ******************************" -FilePath $Logfile -Append
```
glennsarti: Something to watch out for here is that if the user specifies a service password, it may appear in this log in clear text

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+  $msiexec_proc = Start-Process -FilePath msiexec.exe -PassThru -ArgumentList $msi_arguments
+  # Using the -Wait option in Start-Process will wait on the process group instead of just the one msiexec process in older versions of windows.
+  # but using WaitForExit works fine.
+  $msiexec_proc.WaitForExit()
+  Out-File -InputObject "****************************** End msiexec.exe output ******************************" -FilePath $Logfile -Append
+  if ($msiexec_proc.ExitCode -ne 0){
```
glennsarti: Not quite true.

Exit code could be 0, 3010 or 1604 (? I think it's 1604 - Soft reboot required)

ekinanp:
@glennsarti I was looking into https://tickets.puppetlabs.com/browse/PA-663 to understand the `puppetres.dll` issue and it looks like your comment on 2017/03/01 seems to imply that the workaround is only necessary until https://tickets.puppetlabs.com/browse/PA-768 is resolved?

spec/classes/puppet_agent_windows_install_spec.rb
ekinanp: Our team's decided to remove Puppet 3 tests as a part of releasing the puppet-agent module 2.0. See https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/350

templates/install_puppet.ps1.erb
ekinanp: I agree with Glenn

templates/install_puppet.ps1.erb
ekinanp: May be easier to abstract this into a method? `Log <error_message>` is more readable.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+}
+if ($PuppetStartType) {
+  $msi_arguments += " PUPPET_AGENT_STARTUP_MODE=`"$PuppetStartType`""
+}
+Out-File -InputObject $PID -FilePath $InstallScriptPIDFile
+try {
+  # Wait for the puppet run to finish
+  Out-File -InputObject "Waiting for puppet to stop, PID:$PuppetPID" -FilePath $Logfile -Append
+  try {
+    Wait-Process -Id $PuppetPID -Timeout 120
```
ekinanp: You could also do
```
$proc = Get-Process -Id $PuppetPID -ErrorAction SilentlyContinue
if ($proc) {
  if (-Not $proc.WaitForExit(120000)) { // milliseconds
    // Fail
  }
}
```

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+      Out-File -InputObject "Puppet was already finished, continuing..." -FilePath $Logfile -Append
+    } elseif ($_.Exception.GetType() -eq [System.TimeoutException]) {
+      Out-File -InputObject "ERROR: timed out waiting for puppet run to finish!" -FilePath $Logfile -Append
+      throw $_
+    } else {
+      Out-File -InputObject "ERROR: failed while trying to wait for puppet process with $($_.Exception.GetType())" -FilePath $Logfile -Append
+      throw $_
+    }
+  }
+  foreach($service in $service_names) {
```
ekinanp: The previous `install_puppet.bat` file had a long comment about why stopping the services was important. Maybe that comment should be here too?

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+}
+
+# ************** Execution start **************
+$ErrorActionPreference = "Stop"
+$service_names=@(
+  "puppet",
+  "pxp-agent",
+  "mcollective"
+)
+$msi_arguments = "/qn /norestart /i `"$Source`" /l*vx+ `"$Logfile`" $InstallArgs"
```
ekinanp: I think this should move closer to where the installation happens, as it's easy to forget what these arguments look like later down the code.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+
+if (Test-Path "HKLM:\SOFTWARE\Puppet Labs") {
+  $msi_arguments += " ENVIRONMENT=`"$(puppet.bat config print --section agent environment)`""
```
ekinanp: Shouldn't this be `PUPPET_AGENT_ENVIRONMENT`?

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,185 @@
...
+  }
+  if ($UseLockedFilesWorkaround) {
+    $temp_puppetres = move_puppetres_dll
+  }
+  Out-File -InputObject "Beginning MSI installation with Arguments: $msi_arguments" -FilePath $Logfile -Append
+  Out-File -InputObject "****************************** Begin msiexec.exe output ******************************" -FilePath $Logfile -Append
+  $msiexec_proc = Start-Process -FilePath msiexec.exe -PassThru -ArgumentList $msi_arguments
+  # Using the -Wait option in Start-Process will wait on the process group instead of just the one msiexec process in older versions of windows.
+  # but using WaitForExit works fine.
+  $msiexec_proc.WaitForExit()
```
ekinanp: Would it be useful to log the MSI's exit code?

templates/install_puppet.ps1.erb
ekinanp: I think the `InstallArgs` should be appended after everything else like it is in the `install_puppet.bat` file. This would ensure that we don't hit some weird "arguments given out of order" thing for people who pass in their own installation options. Maybe it isn't an issue here, but better safe than sorry.

templates/install_puppet.ps1.erb
ekinanp: And also waiting on a timeout?

glennsarti:
@ekinanp 
> seems to imply that the workaround is only necessary until https://tickets.puppetlabs.com/browse/PA-768 is resolved?

Not quite. The workaround is required to use this module to upgrade older agents that DON'T have the MSI workaround in them...  e.g. 4.7 -> 5.4

templates/install_puppet.ps1.erb
glennsarti: Get's tricky here, what value would you choose for a timeout?

What do you do IF it times out? Just wait longer?

TBH I don't think this requires one.

templates/install_puppet.ps1.erb
glennsarti: I would be very careful using un-sanitised input here too (e.g. trailing line endings)

`puppet.bat config print --section agent environment`

+1 to `PUPPET_AGENT_ENVIRONMENT` 
https://puppet.com/docs/puppet/5.4/install_windows.html#puppetagentenvironment

templates/install_puppet.ps1.erb
glennsarti: +1 to refactor to it's own function

templates/install_puppet.ps1.erb
glennsarti: Or more simply;

```
try {
  Get-Process -ID $PuppetPID -ErrorAction SilentlyContinue | Wait-Process -Timeout 120`
}
catch [System.TimeoutException] {
 # TIMEOUT OCCURED
}
catch {
  # unhandled error
}
```

templates/install_puppet.ps1.erb
ekinanp: Not sure about the value, I'm not familiar with how long a typical agent install/upgrade takes on Windows. It may be useful to log something if the install's taking longer than average, such as the PID of the MSI process?

I agree that a timeout isn't strictly necessary.

glennsarti: And that's kind of the problem.  We don't have any statistics to base a number on :-(

templates/install_puppet.ps1.erb
McdonaldSeanp: It would not be an issue, but I can move it for posterity.

templates/install_puppet.ps1.erb
McdonaldSeanp: I tend to agree that we can't assume any kind of reasonable timeout here. I can guess, but we have no idea what versions of windows and how many resources the windows machine will have on installation.

templates/install_puppet.ps1.erb
McdonaldSeanp: Since we force /norestart in the MSI wouldn't the restart exit codes technically be failures as far as we're concerned?

It's my understanding that the agent should absolutely never cause a restart on upgrade.

templates/install_puppet.ps1.erb
McdonaldSeanp: Would the password appear because of the MSI output? or is there something in the script that would show that?

I was a little worried about that, but the current batch file was writing this log output. I can go either way with this, I was following the exact implementation of the batch file as much as I could.

templates/install_puppet.ps1.erb
McdonaldSeanp: I can also test for the existence of the normal puppetres.dll too.

templates/install_puppet.ps1.erb
glennsarti: I believe the MSI logs it.  I mean we can punt the issue till later, but it shouldn't be forgotten

templates/install_puppet.ps1.erb
glennsarti: > Those are both successful installation methods. 1641 simply means "The requested operation completed successfully. The system will be restarted so the changes can take effect" and 3010 means "The requested operation is successful. Changes will not be effective until the system is rebooted."
>
> Thus one is initiating the reboot automatically, and the other is waiting for the user to reboot. The difference is in the switches or properties you've set when calling the MSI.

3010 is a succesfull install but it's pending a reboot.  i.e. the agent package WANTED to reboot but it couldn't due to /norestart

glennsarti:
Pending manual Functional Review of the DLL locking repro case and description of manual testing performed by @McdonaldSeanp

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,261 @@
...
+  "pxp-agent",
+  "mcollective"
+)
+try {
+  Write-Log "Installation PID:$PID"
+  $install_pid_lock = Join-Path -Path (Split-Path -Parent (puppet.bat config print agent_catalog_run_lockfile)) -ChildPath 'puppet_install.pid'
+  Lock-Installation $install_pid_lock
+  # Wait for the puppet run to finish
+  #
+  # We *must* wait for Puppet Agent to finished applying its prior catalog
```
ekinanp: Nit: Typo here -- "to finished applying".

Also, this comment should be expanded on a bit more. Probably something like

> We *must* wait for the puppet agent to finish applying its catalog before we stop all of our services. Otherwise if the catalog has additional resources that manage our services (e.g. such as those from the PE module), then the install will fail to proceed.

i.e. it isn't something specific to pxp-agent, it applies to all services.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,261 @@
...
+.Parameter message
+  String containing the message to write
+#>
+function Script:Write-Log {
+  param(
+    [Parameter(Mandatory=$true)]
+    [String] $message
+  )
+  begin {
+    $message | Out-File -FilePath $Logfile -Append
```
ekinanp: May be worth including a timestamp.

McdonaldSeanp:
Manual testing description: https://tickets.puppetlabs.com/browse/MODULES-8431?focusedCommentId=619991&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-619991

glennsarti:
Pic

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,262 @@
...
+.Parameter UseLockedFilesWorkaround
+  Set to $true to enable execution of the puppetres.dll move workaround. See https://tickets.puppetlabs.com/browse/MODULES-4207
+#>
+[CmdletBinding()]
+param(
+  # PuppetPID _must_ come first!, this script needs PuppetPID to be a positional parameter to execute
+  # correctly from the module.
+  [parameter(Position=0)]
+  [String] $PuppetPID,
+  [String] $Source='<%= @_msi_location %>',
```
Iristyle: I'm wondering if we should be passing params via the invocation, rather than injecting values into a template. I think the only reason we used a template on the batch side was because arg passing with batch can be a real pain.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,262 @@
...
+  Provide any extra argmuments to the MSI installation
+.Parameter UseLockedFilesWorkaround
+  Set to $true to enable execution of the puppetres.dll move workaround. See https://tickets.puppetlabs.com/browse/MODULES-4207
+#>
+[CmdletBinding()]
+param(
+  # PuppetPID _must_ come first!, this script needs PuppetPID to be a positional parameter to execute
+  # correctly from the module.
+  [parameter(Position=0)]
+  [String] $PuppetPID,
```
Iristyle: This should be a numeric value, shouldn't it?

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,262 @@
...
+# Find-InstallDir, Move-PuppetresDLL and Reset-PuppetresDLL serve as a workaround for older
+# installations of puppet: we used to need to move puppetres.dll out of the way during puppet
+# upgrades because the file would lock and cause network stack restarts.
+# See https://tickets.puppetlabs.com/browse/MODULES-4207
+<#
+.Synopsis
+  Fetch the location of the puppet installation from the registry
+#>
+function Script:Find-InstallDir {
+  begin {
```
Iristyle: Since this function doesn't accept pipeline input, `begin`, `process`, `end` are superfluous.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,262 @@
+# install_puppet.ps1
+<#
```
Iristyle: On a similar note about this being a template... these inline docs won't inform the PowerShell runtime during test / debugging since it's not a PS1 until rendered.

templates/install_puppet.ps1.erb
```diff
@@ -0,0 +1,262 @@
...
+    Write-Log "ERROR: msiexec.exe installation failed!!! Return code $($msi_exec.ExitCode)"
+    throw
+  }
+} catch {
+  Write-Log "ERROR: $_"
+  if ($UseLockedFilesWorkaround) {
+    Reset-PuppetresDLL $temp_puppetres
+  }
+} finally {
+  Unlock-Installation $install_pid_lock
```
Iristyle: I think you also want to call `Unregister-Event` to perform cleanup... though exiting PowerShell altogether may do this for you

PR #364, opened 2019-01-14
(MODULES-4730) Do not pass environment on upgrade

reidmv:
Modern agent installs should not rely on environment being specified in
puppet.conf, preferring instead that environment assignment is defined
centrally. In the event puppet.conf does not already have an
environment= line, it is important NOT to add one. Passing
PUPPET_AGENT_ENVIRONMENT to the MSI will have the effect of adding this
line to puppet.conf. Therefore, only do it if an environment line
already exists.


ScottGarman:
I took a different approach in https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/378 - my understanding is it's not even necessary to pass the environment during the MSI install at all, that this is taken care of automatically. Or am I missing something?

McdonaldSeanp:
This is superseded by https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/378

PR #366, opened 2019-01-16
(MODULES-8432) refresh PA repo if the version is not in the local cache

mihaibuzgau:
This PR adds the functionality to refresh the local cache of the Suse package repository if the desired version of the Puppet Agent is not found.

### DIFF
--- a/manifests/osfamily/suse.pp
+++ b/manifests/osfamily/suse.pp
@@ -75,6 +75,7 @@ class puppet_agent::osfamily::suse(
 
         $repo_file = '/etc/zypp/repos.d/pc_repo.repo'
         $repo_name = 'pc_repo'
+        $agent_version = $::puppet_agent::package_version
 
         # In Puppet Enterprise, agent packages are served by the same server
         # as the master, which can be using either a self signed CA, or an external CA.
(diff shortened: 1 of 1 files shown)


manifests/osfamily/suse.pp
```diff
@@ -96,6 +97,13 @@
             value   => $value,
           }
         }
+
+        exec { "refresh-${repo_name}":
```
ekinanp: I think the [`ini_setting` resource should have a `before => Exec["refresh-${repo_name}"]`](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/osfamily/suse.pp#L91-L97) to make sure that the repo's created prior to refreshing it.

speedofdark: Could you please point me at a reference to `before`? I can only find it in the context of an `insert_type`, and not as a standalone element. Thanks.

ekinanp: https://puppet.com/docs/puppet/5.3/lang_relationships.html#syntax-relationship-metaparameters

mihaibuzgau:
@ekinanp @speedofdark can I merge this ?

PR #369, opened 2019-01-17
(maint) make pid file name consistent on windows

McdonaldSeanp:


highb:
I should probably wait until Travis is done...

PR #371, opened 2019-01-23
(MODULES-8348) Acceptance scaffold with beaker-puppet

caseywilliams:
Adds an acceptance directory containing a set of scaffold test files
and helpers to allow for testing this module with beaker-puppet instead
of beaker-puppet_install_helper (which is no longer supported). The
README in the acceptance directory describes how to run the new tests.

This leaves the existing tests in spec/acceptance intact for now; in the
future these old tests will either be migrated to the acceptance
directory, or (if they test puppet 3 functionality) removed.

Take note that this process will also eventually remove beaker-rspec,
since it is incompatible with the use of subcommands in beaker 4.

Do not merge until https://github.com/puppetlabs/beaker-puppet/pull/97 has been merged and released. I will need to update the beaker version in the Gemfile once that happens to require a new enough beaker release.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -1,0 +1,60 @@
+def beaker(command)
+  debug_flag = ''
+  if ENV['BEAKER_debug'] || ENV['DEBUG']
+    debug_flag = '--debug'
+  end
+
+  sh(%(beaker #{command} #{debug_flag}))
+end
+
+desc 'Prepare for running tests: Run beaker init, provision, and pre-suite all at once.'
+task :prepare do
+  errors = []
+
+  if !ENV['MASTER_PACKAGE_VERSION'] && !ENV['MASTER_COLLECTION']
+    errors << 'You must set a starting version of puppet-agent using ' +
+              '$MASTER_PACKAGE_VERSION or $MASTER_COLLECTION'
+  end
+
+  if ENV['HOSTS']
+    hosts_file = File.expand_path(ENV['HOSTS'])
+    errors << "Couldn't find $HOSTS file at #{hosts_file}" unless File.exist?(hosts_file)
+  elsif File.exist?('hosts.yml')
+    hosts_file = './hosts.yml'
+  elsif File.exist?('hosts.yaml')
+    hosts_file = './hosts.yaml'
+  else
+    errors << 'Unable to find a hosts file in $HOSTS or at ./hosts.yml or at ./hosts.yaml'
+  end
+
+  unless errors.empty?
+    raise errors.join("\n")
+  end
+
+  puts "Using beaker hosts file #{hosts_file}"
+
+  # Log the contents of the host file if running in Jenkins
+  if ENV['JENKINS_HOME']
+    pp File.read(hosts_file)
+  end
+
+  beaker("init -h #{hosts_file} -o options.rb")
+  beaker("provision")
+  beaker("exec ./pre_suite")
+
+  # Don't print these human instructions if running in Jenkins
+  unless ENV['JENKINS_HOME']
+    puts 'You can run individual test(s) with `bundle exec beaker exec <path-to-test(s)>`'
+    puts "You can destroy your provisioned hosts with `bundle exec beaker destroy` when you're ready"
+  end
+end
+
+desc 'Run all the tests and destroy the hosts afterward'
+task :ci do
+  begin
+    Rake.application['prepare'].invoke
+    beaker('exec ./tests')
+  ensure
+    beaker('destroy')
+  end
+end
--- a/acceptance/helpers.rb
+++ b/acceptance/helpers.rb
@@ -1,0 +1,231 @@
+require 'beaker-puppet'
+
+module Beaker
+  module DSL
+    # Host selectors by role
+    module Roles
+      # Select any hosts which have the agent role and are not the master.
+      # Beaker's `agents` selector selects all of the hosts that have the
+      # `agent` role, but some masters may have _both_ the agent and master
+      # roles. This `agents_only` selector will not include any agent hosts that
+      # have the `master` role.
+      #
+      # @return [Array<Beaker::Host>] A set of beaker hosts which have the `agent` role but not the `master` role
+      def agents_only
+        hosts_as(:agent).reject { |host| host['roles'].include?('master') }.to_a
+      end
+    end
+
+    # Helpers for testing the puppetlabs-puppet_agent module with Beaker
+    module Helpers
+      # Purging puppet-agent between the tests requires a helper script for some
+      # platforms (for example, on Windows). This directory holds those scripts.
+      SUPPORTING_FILES = File.expand_path('./files').freeze
+
+      # These platforms are only supported under Puppet Enterprise, and FOSS
+      # tests should be skipped on them.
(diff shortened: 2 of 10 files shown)


acceptance/Rakefile
```diff
@@ -0,0 +1,58 @@
...
+  puts "Using beaker hosts file #{hosts_file}"
+
+  # Log the contents of the host file if running in Jenkins
+  if ENV['JENKINS_HOME']
+    pp File.read(hosts_file)
+  end
+
+  beaker("init -h #{hosts_file} -o options.rb")
+  beaker("provision")
+  beaker("exec ./pre_suite")
```
speedofdark: I found during my work adding only-fails to beaker(-puppet) that if one of the beaker subcommands fails, it will kill the rake run right then and there, so no subsequent beaker subcommands (or indeed anything else in the Rakefile in question) will be executed. This means that provisioned machines will wait for their timeout before they are reaped.

I addressed this by wrapping things with a 'begin/rescue/ensure' block, as illustrated here: https://github.com/puppetlabs/beaker-puppet/blob/1d269589adbb9c47888e47797e63c749133a58a7/tasks/ci.rake#L275-L297

caseywilliams: I'm not sure I would want that to happen as a _developer_ using these tests -- I would want to look at the SUTs and see what happened and then manually destroy them -- but I'll add that to the CI task now.

acceptance/helpers.rb
```diff
@@ -0,0 +1,230 @@
...
+      # These platforms are only supported under Puppet Enterprise, and FOSS
+      # tests should be skipped on them.
+      PE_ONLY_PLATFORMS = %w[aix amazon sles solaris osx]
+
+      # Use `puppet module install` to install the puppet_agent module's
+      # dependencies on the target host, and then install the puppet_agent
+      # module itself. Requires that puppet is installed in advance.
+      #
+      # @param [Beaker::Host] host The target host
+      def install_modules_on(host)
```
ekinanp: `install_puppet_agent_module_on` would be better.

acceptance/Rakefile
speedofdark: I see your point. Would we want to add the complexity of making that conditional, along the line of the -`-preserve-hosts` flag?

acceptance/helpers.rb
```diff
@@ -0,0 +1,230 @@
...
+                                     module_name: 'puppet_agent')
+      end
+
+      # Install puppet-agent on all agent nodes and connect them to the master.
+      # This is intended to prepare SUTs for tests that upgrade puppet-agent
+      # from some initial version.
+      #
+      # @param [String] initial_package_version_or_collection Either a version
+      #   of puppet-agent or the name of a puppet collection to install the agent from.
+      def prepare_upgrade_with(initial_package_version_or_collection)
```
ekinanp: `prepare_upgrade_with` is too specific and should be something left to the tests instead of handled at the helper level. How about

```
setup_puppet_agent_on(hosts, package_version_or_collection, options = {})
```

and you can specify the master as an option so that the SSL cert-sign stuff can be conditionally triggered. This way, you ca re-use this helper in the master set-up stage without losing the clarity (since you're basically doing the same thing there).

acceptance/helpers.rb
```diff
@@ -0,0 +1,230 @@
...
+      # - Install the a puppet-agent package (`initial_package_version_or_collection`)
+      #   on all the agent hosts,
+      # - Put a teardown step in place so that this package is uninstalled after the test,
+      # - Run a default manifest on all the agent hosts, and
+      # - Allow for assertions inside in a block
+      #
+      # @param [String] initial_package_version_or_collection Either a version
+      #   of puppet-agent or the name of a puppet collection to install on agent hosts
+      # @param [String] upgrade_manifest A manifest to apply to all agent nodes
+      def run_foss_upgrade_with_manifest(initial_package_version_or_collection, upgrade_manifest)
```
ekinanp: I think the code here's short enough and that there are enough useful helpers already that it would be good enough to leave this code to the individual test files, especially when we'll start adding more complicated tests where the preparation's likely going to be more involved (like Windows service stuff).

acceptance/Rakefile
ekinanp: I agree that an option to destroy the hosts would be useful.

acceptance/tests/test_upgrade_puppet5_to_puppet6.rb
```diff
@@ -0,0 +1,23 @@
...
+# This test will only run if the version of puppet on the master host is less than 6.
+test_name 'puppet_agent class: collection parameter for FOSS upgrades' do
+  master_agent_version = puppet_agent_version_on(master)
+
+  unless version_is_less(master_agent_version, '6.0.0')
+    skip_test("The puppet-agent package on the master is #{master_agent_version}; Skipping puppet5 to puppet6 upgrade test")
+  end
+
+  upgrade_to = '6.0.0'
+  manifest = "class { puppet_agent: package_version => '#{upgrade_to}' }"
```
ekinanp: It might be useful to add a `puppet_agent_manifest` helper?

acceptance/Rakefile
caseywilliams: This is one of the many places where the beaker-with-subcommands vs. beaker-without-subcommands workflows diverge. `--preserve-hosts` does not make a lot of sense with subcommands unless you're running the entire test suite with one invocation, since `beaker destroy` is already available and intended to be called separately. I'm hoping this is something we can refine in subsequent PRs.

acceptance/helpers.rb
```diff
@@ -0,0 +1,230 @@
...
+
+        step 'Setup: Prepare agents for upgrade' do
+          step 'Install puppet-agent on agent hosts' do
+            initial_package_version_or_collection ||= master_agent_version
+            if initial_package_version_or_collection =~ /(^pc1$|^puppet\d+)/i
+              agent_install_options = { puppet_collection: initial_package_version_or_collection }
+            else
+              agent_install_options = {
+                  puppet_agent_version: initial_package_version_or_collection,
+                  puppet_collection: puppet_collection_for_puppet_agent_version(initial_package_version_or_collection)
```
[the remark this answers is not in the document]
caseywilliams: these are options passed into beaker-puppet's `install_puppet_agent_on` directly -- you'd have to change them there.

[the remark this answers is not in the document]
caseywilliams: oh, you meant puppet_collection_for_puppet_agent_version -- same problem, though

caseywilliams:
(I've updated with a bugfix from sean and a few of the simpler renaming comments, but the design feedback will take me longer to address)

caseywilliams:
Ok, I believe this is ready to merge, but first:
- We need to release beaker-puppet
- I need to update the minimum beaker-puppet version here to match that new version

I'll remove the do not merge label once that's done.

caseywilliams:
(ok, updated this code to use the new puppet_collection_for helper)

ekinanp:
beaker-puppet PR was merged, now just need to wait on a beaker-puppet release before this can be merged.

caseywilliams:
beaker-puppet 1.15.0 has been release with the changes needed for these tests -- I've updated the Gemfile version here removed the "don't merge" label.

PR #372, opened 2019-01-23
(MODULES-7840) Update docs with better parameter descriptions

McdonaldSeanp:


README.markdown
```diff
@@ -136,35 +114,51 @@ This will ensure the version `1.4.0` of the puppet-agent package is installed. F
...
-##### `package_name`
-
-The package to upgrade to, i.e., `puppet-agent`. Currently, the default and only accepted value is `puppet-agent`.
+``` puppet
+  manage_repo => true
+```
 
 ##### `package_version`
 
 The package version to upgrade to. When upgrading from Puppet < 4.0, defaults to the puppet master's latest supported version
```
caseywilliams: Upgrading from < 4.0 is no longer supported in master -- we can probably just remove that part of this sentence

README.markdown
```diff
@@ -136,35 +114,51 @@ This will ensure the version `1.4.0` of the puppet-agent package is installed. F
...
+The architecture version you wish to install. Defaults to `$::facts['architecture']`.
+``` puppet
+  arch => 'x86_64'
+```
 
 ##### `collection`
 
-The Puppet Collection to track. Defaults to `PC1`.
+The Puppet Collection to track, should be one of `puppet5` or `puppet6`. Puppet collections contain the latest agents included
+in the collection's series, so the latest 5 series in puppet5 (for example: 5.5.10) and the latest 6 series in puppet6 (for
```
caseywilliams: I think these specific examples may become confusing later, since they won't match the latest versions anymore -- maybe we could say, like `the latest 5.y.z` or something like that? Don't know if it actually matters that much though

PR #374, opened 2019-01-24
(MODULES-8523) Remove legacy Puppet 3 code

ekinanp:
The puppet_agent module no longer supports upgrades from Puppet 3 nodes.
This commit removes the remaining Puppet 3 code from the manifest and
any remaining spec tests.


manifests/init.pp
```diff
@@ -73,28 +67,27 @@
...
 
     # Strip git sha from dev builds
-    if ($package_version != undef and $package_version =~ /g/){
+    if $package_version =~ /g/ {
       $_expected_package_version = split($package_version, /[.-]g.*/)[0]
     } else {
       $_expected_package_version = $package_version
     }
 
     $aio_upgrade_required = (getvar('::aio_agent_version') == undef) or
```
ekinanp: I wonder if I can remove the `(getvar('::aio_agent_version') == undef)` check? That fact should be defined on Puppet 4 agents, shouldn't it?

ekinanp:
I've tested some PE upgrades on:
```
osx_1013_x86_64
debian_9_i386
solaris_10_i386
solaris_11_i386
windows 
```

and things seem to work fine. This is ready for merge.

manifests/init.pp
```diff
@@ -73,28 +67,27 @@
     validate_absolute_path($install_dir)
   }
```
ekinanp: I wonder if it'd be worth adding a check here to error if `$::aio_agent_version` is undefined (i.e. error if we're running on a pre-Puppet 4 node).

ekinanp:
Ok, now this is ready for merge.

ekinanp:
It'd be good if somebody can double-check some of the logical substitutions made in this PR, as those can be subtle.

manifests/init.pp
speedofdark: It should blow up a few lines further down if it's missing.

ekinanp: Turns out I can remove it -- checking if it is undef is a quick way to see if you're running on a pre-Puppet 4 node.

manifests/init.pp
ekinanp: My Q. about `$::aio_agent_version` is true, so my other Q. about whether it's worth erroring if that variable is undefined still stands.

speedofdark: I think so, yes. Major things depend on it being set.

manifests/init.pp
```diff
@@ -105,7 +97,6 @@
     # use the full version string for comparisons.
     if $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '11' {
       # Strip letters from development builds. Unique to Solaris 11 packaging.
-      # Need to pass the regex as strings for Puppet 3 compatibility.
```
speedofdark: So does it get us or lose us anything to keep passing as strings? Or is it a wash?

ekinanp: It would be better to pass them in as patterns, but I don't want to risk accidentally breaking something due to some mal-formed regex (e.g. like escaping the wrong thing), so I chose to keep them in as strings.

spec/classes/puppet_agent_prepare_spec.rb
```diff
@@ -14,9 +9,7 @@
         :architecture => 'bar',
         :osfamily => osfamily,
         :lsbdistid => osfamily,
-        :lsbdistcodename => 'baz',
-        :mco_server_config => nil,
-        :mco_client_config => nil,
+        :lsbdistcodename => 'baz'
```
speedofdark: Nit: does this need a trailing comma?

ekinanp: Nope

README.markdown
```diff
@@ -99,12 +65,10 @@ This will ensure the version `1.4.0` of the puppet-agent package is installed. F
 
 ### Private classes
 * `puppet_agent::install`: Installs packages.
-* `puppet_agent::install::remove_packages`: For platforms that can't perform in-place upgrades, removes the old packages.
-* `puppet_agent::install::remove_packages_osx`: Removes the old packages on Mac OS X.
 * `puppet_agent::osfamily::*`: Platform-specific preparation performed before upgrades.
 * `puppet_agent::prepare`: Prepares the agent for upgrade.
 * `puppet_agent::prepare::package`: Stages packages locally for install, on platforms that can't install from remote packages.
-* `puppet_agent::prepare::*`: Prepare various file and ssl configuration.
+* `puppet_agent::prepare::*`: Prepare various file configuration.
```
speedofdark: configurations?

manifests/init.pp
ekinanp: I'll push up another commit to add this capability then.

ekinanp:
Pushed up another commit to error if the `aio_agent_version` fact is undefined.

README.markdown
ekinanp: Fixed

PR #376, opened 2019-01-26
(MODULES-8348) Refactor the acceptance test scaffold

ekinanp:


acceptance/helpers.rb
```diff
@@ -16,16 +16,139 @@ def agents_only
...
+      # the puppet4 collection or newer, but no newer than puppet5. The second restriction
+      # will only run the test if the master's puppetserver is from the puppet4 collection
+      # or newer. The third restriction will only run the test if the master's puppetserver
+      # is from the puppet6 collection or older.
+      #
+      # NOTE: Be sure to put this DSL first in your test, before any confines. Otherwise,
+      # you could accidentally filter out your master and thus raise an exception here.
+      #
+      # @param [Hash] args The specified restriction
+      def master_collection(args)
```
ekinanp: I think something akin to this DSL would be useful since we'll have a few tests that test collection-specific behavior (aside from the existing two that are already there). I just don't know if `master_collection` is the right name. The reason I based it off of the server version instead of the agent version as done previously is because the master's only involvement in agent upgrades is compiling the catalog, which is what the server does.

That being said, I think a name containing the word `collection` does make sense b/c it's how we think of upgrades.

acceptance/tests/test_upgrade_pc1_to_puppet5.rb
```diff
@@ -2,29 +2,11 @@
 require_relative '../helpers'
 
 # Tests FOSS upgrades from the latest puppet 4 (the PC1 collection) to puppet 5.5.10.
-# This test will only run if the version of puppet on the master host is less than 5.
 test_name 'puppet_agent class: collection parameter for FOSS upgrades' do
-  master_agent_version = puppet_agent_version_on(master)
+  master_collection restrict_to: 'puppet5'
```
ekinanp: I think this captures the general pattern of an agent upgrade test based on what I saw from the existing code.

The DSL helper `master_collection` replaces the existing `agent_version` checks. Reason I chose collection is b/c it is more consistent with how we think about upgrades.

`is_foss_upgrade_test` replaces the existing `PE_ONLY_UPGRADES` confine. Once we start testing PE upgrades here, we may likely have a similar `is_pe_upgrade_test` helper.

`setup_agents_to_upgrade` is essentially `prepare_upgrade_with` but there are some caveats:
* The method takes in an optional block for additional set-up. For example for a Windows upgrade test, you can do something like
```
setup_agents_to_upgrade('pc1') do
  <start services here>
end
```

* Any subsequent teardown blocks _after_ the invocation of `setup_agents_to_upgrade` _cannot_ use puppet, facter, i.e. anything from the puppet_agent package. For example, something like
```
setup_agents_to_upgrade('pc1')
teardown { on(agent, puppet('--version')) }
```

will fail because `setup_agents_to_upgrade` has some teardown blocks to clean the agents so that by the time we get to the declared teardown block, puppet-agent will have already been uninstalled. However, something like
```
setup_agents_to_upgrade('pc1')
on(agent, puppet('--version'))
```

will work.

There's a technical reason why `purge_agents`' code was moved over to `setup_agents_to_upgrade` that I'll explain in a separate PR comment underneath the method.


There's a technical reason why the block pattern that I'll discuss in a separate PR comment underneath the method.

acceptance/helpers.rb
```diff
@@ -41,21 +164,35 @@ def install_puppet_agent_module_on(host)
...
+      # Install puppet-agent on all agent nodes and connects them to the master.
+      # This is intended to setup SUTs for tests that upgrade puppet-agent
+      # from some initial version. A key post-condition of this method is that
+      # all of the agents are ready to be upgraded with the puppet_agent module.
       #
       # @param [String] initial_package_version_or_collection Either a version
       #   of puppet-agent or the name of a puppet collection to install the agent from.
-      def prepare_upgrade_with(initial_package_version_or_collection)
+      # @yield Invokes any additional set-up that's required. 
+      def setup_agents_to_upgrade(initial_package_version_or_collection)
```
ekinanp: The reason why `purge_agents`' code was moved here and why this method now takes a block is because that is the best solution I could come up with that would ensure a clean SUT between each test, including _failed_ tests. The latter was a big pain point for me while working on this PR because I'd have to (manually) clean out my agent every time my acceptance tests failed (re-running the `rake prepare` task also took a while).

ekinanp: Here were some of the other solutions I thought of. The existing way
```
setup_agents_to_upgrade('pc1')
teardown { purge_agents }
```

wouldn't work if one needed to do additional preparation before an upgrade. For example, given something like
```
setup_agents_to_upgrade('pc1')
// additonal prep here
teardown { purge_agents }
```
, if anything in the `additional prep` part of the code threw an exception, then the `purge_agents` teardown would not be included in Beaker's clean-up, so you'd need to manually go and clean your SUTs.

Something like
```
begin
  setup_agents_to_upgrade('pc1')
  // additional prep here
ensure
  teardown { purge_agents }
end
```

is also bad because (1) you need to remember to do that for every agent upgrade test to ensure a clean SUT and (2), if something in `setup_agents_to_upgrade` throws an exception, `purge_agents` will be called on agents that aren't properly configured.

ekinanp: The current solution of
```
setup_agents_to_upgrade('pc1') do
   // additional prep here
end
```

is also not great because you need to remember to declare your prep. in a block, but outside of that one thing, you can code your prep. anyway you like since any teardowns inside the block are applied _before_ the code in `purge_agents`.

acceptance/helpers.rb
```diff
@@ -66,114 +203,125 @@ def prepare_upgrade_with(initial_package_version_or_collection)
...
-            master_fqdn = on(master, facter('fqdn')).stdout.strip
-            master_hostname = on(master, facter('hostname')).stdout.strip
-            master_conf = { 'main' => {
-                'autosign'      => true,
-                'dns_alt_names' => "puppet,#{master_hostname},#{master_fqdn}",
-                'verbose'       => true,
-                'daemonize'     => true,
-            }}
-
-            with_puppet_running_on(master, master_conf) do
```
ekinanp: Another thing I noticed was that `with_puppet_running_on` spent a lot of time restarting puppetserver and it also cluttered the output when `BEAKER_debug` was set. I based the cert-signing on what's in [ValidateSignCert](https://github.com/puppetlabs/beaker-puppet/blob/master/setup/common/040_ValidateSignCert.rb) instead since that avoids the restart.

acceptance/helpers.rb
ekinanp: I'm happy to remove this too if it is a bit of a "too much too early" type of thing.

ekinanp:
I'll need to update the README prior to merging this if the changes here are approved.

acceptance/helpers.rb
```diff
@@ -66,173 +203,205 @@ def prepare_upgrade_with(initial_package_version_or_collection)
...
+      # Generates a manifest that declares the puppet_agent class.
       #
-      # - Install the a puppet-agent package (`initial_package_version_or_collection`)
-      #   on all the agent hosts,
-      # - Put a teardown step in place so that this package is uninstalled after the test,
-      # - Run a default manifest on all the agent hosts, and
-      # - Allow for assertions inside in a block
+      # @param [Hash] params The class parameters
+      # @return [String] The generated puppet_agent class declaration
+      def class__puppet_agent(params = {})
```
McdonaldSeanp: Nit: Do we lose any clarity around what the manifest should look like by including this helper?

In the original tests seeing the manifest in it's entirety might make it easier for someone new to the codebase to grasp what the manifest would look like.  Not a huge deal since the helper just wraps the manifest params in the classname, but if I were new to the module I would still need to look up this helper to find that out.

acceptance/helpers.rb
McdonaldSeanp: Do you have to restart puppetserver after clearing out old certs? I don't actually know if that's the case but if it is we need to ensure that's happening.

acceptance/helpers.rb
McdonaldSeanp: Is it possible to include the teardown inside the setup_agents_to_upgrade function? https://github.com/puppetlabs/beaker/blob/a01b5732a7d6cec146bd18d08795c64aafe1d2f1/docs/concepts/style_guide.md#teardowns indicates there's no reason we need to wait until after the function completes to specify the teardown.

Then maybe a check in purge_agents to identify if there's an agent installed and skip the teardown if there isn't one?

acceptance/helpers.rb
```diff
@@ -66,173 +203,205 @@ def prepare_upgrade_with(initial_package_version_or_collection)
...
+
+        step "(Master) Save the current site.pp file" do
           if file_exists_on(master, site_pp_path)
             original_contents = file_contents_on(master, site_pp_path)
             original_perms = on(master, %(stat -c "%a" #{site_pp_path})).stdout.strip
-
-            teardown do
-              step "restore original manifest" do
+  
+            cleanup = lambda do
```
McdonaldSeanp: What was the reason to change the teardowns to lambdas?

ekinanp: This is so we'd reset the `site.pp` file after each method invocation instead of after the test finishes. That way, this function can be re-used many times in the same test.

acceptance/helpers.rb
ekinanp: Asked in the server room just now and it looks like the answer is "no".

acceptance/helpers.rb
ekinanp: Addressed this in HipChat. Reason we need the purge_agents teardown at the end of the function is so that something like
```
setup_agents_to_upgrade('pc1') do
  <manipulate some state with Puppet>
  teardown { <restore that state with puppet> }
end
```

would work. Otherwise, when the "restore state" teardown is reached, the agent will have already been uninstalled.

McdonaldSeanp:
Based on the responses the reasoning behind replacing things with blocks/lambdas seems reasonable.

acceptance/tests/test_upgrade_puppet5_to_puppet6.rb
```diff
@@ -2,28 +2,21 @@
 require_relative '../helpers'
 
 # Tests FOSS upgrades from the latest puppet 5 (the puppet5 collection) to puppet 6.0.0.
-# This test will only run if the version of puppet on the master host is less than 6.
 test_name 'puppet_agent class: collection parameter for FOSS upgrades' do
-  master_agent_version = puppet_agent_version_on(master)
+  master_collection restrict_to: 'puppet5'
```
caseywilliams: There are a few things in these updated tests that look like readability problems to me:

- It's not obvious to me from the name that `is_foss_ugprade_test` is restricting the upgrade from PE platforms -- that name sounds like a boolean check of some kind. I'd prefer something like `exclude_pe_upgrade_platforms` or similar that describes the action it's doing.
- Likewise, I'd prefer a verb-oriented method name to replace `master_collection`. I'd prefer something like `require_master_collection('puppet5')`, `require_master_collection(max: 'puppet5')`, or similar.
- It seems like `assert_successful_upgrade` would be better titled `assert_agent_version` or similar. A 'successful upgrade' may entail more than just updating the puppet-agent version: we may want to assert that certain services are stopped or running depending on the parameters that were used, too.
- This is more of a nitpick, but the words 'setup', 'set-up', and 'set up' are all being used here. 'set up' is the verb, and 'setup' is the noun version -- I'd prefer these to be consistent.

I am not a fan of the lambdas, which are not very ruby-like and unfamiliar to people who are used to writing normal beaker tests already, but it's not a big deal.

ekinanp: Updated the PR to address your concerns. Here are the changes:

* Changed `is_foss_upgrade_test` => `exclude_pe_upgrades`
* Changed to `require_master_collection` as you specified -- that is cleaner.
* `assert_successful_upgrade` now takes a block so you can do something like
```
assert_succesful_upgrade('6.0.0') do |agent|
  // additional assertions
end
```
* Addressed the set-up (verb) vs. setup (noun) inconsistency
* Updated the commit to reference MODULES-8348

ekinanp:
https://github.com/puppetlabs/beaker-puppet/pull/104 + a beaker-puppet release will need to happen before this one can go in. Sorry it's late, just now remembered to put up that PR.

caseywilliams:
I've released beaker-puppet 1.16.0 with your changes. I will merge this and update the minimum beaker-puppet version in the acceptance Gemfile separately.

PR #379, opened 2019-01-30
(PE-25542) add RHEL8 to puppet agent module

loopinu:
added red hat 8 to metadata.json & puppet_agent_osfamily_redhad_spec.rb

Please do not merge yet.


caseywilliams:
@donoghuc I just saw your comment -- my understanding is that the invalid path that the task currently relies on is not the "official" download path, and that the second one you mention is correct for all platforms (although checking with Release Engineering may be helpful) -- I believe that the older paths there are kept intact for those who still rely on them (although I do find their presence confusing!)

mwaggett:
release packages for el 8 have now been shipped, so downloading from http://yum.puppetlabs.com/puppet/puppet-release-el-8.noarch.rpm should work

donoghuc:
Just confirmed that this now works!

PR #380, opened 2019-01-31
(MODULES-8320) Remove old platforms

ekinanp:


manifests/install.pp
```diff
@@ -63,7 +59,7 @@
         command => "/usr/bin/ctrun -l none ${_installsh} ${::puppet_agent_pid} 2>&1 > ${_logfile} &",
       }
     }
-  } elsif $::operatingsystem == 'Darwin' and $::macosx_productversion_major =~ /^10\.(9|10|11|12|13)/ {
+  } elsif $::operatingsystem == 'Darwin' {
```
ekinanp: We don't need the regex check at the end because the `puppet_agent::prepare` class is included before the `puppet_agent::install` class, and that `prepare` class includes the Darwin OS family class which [has the regex check](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/manifests/osfamily/darwin.pp#L6)

PR #381, opened 2019-02-01
(MODULES-5535) Update test matrix to work with MacOS/Solaris/Windows

McdonaldSeanp:
Currently, the update testing in the acceptance directory does not
test by waiting for pidfiles during installations. For
MacOS/Solaris/Windows this means the acceptance tests will not work,
since those platforms will spawn scripts to perform installations
outside of an agent run

So this commit updates the acceptance testing to include waiting for
pidfiles after the puppet run completes so the testing will wait for
the full installation.

There were a few larger refactors done to help facilitate the changes:

1. new functions for waiting for puppet to finish and waiting on
installation PIDs

2. the actual execution of puppet to perform the upgrade now consists
of turning on the puppet service rather than executing 'puppet agent
-t'. This is the approach we use when testing upgrades in PE and
should be consistent

3. several functions were refactored to be smaller and more focused.

### DIFF
--- a/acceptance/helpers.rb
+++ b/acceptance/helpers.rb
@@ -74,7 +74,7 @@ module Beaker
         args_are_not_a_valid_hash =
           ! args.is_a?(Hash) ||
           args.empty? ||
-          ! Set[*args.keys].subset?(Set[:restrict_to, :max, :min]) 
+          ! Set[*args.keys].subset?(Set[:restrict_to, :max, :min])
 
         if (! args.is_a?(Symbol)) && args_are_not_a_valid_hash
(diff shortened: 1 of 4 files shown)


acceptance/helpers.rb
```diff
@@ -177,27 +204,29 @@ def host_to_info_s(host)
...
+      # Installs an initial puppet agent package on a host and connect the new agent node to the
+      # puppet master. This 'initial' agent is likely a lower version than the puppet master on
+      # purpose to facilitate an upgrade scenario.
       #
+      # @param [Beaker::Host] host The host
       # @param [String] initial_package_version_or_collection Either a version
       #   of puppet-agent or the name of a puppet collection to install the agent from.
-      # @yield Invokes any additional setup that's required. 
-      def set_up_agents_to_upgrade(initial_package_version_or_collection)
+      def set_up_initial_agent(host, initial_package_version_or_collection)
```
caseywilliams: I think this method should have `_on` appended, too, to make it clear that it expects a host

PR #384, opened 2019-02-06
(MODULES-4986) Remove deprecated puppet 4 settings on upgrading to puppet 5+

caseywilliams:
There are two commits here; The first includes a few readability and typo updates to the acceptance tests, and the second ensures that deprecated settings from puppet 4 (app_management, ignorecache, configtimeout, and trusted_server_facts) are removed on upgrading to any version of puppet-agent beyond 5.0.0.


acceptance/pre_suite/00_master_setup.rb
```diff
@@ -43,8 +43,6 @@
 
     server_version = puppetserver_version_on(master)
     fail_test('Failed to install puppetserver') unless server_version
-
-    logger.notify("Installed puppetserver #{server_version} on master")
```
caseywilliams: the `puppetserver_version_on` helper already logs this

acceptance/helpers.rb
```diff
@@ -23,8 +23,8 @@ def collection_cmp(col_1, col_2)
 
         return 0 if col_1 == col_2
 
-        return -1 if col_1 == 'pc1'
-        return 1 if col_2 == 'pc1'
+        return -1 if col_1 =~ /pc1/i
+        return 1 if col_2 =~ /pc1/i
```
smcclellan: Are these changes covered by tests?

And is the only intended change here to handle mixed case? `casecmp` might be a better candidate if that's so.

caseywilliams: Well, there are no tests for these helpers, but this method is used in the acceptance tests to restrict which tests run based on the version of puppetserver installed on the master -- that functionality still works for me. This change is intended to allow for using either PC1 (the "official" name of the collection) or pc1 (required in some download URLs, so also common).

caseywilliams: oh, I totally igonred the second part of your comment -- I'll use casecmmp

PR #385, opened 2019-02-07
(MODULES-8599) Refactor sources and enable Darwin FOSS installs

McdonaldSeanp:
This commit performs a major refactor of gathering source URLs for packages in order to facilitate open-source upgrade scenarios for Darwin OSes. 

package sources are now all generated inside the osfamily classes.

Additionally, pidfile locations for MacOS and Solaris have been moved from tempdirs to the puppet statedir


manifests/params.pp
```diff
@@ -15,101 +15,67 @@
...
+  $config = "${confdir}/puppet.conf"
 
-      $user  = 0
-      $group = 0
+  # The `is_pe` fact currently works by echoing out the puppet version
+  # and greping for "puppet enterprise". With Puppet 4 and PE 2015.2, there
+  # is no longer a "PE Puppet", and so that fact will no longer work.
+  # Instead check both the `is_pe` fact as well as if a PE provided
+  # function is available
+  $_is_pe = (getvar('::is_pe') or is_function_available('pe_compiling_server_version'))
```
speedofdark: This is checking for either, and the comment says checking for both?

manifests/osfamily/darwin.pp
```diff
@@ -1,14 +1,6 @@
 class puppet_agent::osfamily::darwin{
```
speedofdark: What do the last two lines of the commit message mean?

manifests/osfamily/aix.pp
```diff
@@ -11,9 +9,23 @@
     fail('AIX upgrades are only supported on Puppet Enterprise')
   }
 
-  class { 'puppet_agent::prepare::package':
-    package_file_name => $package_file_name,
+  $pe_server_version = pe_build_version()
+
+  $_aix_ver_number = regsubst($::platform_tag,'aix-(\d+\.\d+)-power','\1')
```
caseywilliams: I feel like if I were learning this codebase later, I'd appreciate a comment here about why this this part is necessary for puppet 4 and 5

speedofdark:
Also looks like travis is griping about mcollective?

manifests/params.pp
```diff
@@ -15,105 +15,67 @@
...
-    true    => "el-6-${::architecture}",
-    default =>  $_platform_tag,
-  }
-
-  # The aio puppet-agent version currently installed on the compiling master
-  # (only used in PE)
-  if ($_is_pe and is_function_available('pe_compiling_server_aio_build')) {
-    $master_agent_version = pe_compiling_server_aio_build()
   } else {
+    $_pe_version = undef
```
speedofdark: Yay!

manifests/osfamily/darwin.pp
McdonaldSeanp: Oh, those are garbage. Must have been a copy/paste error in my editor

manifests/params.pp
branan: does `getvar('::is_pe')` work when that fact is missing and Puppet is in strict mode?

McdonaldSeanp: No clue

McdonaldSeanp:
I tested this in PE with some smoke test platforms: AIX/Redhat/Windows and then with a bunch more FOSS: debian/redhat/MacOS/Windows

manifests/params.pp
branan: I think that's the point of `getvar`, but I honestly can't remember. The docs on it aren't super clear how it behaves in strict mode

branan: If we just moved it, it's fine

caseywilliams:
Mk, ran a few more tests -- macOS looks good now, but I'm seeing failures for centos 5 (but not redhat 5) when upgrading PC1 -> puppet5

PR #388, opened 2019-02-14
(MODULES-8665) Add missing puppetlabs-facts dependency

Sharpie:
This commit adds the puppetlabs-facts module to the dependency list of
puppetlabs-puppet_agent. Tasks from the facts module are required by
the puppet_agent::install task.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -11,5 +11,8 @@ fixtures:
       ref: "6.0.0"
     yumrepo_core:
       repo: "puppetlabs/yumrepo_core"
+    facts:
+      repo: "puppetlabs/facts"
+      ref: "0.5.1"
   symlinks:
     puppet_agent: "#{source_dir}"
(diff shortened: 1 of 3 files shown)


McdonaldSeanp:
@Sharpie you'll need to add the new module to https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/.fixtures.yml for spec testing to work

McdonaldSeanp:
@Sharpie if you could also add this new module to the installs at https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/acceptance/helpers.rb#L184 that would make acceptance continue to work too.

Sharpie:
Flagged as don't merge as there are some open questions in [MODULES-8665](https://tickets.puppetlabs.com/browse/MODULES-8665) as to whether this is the right thing to do.

PE users should use tasks from the `pe_bootstrap` module instead of `puppet_agent::install`.

Sharpie:
Rebased onto master and fixed test failure.

Sharpie:
Bumped `task_spec/.fixtures.yml` to 0.5.1.

PR #390, opened 2019-02-21
(FM-7628) Update install_puppet.ps1 to catch hanging pxp-agent processes

McdonaldSeanp:
We have seen cases with customers where after shutting down the pxp-agent
service there are still pxp-agent processes open.

This commit adds a workaround to the install_puppet.ps1 powershell script to
wait for any open pxp-agent processes to close, and if they don't the upgrade
script will stop before actually attempting the upgrade.

### DIFF
--- a/templates/install_puppet.ps1.erb
+++ b/templates/install_puppet.ps1.erb
@@ -284,6 +284,21 @@ try {
       Stop-Service $service
     }
   }
+  # Wait for any pxp-agent process still hanging around
+  #
+  # There is a known problem for pxp-agent shutdown: there are cases where after service
+  # shutdown pxp-agent processes are still open. See https://tickets.puppetlabs.com/browse/FM-7628
+  # for more details on the symptoms of this.
+  Write-Log "Waiting for pxp-agent processes to stop"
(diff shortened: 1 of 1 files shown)


templates/install_puppet.ps1.erb
```diff
@@ -284,6 +284,20 @@ try {
...
   }
+  # Wait for any pxp-agent process still hanging around
+  #
+  # There is a known problem for pxp-agent shutdown: there are cases where after service
+  # shutdown pxp-agent processes are still open. See https://tickets.puppetlabs.com/browse/FM-7628
+  # for more details on the symptoms of this.
+  Write-Log "Waiting for pxp-agent processes to stop"
+  Get-Process -Name "pxp-agent" -ErrorAction SilentlyContinue | ForEach-Object {
+    if ($_) {
+      if (!$_.WaitForExit(120000)){
```
branan: What unit is this timeout in? Would be nice to have a comment with human-readable format

McdonaldSeanp: It's milliseconds: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netframework-4.7.2#System_Diagnostics_Process_WaitForExit_System_Int32_

I'll add a comment

PR #392, opened 2019-03-01
Fix unnecessary changes in load balanced envs due to pkg source

ragnarkon:
In environments with multiple compile masters behind a load balancer,
using `$::servername` could cause Puppet to constantly update the repo
configuration based on the compile master that happens to be compiling
the catalog at that time.

Changing the value to `$::puppet_master_server` uses the server value in
puppet.conf, which in most (all?) cases should be the load balanced point.

### DIFF
--- a/manifests/osfamily/debian.pp
+++ b/manifests/osfamily/debian.pp
@@ -10,7 +10,7 @@ class puppet_agent::osfamily::debian{
       if $::puppet_agent::source {
         $source = $::puppet_agent::source
       } else {
-        $source = "https://${::servername}:8140/packages/${pe_server_version}/${::platform_tag}"
+        $source = "https://${::puppet_master_server}:8140/packages/${pe_server_version}/${::platform_tag}"
       }
       # In Puppet Enterprise, agent packages are served by the same server
       # as the master, which can be using either a self signed CA, or an external CA.
--- a/lib/facter/settings.rb
+++ b/lib/facter/settings.rb
@@ -36,7 +36,17 @@ end
 
 Facter.add('puppet_master_server') do
   setcode do
-    Puppet.settings['server']
+    if Puppet.settings['server_list'].nil? || Puppet.settings['server_list'].empty?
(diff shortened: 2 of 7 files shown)


ragnarkon:
I'd also be open to simply updating the README with a note about adjusting the `source` parameter. But using `$::puppet_master_server` seems to be the more appropriate value to use as the default (unless I'm missing something).

manifests/osfamily/debian.pp
```diff
@@ -7,7 +7,7 @@
 
     if getvar('::puppet_agent::is_pe') == true {
       $pe_server_version = pe_build_version()
-      $source = "https://${::servername}:8140/packages/${pe_server_version}/${::platform_tag}"
+      $source = "https://${::puppet_master_server}:8140/packages/${pe_server_version}/${::platform_tag}"
```
branan: I'm not certain that `puppet_master_server` will contain the right value when a node has been configured with `server_list` only (and no `server` setting)

ragnarkon: So... I tested it and yes... if you specify a `server_list` setting without specifying a `server`, it'll set the fact to the default setting of `puppet`.  Good catch.

Two paths forward that I can think of:

1. Update the fact to use the value for `server` _or_ the first element in the `server_list` array if it is defined.
2. Scrap the whole thing and just do a README update.

We could just update the default value for `$source` as it have it in this PR anyway, but that could potentially be a breaking change for the few people out there that do use `server_list`.

ragnarkon: Option 1 presented in new commit c48427d (compliments to @Sharpie).

ragnarkon:
Rebased to account for #394.

Also noticed that #394 will be a breaking change those PE users who previously set the `source` parameter in the the module to point to a load balanced point. Which gives me greater hope that this will be merged 👍

ragnarkon:
Rebased to account for #394.

PR #393, opened 2019-03-05
(MODULES-8554) Add error reporting for background upgrades

McdonaldSeanp:


lib/puppet/type/upgrade_error.rb
```diff
@@ -0,0 +1,25 @@
+Puppet::Type.newtype(:upgrade_error) do
+  @doc = <<-DOC
+Fails when a previous backgroun installation failed. The type
```
caseywilliams: Small typo here

lib/puppet/provider/puppet_agent_upgrade_error/puppet_agent_upgrade_error.rb
```diff
@@ -0,0 +1,25 @@
+Puppet::Type.type(:puppet_agent_upgrade_error).provide :puppet_agent_upgrade_error do
+  def ensure_notexist
+    logfile = File.join(Puppet["statedir"].to_s, @resource[:name])
+    Puppet.debug "Checking for Error logfile #{logfile}"
+    not File.exists?(logfile)
+  end
+
+  def delete_file_and_read_content(filename)
```
speedofdark: Nit: swap order of verbs? Since we're reading it, then deleting it?

PR #395, opened 2019-03-12
(MODULES-6604) Add new source parameters

McdonaldSeanp:
This commit adds new source parameters:
* absolute_source
* yum_source
* apt_source
* mac_source
* windows_source
* solaris_source
* aix_source
* use_alternate_sources
* alternate_master_source

Each source parameter will modify where the module attempts to pull packages
for different use cases.

Absolute_source: This parameter is for use cases where the package source
should be an absolute path on the filesystem. It's a common windows paradigm
to use a network share to serve files to systems behind a firewall. This param
should server that purpose.

(yum/apt/mac etc.)_source: these parameters are for use cases where someone has
mirrored the public downloads sites for puppet and they just need to change the
underlying base URL for grabbing packages. Hopefully this enables workflows
like using artifactory with PE to serve agents.

use_alternate_sources: This will force PE installs to use the yum/apt/mac etc.
sources rather than the pe_repo file share from the master. This should give
PE users the same functionality as FOSS users with respect to the (yum/apt/mac
etc.)_source params.

alternate_master_source: this parameter is for use cases where files should be
served from a server other than the PE master, but the files are still served
in the same directory structure you would see in a PE install. This should
serve use cases where either 1. a user wants to use a specific master to serve
all agent packages or 2. a situation in which pe_repo is deployed somewhere
other than the PE master

Additionally: the normal $source parameter has been modified to work in a
backwards compatible way to older versions of this module:

For unix/Mac agents, $source will replace the PE master base URL. For windows
agents, $source will be a fully qualified path to an MSI package.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3,22 +3,50 @@
 [![Build Status](https://travis-ci.org/puppetlabs/puppetlabs-puppet_agent.svg?branch=master)](https://travis-ci.org/puppetlabs/puppetlabs-puppet_agent)
 
 #### Table of Contents
-
-1. [Overview](#overview)
-2. [Module Description - What the module does and why it is useful](#module-description)
-3. [Setup - The basics of getting started with puppet_agent](#setup)
-    * [What puppet_agent affects](#what-puppet_agent-affects)
-    * [Setup requirements](#setup-requirements)
-    * [Beginning with puppet_agent](#beginning-with-puppet_agent)
-4. [Usage - Configuration options and additional functionality](#usage)
-5. [Reference](#reference)
-    * [Public classes](#public-classes)
-    * [Private classes](#private-classes)
-    * [Parameters](#parameters)
-    * [Tasks](#tasks)
-6. [Limitations - OS compatibility, etc.](#limitations)
-    * [Known issues](#known-issues)
-7. [Development - Guide for contributing to the module](#development)
+- [puppet_agent](#puppet_agent)
+      - [Table of Contents](#table-of-contents)
+  - [Overview](#overview)
+  - [Module Description](#module-description)
+  - [Setup](#setup)
+    - [What puppet_agent affects](#what-puppet_agent-affects)
+    - [Setup requirements](#setup-requirements)
+    - [Beginning with puppet_agent](#beginning-with-puppet_agent)
+  - [Usage](#usage)
+  - [Using alternate sources](#using-alternate-sources)
+    - [Public downloads mirrors](#public-downloads-mirrors)
+    - [Absolute paths to packages](#absolute-paths-to-packages)
+    - [Alternate PE master location](#alternate-pe-master-location)
+  - [Reference](#reference)
+    - [Public classes](#public-classes)
+    - [Private classes](#private-classes)
+    - [Parameters](#parameters)
+      - [Class: puppet_agent](#class-puppet_agent)
+        - [`arch`](#arch)
+        - [`collection`](#collection)
+        - [`is_pe`](#is_pe)
+        - [`manage_repo`](#manage_repo)
+        - [`package_version`](#package_version)
+        - [`service_names`](#service_names)
+        - [`source`](#source)
+        - [`absolute_source`](#absolute_source)
+        - [`yum_source`](#yum_source)
+        - [`apt_source`](#apt_source)
+        - [`mac_source`](#mac_source)
+        - [`windows_source`](#windows_source)
+        - [`solaris_source`](#solaris_source)
+        - [`aix_source`](#aix_source)
+        - [`use_alternate_sources`](#use_alternate_sources)
+        - [`alternate_pe_source`](#alternate_pe_source)
+        - [`install_dir`](#install_dir)
+        - [`install_options`](#install_options)
+        - [`msi_move_locked_files`](#msi_move_locked_files)
+    - [Tasks](#tasks)
+      - [`puppet_agent::version`](#puppet_agentversion)
+      - [`puppet_agent::install`](#puppet_agentinstall)
+  - [Limitations](#limitations)
+    - [Known issues](#known-issues)
+  - [Development](#development)
+  - [Maintenance](#maintenance)
 
 ## Overview
 
@@ -57,6 +85,35 @@ class {'::puppet_agent':
 ~~~
 
 This will ensure the version `1.4.0` of the puppet-agent package is installed. For version `1.4.0` and later, it will also remove the deprecated `pluginsync` setting from `puppet.conf`, unless explicitly managed elsewhere.
+
+## Using alternate sources
+
+In cases where you wish to download agents from sources other than the defaults you can use source parameters to change the location to grab packages from.
+
+### Public downloads mirrors
+
+If you wish to mirror the Puppet public downloads sites (yum.puppet.com, apt.puppet.com, downloads.puppet.com) you can provide the following parameters to change the location of downloads:
+
+* `yum_source`
+* `apt_source`
+* `mac_source`
+* `windows_source`
+* `solaris_source`
+* `aix_source`
+
+For AIX and Solaris packages: because AIX and Solaris are PE only you must use puppetlabs-pe_repo to create repos for these platforms on the PE master, then mirror the PE master package serve.
+
+When working with a PE installation: if you set `use_alternate_sources` to `true` you can force agent downloads to come from downloads sites (or a mirror if you set the source parameters) rather than the PE master. **WARNING** This parameter will override the default settings in PE installations to download packages from the PE master. If you wish to continue to download from the PE master _do not_ set this parameter.
+
+### Absolute paths to packages
+
+If your packages are already available on the target system (for example if you are using a network share) you can provide `absolute_source` the path to the packages to use during installation.
+
+**WARNING** You must provide the full path, including the package name, for this parameter to work. This also means you cannot provide the same `absolute_source` for two different types of packages.
+
+### Alternate PE master location
+
+If you are using puppetlabs-pe_repo to serve packages, but want to provide a location other than the current master to serve packages: use `alternate_pe_source` to specify a seperate location where packages are located in the same structure that would be on a PE master.
 
 ## Reference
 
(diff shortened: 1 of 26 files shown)


manifests/init.pp
```diff
@@ -28,7 +28,14 @@
 #   An array of services to start, normally `puppet` and `mcollective`.
 #   None will be started if the array is empty.
 # [source]
-#   The location to find packages.
+#   The url of a host where the module attempts to grab packages from. When attempting to
+#   find packages with 'source': the module will assume consistent directory structures
+#   that match either the structure of a PE master serving packages or the structure of the
+#   public Puppet repositories. This parameter cannot be used if 'source_package_location'
```
caseywilliams: Small typo: `source_package_location` => `absolute_source`

README.markdown
```diff
@@ -122,11 +122,42 @@ An array of services to start, normally `puppet`. If the array is empty, no serv
...
+``` puppet
+  mirror_source => 'https://my-puppet-mirror.com'
+```
+
+##### `alternate_pe_source`
+
+Base URL of a location where packages are located in the same structure that's served by a PE master (the directory structure in PE for serving packages is created by the puppetlabs-pe_repo module). The general structure served by PE is: `/packages/${pe_server_version}/${platform_tag}/${package_name}`
 ``` puppet
-  source => 'C:/packages/puppet-agent-1.7.0-x64.msi'
+  alternate_pe_source => 'https://my-alternate-pe-master.com'
```
ragnarkon: Example is missing the port number.
`alternate_pe_source => 'https://my-alternate-pe-master.com:8140'`

ragnarkon:
Based off of the README text, I would expect setting an `absolute_source` parameter to skip using a package repo and instead use a package file directly off of the file system for all operating systems, not just those without remote package repository abilities.

ie.
```puppet
package { 'puppet-agent':
  source   => $::puppet_agent::absolute_source,
  provider => 'dpkg',
}
```

McdonaldSeanp:
I've tested the upgrades with PE using nightlies as an alternate source and this appears to be working for everything.

manifests/install.pp
```diff
@@ -14,86 +14,81 @@
...
+      $_install_options = concat(['--ignoreos'],$install_options)
+      $_provider = 'rpm'
+      $_source = "${::puppet_agent::params::local_packages_dir}/${::puppet_agent::prepare::package::package_file_name}"
+    } elsif $::osfamily == 'Debian' {
+      $_install_options = $install_options
+      if $::puppet_agent::absolute_source {
+        # absolute_source means we use dpkg on debian based platforms
+        $_package_version = 'present'
+        $_provider = 'dpkg'
+        $_source = "${::puppet_agent::params::local_packages_dir}/${::puppet_agent::prepare::package::package_file_name}"
```
ragnarkon: Should be `$_source = $::puppet_agent::absolute_source`?

Not sure package would get to `"${::puppet_agent::params::local_packages_dir}/${::puppet_agent::prepare::package::package_file_name}"` unless the `puppet_agent::prepare::package` class was also used for Debian/RHEL/SUSE/etc.

README.markdown
```diff
@@ -58,6 +86,35 @@ class {'::puppet_agent':
...
+If you wish to mirror the Puppet public downloads sites (yum.puppet.com, apt.puppet.com, downloads.puppet.com) you can provide the following parameters to change the location of downloads:
+
+* `yum_source`
+* `apt_source`
+* `mac_source`
+* `windows_source`
+* `solaris_source`
+* `aix_source`
+
+For AIX and Solaris packages: because AIX and Solaris are PE only you must use puppetlabs-pe_repo to create repos for these platforms on the PE master, then mirror the PE master package serve.
```
speedofdark: emphasize **must**?

README.markdown
```diff
@@ -122,11 +179,87 @@ An array of services to start, normally `puppet`. If the array is empty, no serv
   service_names => ['puppet']
 ```
 
+
 ##### `source`
 
-Alternate source from which you wish to download the latest version of Puppet. On the Windows operating system this is the absolute path to the MSI file to install, for example:
+INCLUDED FOR COMPATIBILITY WITH MODULE VERSIONS 1.0/2.0. PREFER USE OF "absolute_source", "(yum/apt/mac etc.)_source", "alternate_pe_source" OVER USE OF "source".
```
speedofdark: make "absolute_source" italic for consistency.

README.markdown
```diff
@@ -122,11 +179,87 @@ An array of services to start, normally `puppet`. If the array is empty, no serv
...
+
+The location to find packages. Replaces base URL for unix/MacOS agents, used as fully qualified path in windows.
```
speedofdark: switch back to not-italic?

README.markdown
```diff
@@ -122,11 +179,87 @@ An array of services to start, normally `puppet`. If the array is empty, no serv
...
+```
+
+##### `aix_source`
+
+Base URL of the location of a mirror for AIX packages. Currently, AIX packages can only be made available by using puppetlabs-pe_repo. This means the mirror must be of a PE master package serve.
+``` puppet
+  solaris_source => 'https://my-pe_master-mirror.com'
+```
+
+
```
speedofdark: I see a lot of "must match structure" types of comments. I'm assuming we can't readily detect this and just do the right thing?

manifests/init.pp
```diff
@@ -47,26 +86,39 @@
...
+  $collection              = $::puppet_agent::params::collection,
+  $is_pe                   = $::puppet_agent::params::_is_pe,
+  $manage_pki_dir          = true,
+  $manage_repo             = true,
+  $package_name            = 'puppet-agent',
+  $package_version         = undef,
+  $service_names           = $::puppet_agent::params::service_names,
+  $source                  = undef,
+  $absolute_source         = undef,
+  $yum_source              = 'http://yum.puppet.com',
```
speedofdark: insecure make me sad.

manifests/install.pp
```diff
@@ -14,86 +14,81 @@
...
+        $_provider = 'apt'
+        $_source = undef
+      }
+    } else { # RPM platforms: EL and SUSE
+      $_install_options = $install_options
+      if $::puppet_agent::absolute_source {
+        # absolute_source means we use rpm on EL/suse based platforms
+        $_package_version = 'present'
+        $_provider = 'rpm'
+        $_source = "${::puppet_agent::params::local_packages_dir}/${::puppet_agent::prepare::package::package_file_name}"
```
speedofdark: does the same comment from line 57 re: absolute source apply here as well?

manifests/install/solaris.pp
```diff
@@ -0,0 +1,53 @@
+# == Class puppet_agent::windows::install
+#
+# Private class called from puppet_agent class
+#
+# Manage the install process for windows specifically
```
speedofdark: windows or solaris?

manifests/install/solaris.pp
```diff
@@ -0,0 +1,53 @@
+# == Class puppet_agent::windows::install
```
speedofdark: windows or solaris?

manifests/install/darwin.pp
```diff
@@ -0,0 +1,26 @@
+# == Class puppet_agent::windows::install
```
speedofdark: windows or darwin?

manifests/install/darwin.pp
```diff
@@ -0,0 +1,26 @@
...
+#
+# Private class called from puppet_agent class
+#
+# Manage the install process for windows specifically
```
speedofdark: windows or darwin?

manifests/install/solaris.pp
```diff
@@ -0,0 +1,53 @@
...
+      $service_names = delete($::puppet_agent::service_names, 'mcollective')
+    } else {
+      $service_names = $::puppet_agent::service_names
+    }
+
+    # Puppet prior to 5.0 would not use a separate process contract when forking from the Puppet
+    # service. That resulted in service-initiated upgrades failing because trying to remove or
+    # upgrade the package would stop the service, thereby killing the Puppet run. Use a script
+    # to perform the upgrade after Puppet is done running.
+    # Puppet 5.0 adds this, but some i18n implementation is loading code fairly late and appearsx
```
speedofdark: appearsx -> appears

manifests/install.pp
McdonaldSeanp: Ah, actually the second comment is what's missing. In cases of absolute source debian/rhel/suse should be using prepare::package.

manifests/init.pp
McdonaldSeanp: Agreed, we found out that EL-5 machines will fail if the source is set to https. I'm also sad about insecure _but_ we do sign our packages. Hopefully even if someone did manage an attack the upgrade would still not do anything since any packages provided by an attacker wouldn't be signed.

README.markdown
McdonaldSeanp: No not really.

It's probably _technically_ possible, but a huge undertaking and certainly not in the scope of this module. 

What we want to support is direct mirrors of our sources, which honestly is probably all anyone needs.

manifests/install.pp
McdonaldSeanp: Yeah, I'll need to update the prepare steps for these platforms to use prepare::package

manifests/osfamily/aix.pp
```diff
@@ -44,10 +44,14 @@
...
-    $source = $::puppet_agent::source
+  if $::puppet_agent::absolute_source {
+    $source = $::puppet_agent::absolute_source
+  } elsif $::puppet_agent::alternate_pe_source {
+    $source = "${::puppet_agent::alternate_pe_source}/packages/${pe_server_version}/aix-${aix_ver_number}-power/${::puppet_agent::package_name}-${::puppet_agent::package_version}-1.aix${aix_ver_number}.ppc.rpm"
+  } elsif $::puppet_agent::source {
+    $source = "${::puppet_agent::source}/packages/${pe_server_version}/aix-${aix_ver_number}-power/${::puppet_agent::package_name}-${::puppet_agent::package_version}-1.aix${aix_ver_number}.ppc.rpm"
   } else {
-    $source = "puppet:///pe_packages/${pe_server_version}/aix-${aix_ver_number}-power/${::puppet_agent::package_name}-${::puppet_agent::package_version}-1.aix${aix_ver_number}.ppc.rpm"
+    $source = "${::puppet_agent::aix_source}/${pe_server_version}/aix-${aix_ver_number}-power/${::puppet_agent::package_name}-${::puppet_agent::package_version}-1.aix${aix_ver_number}.ppc.rpm"
```
speedofdark: How tough would it be to eliminate the measurable amount of duplication in the paths to the RPM? We appear to be changing the root, and the rest of the structure remains the same. Does Ruby take care of that for us?

This is a global question across all the constructs similar to this.

manifests/install/solaris.pp
```diff
@@ -0,0 +1,53 @@
...
+  assert_private()
+  if $::operatingsystemmajrelease == '10' {
+    $_unzipped_package_name = regsubst($::puppet_agent::prepare::package::package_file_name, '\.gz$', '')
+    $install_script = 'solaris_install.sh.erb'
+
+    # The following are expected to be available in the solaris_install.sh.erb template:
+    $adminfile = '/opt/puppetlabs/packages/solaris-noask'
+    $sourcefile = "/opt/puppetlabs/packages/${_unzipped_package_name}"
+    # Starting with puppet6 collections we no longer carry the mcollective service
+    if $::puppet_agent::collection != 'PC1' and $::puppet_agent::collection != 'puppet5' {
```
speedofdark: invert the if for puppet6 an option?

tasks/install_shell.sh
```diff
@@ -417,24 +435,24 @@ case $platform in
   "SLES")
     info "SLES platform! Lets get you an RPM..."
     gpg_key="${tmp_dir}/RPM-GPG-KEY-puppet"
-    do_download "https://yum.puppetlabs.com/RPM-GPG-KEY-puppet" "$gpg_key"
+    do_download "https://yum.puppet.com/RPM-GPG-KEY-puppet" "$gpg_key"
```
speedofdark: not ${yum_source}?

manifests/install/solaris.pp
McdonaldSeanp: We could do that, but then we would need to add each new collection as they are released. This way should be future proof (i.e. no changes required for puppet7/8/9)

tasks/install_shell.sh
McdonaldSeanp: Right now those keys only exist on the public downloads servers, which means you can't install from nightlies. For now this needs to stay a full URL until https://tickets.puppetlabs.com/browse/RE-12161 is resolved.

manifests/install/solaris.pp
speedofdark: That makes sense. Coolness.

manifests/osfamily/aix.pp
McdonaldSeanp: We just talked in person about this: there's enough variation in the calculation of the parts of the path (i.e. the platform name in PE paths needs to be changed for AIX/Fedora/Amazon) that trying to centralize the data won't look any more maintainable than this.

Unfortunately I think we should keep it the way it looks so it's more understandable down the road.

manifests/osfamily/debian.pp
```diff
@@ -5,117 +5,124 @@
 
     include ::apt
 
-    if getvar('::puppet_agent::is_pe') == true {
-      $pe_server_version = pe_build_version()
-      if $::puppet_agent::source {
-        $source = $::puppet_agent::source
+    if $::puppet_agent::absolute_source {
```
ragnarkon: Seems like this `if` needs to move outside the `if getvar('::puppet_agent::manage_repo') == true` on line 4.  (Should be similar to the suse.pp & redhat.pp manifests.)

McdonaldSeanp: Oh, whoops :) Didn't notice the if on line for.

manifests/osfamily/debian.pp
```diff
@@ -1,121 +1,127 @@
...
+    # Absolute sources are expected to be actual packages (not repos)
+    # so when absolute_source is set just download the package to the
+    # system and finish with this class.
+    $source = $::puppet_agent::absolute_source
+    class { '::puppet_agent::prepare::package':
+      source => $source,
+    }
+    contain puppet_agent::prepare::package
+  } else {
+    if getvar('::puppet_agent::manage_repo') == true {
```
ragnarkon: Mixed feelings about putting the if there, and I'm not sure what is the "right" answer is in this case... so I'll just point out the differences.

For Debian this module effectively won't work with a repo at all unless `manage_repo = true`. Which seems a bit heavy handed (what if the user wants to use their own repo?).

manifests/osfamily/redhat.pp
```diff
@@ -43,89 +22,124 @@
...
-      enabled             => true,
-      gpgcheck            => '1',
-      gpgkey              => "${gpg_keys}",
-      proxy               => $_proxy,
-      sslcacert           => $_sslcacert_path,
-      sslclientcert       => $_sslclientcert_path,
-      sslclientkey        => $_sslclientkey_path,
-      skip_if_unavailable => $::puppet_agent::skip_if_unavailable,
+
+    if $::puppet_agent::manage_repo == true {
```
ragnarkon: For RHEL, the module will setup all of the necessary gpg keys to add to the repo, but the `manage_repo` if statement is way at the bottom and only encapsulates the `yumrepo` resource. If effectively does all of the necessary setup to manage the repo except for the very last step.

Is there a reason to setup all of the GPG keys if we are not managing the repo?? Not sure.

manifests/osfamily/suse.pp
```diff
@@ -5,104 +5,119 @@
...
+          require   => File[$gpg_path],
+          logoutput => 'on_failure',
+        }
 
-      if getvar('::puppet_agent::manage_repo') == true {
-        # Set up a zypper repository by creating a .repo file which mimics a ini file
-        $repo_file = '/etc/zypp/repos.d/pc_repo.repo'
-        $repo_name = 'pc_repo'
-        $agent_version = $::puppet_agent::package_version
+        if getvar('::puppet_agent::manage_repo') == true {
```
ragnarkon: SUSE is similar to RHEL. Sets up the keys but stops at the last mile.

Again, not sure if the keys are necessary if we aren't managing the repo—unless I am completely missing something.

All of this `manage_repo` business is admittedly somewhat out-of-scope for MODULES-6604. But it is something that has bothered me for a long time about this module.  (I'd be glad to submit another PR in the future if makes more sense to tackle it later.)

McdonaldSeanp:
@ragnarkon I am totally on-board that the manage repo business is not well enough thought out. But I do think it is out of scope for this PR since the work here is specific to adding params for sources. I tried to keep all other functionality the same and backwards compatible.

FWIW for any future work in this space: personally I think it makes sense to go with what happens with the debian case and just have nothing happen if manage_repo is false. If you were managing the repo outside of this module I think it makes more sense that this module simply does nothing for the repo?

In that case (and I think this is what happens for debian now) the only thing this module will effectively do is specify the file resource for the agent to install in install.pp. In that case users can properly subscribe the puppet_agent class to their own repo management class(es) and then let the puppet_agent class take care of the actual package install.

That's just my opinion though, if you totally disagree please feel free to open up changes in a new PR that are completely different :)

PR #399, opened 2019-04-03
installdir parameter not defined in install_shell task

kkarthee:
installdir parameter not defined in install_shell task, Due to that `install_shell` task is failing with the following error. 

```
Error: Task finished with exit-code 1 
This module depends on the puppetlabs-facts module
```

### DIFF
--- a/tasks/install_shell.json
+++ b/tasks/install_shell.json
@@ -30,6 +30,10 @@
     "install_options": {
       "description": "optional install arguments to the windows installer (defaults to REINSTALLMODE=\"amus\")",
       "type": "Optional[String]"
+    },
+    "installdir": {
+      "description": "optional install arguments -  installer install directory ",
+      "type": "Optional[String]"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


donoghuc:
The `_installdir` is a task metaparameter used by `bolt` for tasks that require multiple files for execution. IN this case, the `install_shell` task requires files https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/3aef2ffbb731cdec1baaa50d0eb48561257171e3/tasks/install.json#L34 provided by the https://github.com/puppetlabs/puppetlabs-facts facts module.

donoghuc:
Also, if you were trying to run the `puppet_agent::install_shell` task directly, there was likely an error due to https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/439 . Thanks for submitting this issue.

joshcooper:
@donoghuc it sounds like this PR is not the right fix? Can it be closed?

joshcooper:
@kkarthee thanks for submitting this issue, but based on @donoghuc's comment, I don't think this PR is needed and am going to close. Please reopen if you can reproduce the issue with the latest of bolt.

PR #400, opened 2019-04-10
Fix Yum URL path for RedHat systems still using the PC1 collection

coreone:
Currently, if you have a RedHat-based machine and you specify the **PC1** collection, you get a yum repository configuration with the `source` set to:

http://yum.puppet.com/PC1/el/6/x86_64

However, the correct URL should be:

http://yum.puppet.com/el/6/PC1/x86_64

This PR fixes the `source` to use the correct URL if **PC1** is the collection being used.

### DIFF
--- a/manifests/osfamily/redhat.pp
+++ b/manifests/osfamily/redhat.pp
@@ -37,7 +37,11 @@ class puppet_agent::osfamily::redhat{
         $source = "https://${::puppet_master_server}:8140/packages/${pe_server_version}/${pe_repo_dir}"
       }
     } else {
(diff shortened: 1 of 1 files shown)


coreone:
I believe this was a trivial contribution as it was literally a copy/paste of the original code, and closely mimics previous code in the history, so this can pretty much be considered a code reversion.

As a side note to this process:
* The URL to the documentation should be updated to https://puppet.com/community/trivial-patch-exemption-policy as the one sent is a dead link
* The certificate on the CLA site has expired

mihaibuzgau:
👍 Thanks, @coreone for the contribution

PR #401, opened 2019-04-11
(MODULES-8923) autodetect package_version based upon the master

tkishel:
With this commit, the user can specify `package_version` as 'master'
to specify that the package version should be the same as the compiling master.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,7 +23,8 @@
 # [package_name]
 #   The package to upgrade to, i.e. `puppet-agent`.
 # [package_version]
-#   The package version to upgrade to. Explicitly specify a version to upgrade
+#   The package version to upgrade to. Explicitly specify the version to upgrade to,
+#   or set to 'auto' to specify the version of the compiling master.
 # [service_names]
 #   An array of services to start, normally `puppet` and `mcollective`.
 #   None will be started if the array is empty.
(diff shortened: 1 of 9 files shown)


manifests/init.pp
```diff
@@ -131,30 +132,36 @@
...
-      fail("invalid version ${package_version} requested")
+    if $::architecture == 'x86' and $arch == 'x64' {
+      fail('Unable to install x64 on a x86 system')
     }
 
-    # Strip git sha from dev builds
-    if $package_version =~ /g/ {
-      $_expected_package_version = split($package_version, /[.-]g.*/)[0]
+    if $package_version == 'master' {
+      $master_or_package_version = $serverversion
```
McdonaldSeanp: I don't know where `$serverversion` comes from

tkishel: > $serverversion — the current version of Puppet on the Puppet master.

https://puppet.com/docs/puppet/5.5/lang_facts_and_builtin_vars.html#puppet-master-variables

Like `pe_compiling_server_aio_build` in `params.pp` but it does not require PE for ...

https://github.com/puppetlabs/puppet-enterprise-modules/blob/irving/modules/puppet_enterprise/lib/puppet/parser/functions/pe_compiling_server_aio_build.rb

and

https://github.com/puppetlabs/puppet-enterprise-modules/blob/irving/modules/puppet_enterprise/lib/shared/aio_build.rb

McdonaldSeanp: Ah, cool. Thank you!

tkishel:
NOTE: The README will need to be updated if this is merged. I didn't do that in this PR, as per my comment in MODULES-8923 regarding {{auto}} vs {{master}}.

manifests/init.pp
MikaelSmith: I think there was a period in time when puppet version and aio package version weren't guaranteed to match, but I believe they should for recent and future versions.

oanatmaria:
So, pipeline passed: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/puppetlabs-puppet_agent%20module/view/ad-hoc/job/forge-module_puppetlabs-puppet-agent-module_intn-sys_smoke-adhoc/32/
I tested with AIX and it works as expected.

PR #414, opened 2019-05-24
puppet_agent::service: option to not start the service on puppet run

makmarius:
We needed a way to not start puppet on testing other environments.

We usually run in `production` but from time to time we need to stop puppet and use `--environment foobar` to check things, where we might need some time to investigate changes. Starting puppet automatically here is not what we want.

If this is already possible, please gladly point me in the direction where I can find the documentation/code!

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -3,8 +3,16 @@
 # This class is meant to be called from puppet_agent.
 # It ensures that managed services are running.
 #
-class puppet_agent::service {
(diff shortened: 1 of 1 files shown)


GabrielNagy:
Hi @makmarius, sorry for the _really_ late answer...

This should be possible by passing `service_names => []` to the puppet_agent class.

makmarius:
No worries! Thank you for your answer 🙇🏼

PR #415, opened 2019-05-28
Add Debian package pin to stop the version changing on apt upgrade

ddryden:
When on Debian and a package version is specified we should also create a Debian package pin to stop `apt upgrade` from changing the version just for the next puppet run to change it back.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -62,6 +62,11 @@ class puppet_agent::install(
           $_package_version = $package_version
         } else {
           $_package_version = "${package_version}-1${::lsbdistcodename}"
+          apt::pin { "pin ${::puppet_agent::package_name} package":
+            packages => $::puppet_agent::package_name,
+            priority => 1000,
+            version  => $_package_version,
+          }
         }
         $_provider = 'apt'
         $_source = undef


manifests/install.pp
```diff
@@ -62,6 +62,11 @@
           $_package_version = $package_version
         } else {
           $_package_version = "${package_version}-1${::lsbdistcodename}"
+          apt::pin { "pin ${::puppet_agent::package_name} package":
+            packages => $::puppet_agent::package_name,
+            priority => 1000,
```
ciprianbadescu: In order to support downgrade, my understanding is that priority should be 1001: https://linux.die.net/man/5/apt_preferences

seoester:
How is this PR going?
Is the requested change the only thing blocking the merge?
The CI seems to be failing because of missing dependencies in the test suite.

I'd love to see this merged & released :)

ciprianbadescu:
@seoester,  there is small change needed from  @ddryden

seoester:
@ciprianbadescu thanks for the quick response.

Would you mind if I replicated the PR and perform the change to speed things along?

ciprianbadescu:
@seoester, would be OK. Pls also check https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/CONTRIBUTING.md

ddryden:
Sorry for being slow to respond. 

Your right 1001 would be a better priority.

I don't know my way around the test's could you link some documentation?

https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/CONTRIBUTING.md seems to have a _Testing_ section listed in the table of contents but does not exist in the document.

seoester:
@ddryden, I replicated your PR and we have been discussing the changes in #441.

joshcooper:
@seoester Our puppetN-release packages on apt.puppet.com now install a preferences file (see https://github.com/puppetlabs/puppetlabs-release/commit/ecf2c8526e4cebc6229dd9c27fc1c7db25bba730#diff-4fbc9b1630d785c70cd453b814c99fa693faae74aabf983473e85ef200b3b32d) This was done because Debian and Ubuntu have started publishing `puppet-agent` packages with the same name as ours. Given that I'm thinking this is no longer needed?

seoester:
@joshcooper, thanks for the ping after so much time.

The pinning you are referencing pins the Apt Source/Repo of the puppet-agent package.
The pinning proposed in this PR (and my PR #441) pins the specific version of the puppet-agent package, so that `apt upgrade` does not unintentionally bump the package to a new version.

I believe this specific PR can be closed in any case. That's because the team wanted to go a different route, using a new feature of the dpkg provider rather than apt pinning: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/441#issuecomment-549810369

joshcooper:
> I believe this specific PR can be closed in any case. 

@seoester thanks for letting us know.

PR #417, opened 2019-06-11
(MODULES-9173) Mcollective service restarting when PA upgrade is done.

loopinu:
This PR fixes the following:

When and upgrade is done for solaris 10 we enable the service only if the service was running before.

### DIFF
--- a/templates/solaris_install.sh.erb
+++ b/templates/solaris_install.sh.erb
@@ -1,13 +1,18 @@
 function start_service() {
   service="${1:?}"
-  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running
+  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running enable=true
 }
-service_was_running_before_agent_update = false
-if /opt/puppetlabs/bin/puppet resource service "${service_name:?}" | grep "ensure => 'running'" ; then
-   service_was_running_before_agent_update = true
+
+<% @service_names.each do |service_name| -%>
+<%= service_name %>_was_running = false
+if /opt/puppetlabs/bin/puppet resource service "<%= service_name %>" | grep "ensure => 'running'" ; then
+  <%= service_name %>_was_running = true
 fi
+<% end -%>
+
 # Remove old package
 /opt/puppetlabs/bin/puppet resource package puppet-agent ensure=absent adminfile=<%= @adminfile %>
+
 # Install package
 pkgadd -a <%= @adminfile %> -d <%= @sourcefile %> -G <%= @install_options.join(' ') %> -n puppet-agent
 
@@ -15,8 +20,8 @@ pkgadd -a <%= @adminfile %> -d <%= @sourcefile %> -G <%= @install_options.join('
 # services on its own since that only happens when the service manifests change, which is
 # highly unlikely in most agent upgrade scenarios.
 
-<% @service_names.each do |service_name| %>
-  if $service_was_running_before_agent_update ; then
-    start_service <%= service_name %>
-  fi
-<% end %>
+<% @service_names.each do |service_name| -%>
+if $<%= service_name %>_was_running ; then
+  start_service <%= service_name %>
+fi
+<% end -%>
--- a/templates/solaris_install.sh.erb
+++ b/templates/solaris_install.sh.erb
@@ -1,11 +1,13 @@
 function start_service() {
   service="${1:?}"
-  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running enable=true
+  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running
 }
-
+service_was_running_before_agent_update = false
+if /opt/puppetlabs/bin/puppet resource service "${service_name:?}" | grep "ensure => 'running'" ; then
+   service_was_running_before_agent_update = true
+fi
 # Remove old package
 /opt/puppetlabs/bin/puppet resource package puppet-agent ensure=absent adminfile=<%= @adminfile %>
-
 # Install package
 pkgadd -a <%= @adminfile %> -d <%= @sourcefile %> -G <%= @install_options.join(' ') %> -n puppet-agent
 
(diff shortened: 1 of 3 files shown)


templates/solaris_install.sh.erb
```diff
@@ -1,6 +1,8 @@
 function start_service() {
   service="${1:?}"
-  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running enable=true
+if /opt/puppetlabs/bin/puppet resource service "${service:?}" | grep "ensure => 'running'" ; then
```
MikaelSmith: I'm not sure how this change accomplishes the described outcome.

If the service is already running, then ensure=running does nothing. If it was not running, then how do you know it was running before? Isn't that something we should check before the upgrade?

Also @ekinanp what was here before doesn't make sense to me; the comment suggests we should be doing a service restart but the command here will only start the service if it was stopped.

ekinanp: Solaris upgrades are weird, I think we stopped the service prior to doing the upgrade. The service should be restarted in https://github.com/puppetlabs/vanagon/blob/master/resources/solaris/10/postinstall.erb#L39-L42 -- think this is just a sanity check.

I don't remember why we did this, I just remember it being a part of https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/1.5.0/templates/solaris_install.sh.epp prior to my refactor.

MikaelSmith: Found where I'm responsible for originally writing this :)

My suggestion is move the test before "Remove old package", then start services based on that test.

loopinu: @ekinanp thanks for explaining, it's exactly what's intended to do, a sanity check. 👍 
@MikaelSmith I hope I understood this correctly, please have a look again.

templates/solaris_install.sh.erb
```diff
@@ -14,5 +13,7 @@ pkgadd -a <%= @adminfile %> -d <%= @sourcefile %> -G <%= @install_options.join('
 # highly unlikely in most agent upgrade scenarios.
 
 <% @service_names.each do |service_name| %>
-start_service <%= service_name %>
+  if /opt/puppetlabs/bin/puppet resource service "${service_name:?}" | grep "ensure => 'running'" ; then
```
MikaelSmith: So if already running, trigger a restart? That seems acceptable, although that's not what this code is doing.

templates/solaris_install.sh.erb
```diff
@@ -1,8 +1,7 @@
 function start_service() {
   service="${1:?}"
-  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running enable=true
+  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running restart=true
```
MikaelSmith: I think you've misunderstood what the `restart` option does. https://puppet.com/docs/puppet/6.4/types/service.html#service-attribute-restart

In this case nothing happens. If there were a **refresh event**, it would attempt to restart using the command `true` (which would do nothing, but do it successfully), but I don't think there's a way to trigger a refresh event using the `puppet resource` command. I think this has to be `puppet resource service ... ensure=stopped; puppet resource service ... ensure=running`.

I'd be tempted to directly use the Solaris service management commands because we use this script for a single platform (Solaris 10).

templates/solaris_install.sh.erb
MikaelSmith: I'm also not 100% sure that the install will always leave the services running. That's why my suggestion was to do this test (and save it to a variable) before the `pkgadd` operation.

loopinu: @MikaelSmith You where right, it wouldn't start the service all the time.
I just moved the if before removing the puppet_agent package to see if the service was running.

Tested this and it works as intended.

templates/solaris_install.sh.erb
```diff
@@ -1,11 +1,13 @@
 function start_service() {
   service="${1:?}"
-  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running enable=true
+  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running
 }
-
+service_was_running_before_agent_update = false
+if /opt/puppetlabs/bin/puppet resource service "${service_name:?}" | grep "ensure => 'running'" ; then
```
MikaelSmith: Where does this service_name variable come from? I think you're confusing shell syntax with erb templating. This is definitely getting a little complicated.

I think this would work
```
<% @service_names.each do |service_name| %>
  <%= service_name >_was_running = false
  if /opt/puppetlabs/bin/puppet resource service "<%= service_name>" | grep "ensure => 'running'" ; then
    <%= service_name >_was_running = true
  fi
<% end %>
```

then below
```
<% @service_names.each do |service_name| %>
  if $<%= service_name %>_was_running ; then
    start_service <%= service_name %>
  fi
<% end %>
```

templates/solaris_install.sh.erb
```diff
@@ -1,11 +1,13 @@
 function start_service() {
   service="${1:?}"
-  /opt/puppetlabs/bin/puppet resource service "${service:?}" ensure=running enable=true
```
MikaelSmith: Why remove `enable=true`?

templates/solaris_install.sh.erb
MikaelSmith: I think we want to update the spec tests to have some positive validation of what's produced too.

MikaelSmith: I can handle the spec test bits. Working on what that would look like.

MikaelSmith: Maybe something like https://github.com/puppetlabs/puppetlabs-puppet_agent/compare/master...MikaelSmith:MODULES-9173.

templates/solaris_install.sh.erb
loopinu: @MikaelSmith sorry for the delay I was on PTO. Can you check again please?

templates/solaris_install.sh.erb
MikaelSmith: I think the only open question I had was whether something's shown we need to actually restart the service rather than ensure running. Is it always stopped after a package upgrade?

PR #419, opened 2019-07-02
(MODULES-8921) Implement puppet_agent::run task

reidmv:
This provides a task to perform on-demand runs of the Puppet agent, with the ability to pass command-line options.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -43,6 +43,7 @@
     - [Tasks](#tasks)
       - [`puppet_agent::version`](#puppet_agentversion)
       - [`puppet_agent::install`](#puppet_agentinstall)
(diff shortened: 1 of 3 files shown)


adreyer:
We've handled a lot of complexity around automating an agent run in the pxp module I think that's the best starting point for this functionality. I think it can be more or less copied and the metadata converted.

https://github.com/puppetlabs/pxp-agent/blob/master/modules/pxp-module-puppet

genebean:
Curious about the status of this, any update?

adreyer:
I think a PR for this should be a port of the pxp-agent functionality since that is well tested for this use case.

PR #426, opened 2019-07-29
(MODULES-7760) relax agent versions on redhat

gimmyxd:


gimmyxd:
For RedHat we the puppet_agent module is creating a `yum.repo` that is installed on the agent and provides the available puppet-agent agent from master.
This changes allow user to specify more flexible versions that can be installed, example: `x.y.z.el7`
It will only check that the version starts as a `x.y.z` which will be compared with the already installed versions, which is reported also as a `x.y.z.`

manifests/init.pp
```diff
@@ -142,13 +142,22 @@
       $master_or_package_version = $package_version
     }
 
-    if $master_or_package_version !~ /^\d+\.\d+\.\d+([.-]?\d*|\.\d+\.g[0-9a-f]+)$/ {
-      fail("invalid version ${master_or_package_version} requested")
+    if $::osfamily == 'redhat' {
+      if $master_or_package_version !~ /^\d+\.\d+\.\d+(.+)?$/ {
```
GabrielNagy: you can replace `(.+)?` with `.*` to match everything to the end of the line, since we don't need capture groups here

PR #432, opened 2019-08-08
(MODULES-9700) - Cached catalog fix for Windows

ragnarkon:
Adds a conditional to the exec resource responsible for executing the Puppet agent install script on Windows.

Normally the module determines if an install is required or not by using the `aio_agent_version` fact. However, this fact will not be updated for nodes using a cached catalog, causing the module to continually trigger the installation script.

### DIFF
--- a/manifests/install/windows.pp
+++ b/manifests/install/windows.pp
@@ -62,6 +62,32 @@ class puppet_agent::install::windows(
                           -InstallArgs '${_install_options}' \
                           ${_move_dll_workaround}",
     path    => $::path,
+    # This unless is necessary to handle nodes using a cached catalog and removes reliance on the aio_agent_version fact.
+    unless  => "${::system32}\\WindowsPowerShell\\v1.0\\powershell.exe \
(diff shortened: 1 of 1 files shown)


ragnarkon:
Tested on Windows 2012 & Windows 2016. I don't have access to 2008 or a 2019 box, but it don't see any reason why it wouldn't work on those platforms—unless there is some oddities with Powershell 2 on Win 2008 that I'm not aware of.

The logic is largely the same logic used in the [version task](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/tasks/version_powershell.ps1). Opted to use that logic rather than crawling through `HKLM\Software\Microsoft\Windows\CurrentVersion\Uninstall` or using the `Win32_Product` WMI class.

gimmyxd:
@ragnarkon thanks for the contribution! 
Gonna close this as it was fixed in: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/437

Let us know if you encounter any issues.

PR #433, opened 2019-08-13
(MODULES-9698) Update facts module used for testing install task

donoghuc:
Update the version of puppetlabs-facts to match what is shipped with the latest bolt version.


donoghuc:
https://github.com/puppetlabs/bolt/pull/1146

PR #434, opened 2019-08-15
Linux Mint Support

luckyraul:


tasks/install_shell.sh
```diff
@@ -483,6 +483,26 @@ case $platform in
...
+  "Linuxmint")
+    info "Mint platform! Lets get you a DEB..."
+    case $major_version in      
+      "3")  deb_codename="stretch";;
+    esac
+    filetype="deb"
+    filename="${collection}-release-${deb_codename}.deb"
+    download_url="http://apt.puppetlabs.com/${filename}"
+    ;;
+  "LinuxMint")
```
ciprianbadescu: can you compact both cases in only one with "Linuxmint"|"LinuxMint" as pattern?

luckyraul: done

tasks/install_shell.sh
```diff
@@ -483,6 +483,18 @@ case $platform in
     filename="${collection}-release-${deb_codename}.deb"
     download_url="${apt_source}/${filename}"
     ;;
+  "Linuxmint"|"LinuxMint")
+    info "Mint platform! Lets get you a DEB..."
+    case $major_version in
+      "3")  deb_codename="stretch";;
```
donoghuc: I do not have a way to test this, is "3" what the $major_version will resolve to? What version of the `facts` module are you using?

tasks/install_shell.sh
```diff
@@ -483,6 +483,18 @@ case $platform in
...
+      "19") deb_codename="bionic";;
+      "18") deb_codename="xenial";;
+      "17") deb_codename="trusty";;
+    esac
+    filetype="deb"
+    filename="${collection}-release-${deb_codename}.deb"
+    download_url="http://apt.puppetlabs.com/${filename}"
```
nicklewis: This should use `${apt_source}/${filename}` as above.

tasks/install_shell.sh
luckyraul: 3 is LDME
18, 19 is LinuxMint

PR #440, opened 2019-09-24
(maint) Add bolt team as codeowners for task content

donoghuc:


donoghuc:
I also added both bolt team and nights watch team to `teams` so that PRs can trigger review requests

PR #441, opened 2019-09-27
Pin Debian package to prevent unintentional upgrade

seoester:
This is a revised version of #415.

Original description (@ddryden)
> When on Debian and a package version is specified we should also create a Debian package pin to stop apt upgrade from changing the version just for the next puppet run to change it back.

/cc @ciprianbadescu

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -62,6 +62,7 @@ class puppet_agent::install(
           $_package_version = $package_version
         } else {
           $_package_version = "${package_version}-1${::lsbdistcodename}"
+          include ::apt
           apt::pin { "pin ${::puppet_agent::package_name} package":
             packages => $::puppet_agent::package_name,
             priority => 1001,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -64,7 +64,7 @@ class puppet_agent::install(
           $_package_version = "${package_version}-1${::lsbdistcodename}"
           apt::pin { "pin ${::puppet_agent::package_name} package":
             packages => $::puppet_agent::package_name,
-            priority => 1000,
+            priority => 1001,
             version  => $_package_version,
           }
         }
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -62,6 +62,11 @@ class puppet_agent::install(
           $_package_version = $package_version
         } else {
           $_package_version = "${package_version}-1${::lsbdistcodename}"
+          apt::pin { "pin ${::puppet_agent::package_name} package":
+            packages => $::puppet_agent::package_name,
+            priority => 1000,
+            version  => $_package_version,
+          }
         }
         $_provider = 'apt'
         $_source = undef
--- a/spec/classes/puppet_agent_osfamily_debian_spec.rb
+++ b/spec/classes/puppet_agent_osfamily_debian_spec.rb
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 describe 'puppet_agent' do
+  package_name = 'puppet-agent'
+
   facts = {
     :lsbdistid            => 'Debian',
     :osfamily             => 'Debian',
@@ -218,6 +220,30 @@ describe 'puppet_agent' do
       it { is_expected.not_to contain_apt__source('pc_repo') }
     end
 
+    context "when requesting a specific version" do
+      let(:params) {
+        {
+          :package_version => package_version
+        }
+      }
+
+      it { is_expected.to contain_apt__pin("pin #{package_name} package").with({
+        'packages' => package_name,
+        'priority' => 1001,
+        'version'  => "#{package_version}-1#{facts[:lsbdistcodename]}",
+      }) }
+    end
+
+    context "when not requesting a specific version" do
+      let(:params) {
+        {
+          :package_version => 'auto'
+        }
+      }
+
+      it { is_expected.not_to contain_apt__pin('pin #{package_name} package') }
+    end
+
     it { is_expected.to contain_class("puppet_agent::osfamily::debian") }
 
   end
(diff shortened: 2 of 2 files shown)


ciprianbadescu:
@seoester, can you fix the tests?

seoester:
@ciprianbadescu I missed your comment edits and fixed the tests as well.
Tests for the pinning behaviour are still missing. I would appreciate some guidance on how to implement the tests.

I believe pinning should occur iff a specific version is passed:

 * `$version` is `"present"` => no pinning, `apt upgrade` will install latest version available in repo.
 * otherwise => pin at `$version` (whether the repo is managed or not).

Considering this, the tests would have to be orthogonal to the existing `context` structure. I could duplicate and wrap the entire content of "when PE" and "when FOSS" in further `context` blocks: "when specific version" and "when no specific version". But this would result in a lot of copied code.

Alternatively, I could add the checks as separate `context` blocks in "when PE" and "when FOSS". I did so in the latest commit https://github.com/seoester/puppetlabs-puppet_agent/commit/0330f316b1920390ae34aee6e0d849b41ed9a552.

seoester:
I apparently misunderstood the `$version` parameter.
`puppet_agent::install` accepts `$package_version` and handles the value `"present"` specifically. If `"present"` is passed, it installs the package without passing version info to APT. However, `"present"` is never passed from `puppet_agent::init` to `puppet_agent::install`.

So there is no way of installing puppet-agent with the latest version according to APT.
I guess that means the package version should always be pinned.
Removing `"present"` from `puppet_agent::install` might make this even clearer.

Any opinions?

ciprianbadescu:
@soester, the added code in install.pp is already for the case when version is specified: https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/c58b00bb3845f938477b66d1acc41c16e5959d17/manifests/install.pp#L61-L70

ciprianbadescu:
@seoester, I recommend you that you cherrypick my commit https://github.com/ciprianbadescu/puppetlabs-puppet_agent/commit/424b8c8f98a57873c17b24a761814e2b6d5bf55c

The changes I did:
- pin version only in case `::puppet_agent::manage_repo` is `true`
- extend existing tests with `apt__pin` checks

soester:
> @soester, the added code in install.pp is already for the case when version is specified as the new code is in `else` branch :
> 
> https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/c58b00bb3845f938477b66d1acc41c16e5959d17/manifests/install.pp#L61-L70

@seoester

seoester:
> @seoester, I recommend you that you cherrypick my commit [ciprianbadescu@424b8c8](https://github.com/ciprianbadescu/puppetlabs-puppet_agent/commit/424b8c8f98a57873c17b24a761814e2b6d5bf55c)
> 
> The changes I did:
> 
> * pin version only in case `::puppet_agent::manage_repo` is `true`
> * extend existing tests with `apt__pin` checks

I believe the version should also be pinned if the repo is not managed. What is your rationale behind not pinning when `manage_repo` is `true`?

> @soester, the added code in install.pp is already for the case when version is specified as the new code is in `else` branch :

`"present"` is never passed as `$package_version` to `::puppet_agent::install`, so the `if` statement is superfluous. Should it be removed to enhance clarity?

ciprianbadescu:
> > @seoester, I recommend you that you cherrypick my commit [ciprianbadescu@424b8c8](https://github.com/ciprianbadescu/puppetlabs-puppet_agent/commit/424b8c8f98a57873c17b24a761814e2b6d5bf55c)
> > The changes I did:
> > 
> > * pin version only in case `::puppet_agent::manage_repo` is `true`
> > * extend existing tests with `apt__pin` checks
> 
> I believe the version should also be pinned if the repo is not managed. What is your rationale behind not pinning when `manage_repo` is `true`?
If manage_repo is false we should not use apt module to do apt management (and pinning in part of the module)

> 
> > @soester, the added code in install.pp is already for the case when version is specified as the new code is in `else` branch :
> 
> `"present"` is never passed as `$package_version` to `::puppet_agent::install`, so the `if` statement is superfluous ([#441 (comment)](https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/441#issuecomment-536642813)). Should it be removed to enhance clarity?
I suppose `present` will be the default value due to: https://github.com/ciprianbadescu/puppetlabs-puppet_agent/blob/424b8c8f98a57873c17b24a761814e2b6d5bf55c/manifests/install.pp#L11

@gimmyxd , any opinion?

seoester:
> > ...
> 
> If manage_repo is false we should not use apt module to do apt management (and pinning in part of the module)

I don't think the value of `manage_repo` should have an effect on the separate `package_version` parameter. So far, it hasn't. As a user, I would be surprised to find out that setting `manage_repo` to `false` would unpin the version.
No side effects come from including the `::apt` class.

> > ...
>
> I suppose `present` will be the default value due to: https://github.com/ciprianbadescu/puppetlabs-puppet_agent/blob/424b8c8f98a57873c17b24a761814e2b6d5bf55c/manifests/install.pp#L11
> 
> @gimmyxd , any opinion?

That is the problem exactly. Because it is the default, I assumed it had some significance. **But** `$package_version` is always passed by `init.pp` and because `::puppet_agent::install` is declared, there is no way to override its value.

https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/7cc847fc4989d68bc59336faabecac9475b42ea9/manifests/init.pp#L179-L183

gimmyxd:
@seoester we need to investigate this a bit more to make sure that the behaviour is correct in all cases. At a first look seems that you are right about `package_version`. We will check to see how `manage_repo` impacts the pinning of a version.

seoester:
@gimmyxd, thanks for getting back on this.

Could you post updates in this thread?

This change makes the module's behaviour more intuitive and leads to cleaner code. Inclusion in a near-future release would be much appreciated :)

gimmyxd:
@seoester sorry for the late reply.
We are working on `held` functionality for puppet dpkg provider which will help us unblock this issue: https://github.com/puppetlabs/puppet/pull/7773

We can't move forward with the current solution using `apt:pin` because that functionality does not play well with how this modules works for doing upgrades.

seoester:
Thanks for the update.

> @seoester sorry for the late reply.
> We are working on `held` functionality for puppet dpkg provider which will help us unblock this issue: [puppetlabs/puppet#7773](https://github.com/puppetlabs/puppet/pull/7773)

Alright :+1: The dpkg provider is part of Puppet itself... Will this module running with the non-latest Puppet version be able to provide pinning functionality?

> We can't move forward with the current solution using `apt:pin` because that functionality does not play well with how this modules works for doing upgrades.

The documentation of this module does not talk about updates. Could you outline the behaviour?

gimmyxd:
> Alright 👍 The dpkg provider is part of Puppet itself... Will this module running with the non-latest Puppet version be able to provide pinning functionality?
No, that functionality will be only in newer puppet version

> The documentation of this module does not talk about updates. Could you outline the behaviour?
What do you mean? the purpose of this module is to upgrade puppet agents.

PR #445, opened 2019-10-14
(maint) Make the puppet_agent task available as a plugin

adreyer:


adreyer:
note there is a bug in the puppet_library plugin implementation that prevents this from working now.

donoghuc:
@mihaibuzgau Can we get this merged and released with https://tickets.puppetlabs.com/browse/MODULES-9954 this week? We would like to ship this with bolt system packages.

mihaibuzgau:
@donoghuc it's unlikely that this will be released this week. We've planned to release this last sprint but because of some unforeseen issues, we couldn't.
We're OK if your team wants to release this module to the forge.

donoghuc:
@mihaibuzgau Not a problem. We can target next week. WIll check in early next week with you about who can work on it. Thanks!

PR #448, opened 2019-10-23
(MODULES-27043) Puppet agent upgrade for windows FIPS

Dorin-Pleava:
This commit ensures that the correct version of puppet agent is
installed on windows platforms that are FIPS compliant based on the
enabled_fips fact.

### DIFF
--- a/manifests/osfamily/windows.pp
+++ b/manifests/osfamily/windows.pp
@@ -8,8 +8,10 @@ class puppet_agent::osfamily::windows{
   } elsif  ($::puppet_agent::is_pe and (!$::puppet_agent::use_alternate_sources)) {
     $pe_server_version = pe_build_version()
     $tag = $::puppet_agent::arch ? {
-      'x64' => 'windows-x86_64',
-      'x86' => 'windows-i386',
+      'x64' => $::fips_enabled ? {
+          true => 'windowsfips-x64',
+          default => 'windows-x86_64' },
+      'x86' => 'windows-i386'
     }
     if $::puppet_agent::alternate_pe_source {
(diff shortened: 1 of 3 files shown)


manifests/osfamily/windows.pp
```diff
@@ -8,8 +8,12 @@
...
     $pe_server_version = pe_build_version()
     $tag = $::puppet_agent::arch ? {
-      'x64' => 'windows-x86_64',
-      'x86' => 'windows-i386',
+      'x64' => $::fips_enabled ? {
+          true => 'windowsfips-x64',
+          default => 'windows-x86_64' },
+      'x86' => $::fips_enabled ? {
+          true => 'windowsfips-i386',
+          default => 'windows-i386' },
```
gimmyxd: We should test that based `fips_enabled` flag the correct package is set in the manifest

manifests/osfamily/windows.pp
```diff
@@ -8,8 +8,12 @@
   } elsif  ($::puppet_agent::is_pe and (!$::puppet_agent::use_alternate_sources)) {
     $pe_server_version = pe_build_version()
     $tag = $::puppet_agent::arch ? {
-      'x64' => 'windows-x86_64',
-      'x86' => 'windows-i386',
+      'x64' => $::fips_enabled ? {
+          true => 'windowsfips-x64',
+          default => 'windows-x86_64' },
+      'x86' => $::fips_enabled ? {
```
gimmyxd: This will not work as we are not providing windowsfips x86 packages, we should let it install `windows-i386` no matter the `fips_enabled` flag

PR #449, opened 2019-10-30
(maint) Updated metadata.json with support for Windows Server 2019

ducksecops:
Add WIndows Server 2019 to list of supported operating. Currently using on this OS


gimmyxd:
@ducksecops please update your commit message to start with a `(MODULES)` ticket number of `maint` for trivial changes

gimmyxd:
@ducksecops to be valid, the commit message should be something like:
`(maint) Updated metadata.json with support for Windows Server 2019`

you can do this by amending you current commit message:
`git commit --amend -m "(maint) Updated metadata.json with support for Windows Server 2019"`

and then force push to your branch.

We have a check that fails if the commit message does not match specific rules.

PR #450, opened 2019-10-31
(MODULES-10038) nightly build download location fix

Dorin-Pleava:
This commit fixes the path from where to download puppet-agent if the
requested collections is a nightly build.
If the windows_source is explicitly specified as a parameter, it will
use it instead.

### DIFF
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -31,6 +31,11 @@ else {
     $msi_name = "puppet-agent-${arch}-latest.msi"
 }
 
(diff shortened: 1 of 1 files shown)


gimmyxd:
@donoghuc that would be a breaking

gimmyxd:
@donoghuc that would be a bigger change and we would need to do it also for install_shell.sh to be consistent. Also we need to make sure that we don't break downstream projects, like litmus which rely just on collection.

gimmyxd:
I'm gonna merge this since we need to support nightly builds for the windows modules, we can followup on refactoring on a new ticket.

PR #457, opened 2019-12-10
(MODULES-10308) Allow downgrades when using apt

beechtom:
This updates the install task to allow for downgrading the Puppet agent
when installing with `apt`. This aligns the behavior of the task with
other package handlers, such as `rpm`, that allow for downgrading the
Puppet agent version.

Part of https://github.com/puppetlabs/bolt/issues/1208

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -433,12 +433,12 @@ install_file() {
       apt-get update -y
 
       if test "$version" = 'latest'; then
-        apt-get install -y puppet-agent
+        apt-get install -y puppet-agent --allow-downgrades
       else
         if test "x$deb_codename" != "x"; then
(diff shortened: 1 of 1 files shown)


Dorin-Pleava:
Tested on a debian 10 machine and it is working as expected.

donoghuc:
@beechtom I dont think this will work downgrading across platform versions (IE 6 to 5) even with the `--allow-downgrades` option. Also I think this should have acceptance tests so we get a sense of how it works across all the platforms we test on.

gimmyxd:
Tomorrow we are going to release a new version of this module.
Would be better to revert this PR for that release? @donoghuc @beechtom

donoghuc:
I dont think the change will hurt anything and will be fine to add in a release. I just dont think the functionality fully works as described.

beechtom:
I would revert the PR for release, and I'll work on getting acceptance tests written.

PR #460, opened 2020-01-08
(PA-3018) Add warning for Win2008

GeorgeMrejea:


manifests/install/windows.pp
```diff
@@ -42,6 +42,7 @@
 
   notice ("Puppet upgrade log file at ${_logfile}")
   debug ("Installing puppet from ${_msi_location}")
+  warning("Deprecation warning: On January 14, 2020, support for Windows Server 2008 and 2008 R2 will end.")
```
gimmyxd: Where will this be displayed?
Also this should be applied only for windows 2008

PR #461, opened 2020-01-16
rework how module manages the service

rstuart-indue:
The module manages the service for some OSs only (and when not PE). This change makes this more consistent to manage the service on all platforms, regardless of PE. 

In particular, windows was a case where the module was excluded from managing the service. We want the puppet agent started again even when doing a manual run so that it is restored to service. But there is also the option added to disable this behavior.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -25,6 +25,9 @@
 # [package_version]
 #   The package version to upgrade to. Explicitly specify the version to upgrade to,
 #   or set to 'auto' to specify the version of the compiling master.
+# [manage_service]
+#   A boolean to simply indicate if we should ensure the service is running.
+#   If not, don't touch it
 # [service_names]
 #   An array of services to start, normally `puppet` and `mcollective`.
 #   None will be started if the array is empty.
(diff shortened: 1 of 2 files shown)


McdonaldSeanp:
The windows service is managed differently (and not with puppet) during the upgrade, because the services need to be shut off during upgrade and the upgrade cannot be performed by puppet itself. `install_puppet.ps1` includes code that should restore service operation after the upgrade: https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/master/files/install_puppet.ps1#L358

rstuart-indue:
It sounds like you are describing the restart of the application so that it doesn't interfere with installation, but this is different to managing the service. It sounds like there might be two different purposes being talked about here:
- starting the agent after installation or upgrade
- ensuring the service is running all the time

It seems to me that the first is what you are describing as the intent of the code in this module, but the second is what is being done for a large number of systems. If managing the service is not wanted, then I would suggest removing it completely and triggering starts of the agent with whatever tools are necessary but avoiding managing the service. Otherwise, there is going to be conflicts if people choose to manage the service and also use this module on non-PE systems and non-windows systems (eg RHEL).

wilrodriguez:
Another thing to mention here is that the module does not attempt to manage the service on PE at all, even though PE does not manage the puppet service. So, that probably needs to be changed too.

PR #462, opened 2020-01-17
(MODULES-10452) defensive pattern in prepare/puppet_config.pp creates…

kaessert:
… duplicate-resourcee problems


joshcooper:
@kaessert could you update your commit message to describe what problem this is solving and how it solves it?

PR #463, opened 2020-01-17
(MODULES-10454) Install 'latest' only when no agent is present

beechtom:
This modifies the behavior of the `puppet_agent::install` task so that
it only assumes that the latest version of the agent should be installed
when no agent is present on the system. Previously, if the `version`
parameter was undefined, the task would default to the latest version
and install it on the system, even if an agent was already present. This
would clobber any existing version of the agent, even unintentionally.

Part of puppetlabs/bolt#1208

### DIFF
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -24,10 +24,24 @@ catch [System.Management.Automation.CommandNotFoundException] {
   }
 }
 
+function Test-PuppetInstalled {
+  $result = & "$PSScriptRoot\version_powershell.ps1" | ConvertFrom-Json
+  return $result.version
+}
+
 if ($version) {
-    $msi_name = "puppet-agent-${version}-${arch}.msi"
+    if ($version -eq "latest") {
+      $msi_name = "puppet-agent-${arch}-latest.msi"
+    } else {
+      $msi_name = "puppet-agent-${version}-${arch}.msi"
+    }
 }
 else {
+    if (Test-PuppetInstalled) {
+      Write-Output "Puppet Agent detected and no version specified. Nothing to do."
+      Exit
+    }
+
     $msi_name = "puppet-agent-${arch}-latest.msi"
 }
 
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -102,6 +102,26 @@ else
   installed_version=uninstalled
 fi
 
+# Only install the agent in cases where no agent is present, or the version of the agent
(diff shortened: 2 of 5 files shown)


donoghuc:
Unfortunately acceptance testing is annoying for this particular module. In the past I have made a pr to CJC and built an experimental pipeline pointing at fork/branch to be tested. https://github.com/puppetlabs/ci-job-configs/pull/6306/files I would not be suprised at all if there are issues on some of these platforms. I think that we should also make sure the acceptance tests really cover the intended behavior.

tasks/install_shell.sh
```diff
@@ -102,6 +102,26 @@ else
   installed_version=uninstalled
 fi
 
+# Only install the agent in cases where no agent is present, or the version of the agent
+# has been explicitly defined and does not match the version of an installed agent.
+if [ -z "$version" ]; then
+  if [ "$installed_version" == "uninstalled" ]; then
+    info "Version parameter not defined and no agent detected. Assuming latest."
+    version=latest
```
donoghuc: Should this be quoted?

donoghuc: version="latest"

donoghuc: I guess bash doesnt care about bare words. we do it other places.

beechtom:
Updated the powershell implementation and re-ran the acceptance tests. All green.

tasks/install_powershell.ps1
```diff
@@ -24,10 +24,23 @@ catch [System.Management.Automation.CommandNotFoundException] {
   }
 }
 
+function Test-PuppetInstalled {
+  Get-WmiObject -Class Win32_Product | sort-object Name | select Name | where { $_.Name -match "Puppet Agent"}
```
donoghuc: Can we not just call the powershell and bash code for the version task instead of having separate logic for detecting a puppet agent?

beechtom: Updated to use `version_powershell.ps1` to detect if an agent is installed. The `install_bash.sh` task could also be updated to use `version_bash.sh` to detect if an agent is installed, but I think that'd be better suited for another ticket since it would require a change to both tasks.

donoghuc: Yeah dealing with JSON in bash is a pain. Copying over the one-liner from the shell script is OK for now. I think really what we would want is to make this a plan. But again, another day another ticket...

It looks like the travis failure is unrelated to this work, I will let someone from nightswatch decide if they want to merge this before that is fixed.

PR #466, opened 2020-01-24
(maint) Prepare 3.0.0 release

beechtom:
Part of puppetlabs/bolt#1208


donoghuc:
Given that tasks are such a small part of this module, i'm not sure the changes to the install task https://github.com/puppetlabs/puppetlabs-puppet_agent/commit/a91db25b394ae6fda29e59c447f3673b8ccd0adb really need a major version release.

lucywyman:
We discussed the versioning at standup, and decided that if Night's Watch was ok with it we'd like to do an X release

PR #467, opened 2020-01-28
(MODULES-10514) Remove use of version_powershell task in install task

beechtom:
This removes the use of the `version_powershell` task in the
`install_powershell` task. Previously, this task was used to determine
whether the `puppet-agent` package was already installed. However, some
features used to run the script and parse its output are not available
in versions of powershell older than 3.0.

The `install_powershell` task instead uses behavior similar to the
`version_powershell` task by checking the system's registry entries to
check if the `puppet-agent` package is installed.

### DIFF
--- a/tasks/install.json
+++ b/tasks/install.json
@@ -36,6 +36,6 @@
   },
   "implementations": [
     {"name": "install_shell.sh", "requirements": ["shell"], "files": ["facts/tasks/bash.sh"], "input_method": "environment"},
-    {"name": "install_powershell.ps1", "requirements": ["powershell"], "files": ["puppet_agent/tasks/version_powershell.ps1"]}
+    {"name": "install_powershell.ps1", "requirements": ["powershell"]}
   ]
 }
--- a/tasks/install_powershell.json
+++ b/tasks/install_powershell.json
@@ -34,6 +34,5 @@
       "description": "Whether to stop the puppet agent service after install",
       "type": "Optional[Boolean]"
     }
-  },
-  "files": ["puppet_agent/tasks/version_powershell.ps1"]
+  }
 }
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -25,8 +25,13 @@ catch [System.Management.Automation.CommandNotFoundException] {
 }
 
 function Test-PuppetInstalled {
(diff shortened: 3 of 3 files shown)


donoghuc:
```powershell
Windows PowerShell
Copyright (C) 2009 Microsoft Corporation. All rights reserved.

PS C:\Users\Administrator> $rootPath = 'HKLM:\SOFTWARE\Puppet Labs\Puppet'
PS C:\Users\Administrator>
PS C:\Users\Administrator> $reg = Get-ItemProperty -Path $rootPath -ErrorAction SilentlyContinue
PS C:\Users\Administrator> if ($null -ne $reg) {
>>   if ($null -ne $reg.RememberedInstallDir64) {
>>     $loc = $reg.RememberedInstallDir64+'VERSION'
>>   } elseif ($null -ne $reg.RememberedInstallDir) {
>>     $loc = $reg.RememberedInstallDir+'VERSION'
>>   }
>> }
>>
PS C:\Users\Administrator> if ( ($null -ne $loc) -and (Test-Path -Path $loc) ) {
>>   Write-Output "{`"version`":`"$(Get-Content -Path $loc -ErrorAction Stop)`",`"source`":`"$($loc.replace('\', '/'))`"
}"
>> } else {
>>   Write-Output '{"version":null,"source":null}'
>> }
>>
{"version":"6.12.0","source":"C:/Program Files/Puppet Labs/Puppet/VERSION"}
PS C:\Users\Administrator> $psversiontable

Name                           Value
----                           -----
CLRVersion                     2.0.50727.8806
BuildVersion                   6.0.6002.18111
PSVersion                      2.0
WSManStackVersion              2.0
PSCompatibleVersions           {1.0, 2.0}
SerializationVersion           1.1.0.1
PSRemotingProtocolVersion      2.1


PS C:\Users\Administrator>
```

beechtom:
The version task works in powershell < 3. The issue is how the version task is used by the install task. Currently, the install task uses the `$PSScriptRoot` variable to get the directory that both scripts were saved to and then uses the `ConvertFrom-Json` cmdlet to parse the output from the version task. Neither of these are available in powershell < 3.

There are ways to get the parent directory and parse json in powershell < 3, but it seems overly complex for what the install task needs to do, which is just determine if the agent is installed at all, not determine its version or the source.

PR #471, opened 2020-02-26
(maint) Allow install task to downlad macos 10.15

donoghuc:
Add condintion to allow downloading puppet agent packages from 10.15 repo.

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -176,6 +176,7 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
       "10.12") platform_version="10.12";;
       "10.13") platform_version="10.13";;
       "10.14") platform_version="10.14";;
+      "10.15") platform_version="10.15";;
       *) echo "No builds for platform: $major_version"
          exit 1
          ;;


donoghuc:
I just saw https://github.com/puppetlabs/ci-job-configs/commit/de5120505eef07097be8ba3c9beea07a4b789664#diff-420f83aded712be3e269ac56d953dc18R1325-R1326 was merged. Those acceptance tests have not yet triggered but when they do they will certainly fail without at least this commit. I suspect there will also be some other work making the tests not fail due to not having a puppet 5 version to upgrade from https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/e85f549f0877387bce97372f137b53fbef60ea12/task_spec/spec/acceptance/init_spec.rb#L31-L40

nicklewis:
There has to be a better way!

I wonder if it would be possible to make the module look where it _expects_ the file to be and fail if it's not there, rather than continue to update it with new whitelists...

tasks/install_shell.sh
```diff
@@ -176,6 +176,7 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
       "10.12") platform_version="10.12";;
       "10.13") platform_version="10.13";;
       "10.14") platform_version="10.14";;
+      "10.15") platform_version="10.15";;
```
GabrielNagy, line 179: Maybe check if `major_version` matches `10\.1[1-9]` (to be on the safe side at least), then pass it on as is? We would get an uglier error when curl-ing the package but the file wouldn't need to be changed for each macOS release.

PR #472, opened 2020-03-03
(MODULES-10589) Exit early when puppet.list config file has been modified

beechtom:
This updates the `install_shell` task to check whether the configuration
file `/etc/apt/repos.list.d/puppet.list` has been modified prior to
installing the `puppet-release` repo.

Previously, if a user had modified the file (such as by manually
maintaining the repo), `dpkg` would prompt the user whether they wanted
to keep the old config file or replace it with the maintainer's config
file. This caused the script to stall as it waited for user input. While
`dpkg` can answer these prompts non-interactively using the
`--force-conf*` options, this behavior would be undesirable as it could
result in the script making a critical decision for the user. Instead,
the script will provide a helpful warning that the config file has been
modified and that agent installation will be skipped.

Closes https://github.com/puppetlabs/bolt/issues/1593

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -35,6 +35,32 @@ exists() {
   fi
 }
 
+# Check whether the apt config file has been modified, warning and exiting early if it has
+assert_unmodified_apt_config() {
+  puppet_list=/etc/apt/sources.list.d/puppet.list
+  puppet6_list=/etc/apt/sources.list.d/puppet6.list
+
+  if [[ -f $puppet_list ]]; then
+    list_file=puppet_list
+  elif [[ -f $puppet6_list ]]; then
+    list_file=puppet6_list
+  fi
+
+  # If puppet.list exists, get its md5sum on disk and its md5sum from the puppet-release package
+  if [[ -n $list_file ]]; then
+    # For md5sum, the checksum is the first word
+    file_md5=($(md5sum "$list_file"))
+    # For dpkg-query with this output format, the sum is the second word
+    package_md5=($(dpkg-query -W -f='${Conffiles}\n' 'puppet-release' | grep -F "$list_file"))
+
+    # If the $package_md5 array is set, and the md5sum on disk doesn't match the md5sum from dpkg-query, it has been modified
+    if [[ $package_md5 && ${file_md5[0]} != ${package_md5[1]} ]]; then
+      warn "Configuration file $list_file has been modified from the default. Skipping agent installation."
+      exit 1
+    fi
+  fi
+}
+
 # Check whether perl and LWP::Simple module are installed
 exists_perl() {
   if perl -e 'use LWP::Simple;' >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


tasks/install_shell.sh
```diff
@@ -35,6 +35,19 @@ exists() {
...
 }
 
+# Check whether the apt config file has been modified, warning and exiting early if it has
+assert_unmodified_apt_config() {
+  dpkg-query -W -f='${Conffiles}\n' '*' | awk 'OFS="  "{print $2,$1}' | md5sum -c 2>/dev/null | awk -F': ' '$2 !~ /OK/{print $1}' | grep -q puppet.list
+
+  local result=$?
+  local file="/etc/apt/sources.list.d/puppet.list"
+
+  if [ -f "$file" ] && [ "$result" -eq 0 ]; then
```
beechtom: The file check is here since the previous command will report that the file was modified if it doesn't exist. If the file doesn't exist, then there won't be a conflict with a manually maintained file, so we should go ahead and install the repo and agent.

tasks/install_shell.sh
```diff
@@ -441,7 +454,9 @@ install_file() {
         fi
       fi
 
-      dpkg -i "$2"
+      assert_unmodified_apt_config
+
+      dpkg -i --force-confmiss "$2"
```
beechtom: `--force-confmiss` will install any missing config files.

m0dular:
I updated my account but haven't receive the cla email ¯\_(ツ)_/¯

lucywyman:
@beechtom I think you need to update the commit message to adhere to contributing guidelines:
```
$ bundle exec rake $CHECK
Checking commits e85f549f0877387bce97372f137b53fbef60ea12..6f0d3480096eea365763e071c71be0c576cf20ff
rake aborted!

	This commit summary didn't match CONTRIBUTING.md guidelines:
		Only check for puppet.list in the puppet-release package
	The commit summary (i.e. the first line of the commit message) should start with one of:
		(docs)
		(docs)(DOCUMENT-<digits>)
		(packaging)
```

We might need to make a modules ticket for this :/

tasks/install_shell.sh
```diff
@@ -35,6 +35,25 @@ exists() {
   fi
 }
 
+# Check whether the apt config file has been modified, warning and exiting early if it has
+assert_unmodified_apt_config() {
+  list_file=/etc/apt/sources.list.d/puppet.list
```
lucywyman: With the latest puppet docs this would actually be puppet6.list - we should definitely look for both. I don't think dpkg will let a user have both at the same time, so just checking for whichever exists is sufficient until someone runs into issues with both existing.

donoghuc:
Dont want to be bug but https://jenkins-master-prod-1.delivery.puppetlabs.net/view/Bolt-bundled-content/job/forge-module_puppetlabs-puppet-agent-module_intn-run_task_acceptance-master/ are still failing. Started with https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/471 but that is not a full solution. Is getting CI green a priority before merging new task content?

PR #477, opened 2020-03-19
(packaging) Update inifile dependency to allow all 4.x versions

MikaelSmith:
A prior change ensured that puppet_agent didn't use inifile version 3.2.0. That version comparison seems to have been preserved when moving to any 3.x version in a non-sensical way (allowing 4.0.0 but not anything later). Switch to allowing all versions prior to an unreleased breaking 5.0 version.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -120,7 +120,7 @@
   ],
   "dependencies": [
     {"name":"puppetlabs-stdlib","version_requirement":">= 5.1.0 < 7.0.0"},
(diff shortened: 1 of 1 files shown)


GabrielNagy:
Hi @MikaelSmith,

Can you amend your commit message title to match one of:
```
(docs)
(docs)(DOCUMENT-<digits>)
(packaging)
```
I'd go for `packaging` or creating a MODULES ticket.

Thanks!

PR #480, opened 2020-04-01
(maint) Fix task acceptance for osx 10.14 & 10.15

GabrielNagy:
Change versions to upgrade from for OS X 10.14 and 10.15.

OS X 10.15 should be enabled in CI after puppet 5.5.20 is released, so it would have a version to upgrade from.


donoghuc:
Nice. I think for now we can merge https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/471/files Did ya'll have any ideas on how to improve that logic?

GabrielNagy:
@donoghuc I'm gonna merge #471 to have it in the next release, but I left a comment with a possible improvement.

PR #483, opened 2020-04-09
Add a plan for upgrading puppet-agent

npwalker:


donoghuc:
Does this upgrade restart the pxp-agent service? Seems like that would not work over PCP. Maybe an argument for apply over ssh/winrm?

PR #487, opened 2020-05-07
(MODULES-10661) Add OS X 10.15 support

GabrielNagy:
Tested with PE by upgrading from agent version 6.13.0 to 6.15.0.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -108,7 +108,9 @@
       "operatingsystem": "OSX",
       "operatingsystemrelease": [
         "10.12",
-        "10.13"
+        "10.13",
+        "10.14",
+        "10.15"
        ]
     }
   ],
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -55,11 +55,12 @@ task(:commits) do
   %x{git log --no-merges --pretty=%s #{commit_range}}.each_line do |commit_summary|
     # This regex tests for the currently supported commit summary tokens.
     # The exception tries to explain it in more full.
-    if /^\((maint|packaging|doc|docs|modules-\d+)\)|revert/i.match(commit_summary).nil?
+    if /^\((maint|packaging|doc|docs|(modules|pe)-\d+)\)|revert/i.match(commit_summary).nil?
```
gimmyxd, line 58: I think we decided to allow only `MODULES` for this repos to be easier to track them, 
Upgrade support is also for FOSS but PE tickets are not public.

PR #488, opened 2020-05-07
(MODULES-10655) Fix up/downgrade of agent to specified version

hajee:
This fix will make sure then on RedHat like kernels, the `package_version` is used to determine if the agent should be downgraded or upgraded  when when using `absolute_source` parameter.


GabrielNagy:
Hi @hajee,

Thanks for your contribution! I validated your change on a CentOS 8 machine and it fixes the original issue. :tada:

I also triggered a run of the acceptance tests in our internal infrastructure. I'll merge the PR when it passes.

PR #493, opened 2020-05-14
(maint) bump agent version in tests

gimmyxd:


gimmyxd:
fixed smoke tests:
image

PR #496, opened 2020-05-21
(MODULES-10666) Stop agent run after an upgrade

GabrielNagy:
Previously, when evaluating additional resources after a puppet-agent
upgrade, we could get to a point where we had a mixture of old/new
Ruby code loaded in memory. This may cause unexpected behavior if the
new code tries to execute something not available in the already
loaded old code.

This commit introduces a new type/provider that is evaluated whenever
a puppet-agent upgrade happens on a platform managed with the package
resource (AIX, Debian, EL, Suse).

The newly added resource attempts to figure out if a Puppet upgrade
was performed, then stops the run altogether. If Puppet is being run
as a service, the HUP signal will have the daemon trigger a subsequent
Puppet run that will evaluate the remaining resources.

When running `puppet agent -t` manually, it is up to the user to rerun
the command after the puppet-agent upgrade.

### DIFF
--- a/lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
+++ b/lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
@@ -1,0 +1,35 @@
+Puppet::Type.type(:puppet_agent_end_run).provide :puppet_agent_end_run do
+  def end_run
+    false
+  end
+
+  def stop
+    if needs_upgrade?
+      run_mode = Puppet.run_mode.name
+
+      # Handle CLI runs of `puppet agent` and `apply`
+      if Puppet[:onetime] || run_mode != :agent
+        Puppet.notice("Stopping run after puppet-agent upgrade. Run puppet agent -t or apply your manifest again to finish the transaction.")
+      end
+
+      Puppet::Application.stop!
+
+      # Sending the HUP signal to the daemon causes it to restart and finish applying the catalog
+      if Puppet[:daemonize] && run_mode == :agent
+        at_exit { Process.kill(:HUP, Process.pid) }
+      end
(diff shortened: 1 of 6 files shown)


lib/puppet/type/puppet_agent_abort_run.rb
```diff
@@ -0,0 +1,30 @@
+require 'puppet/property/boolean'
+
+Puppet::Type.newtype(:puppet_agent_abort_run) do
```
joshcooper, line 3: `Abort` makes me think there was an error during the run. Maybe refer to the type and provider as `:puppet_agent_restart`?

GabrielNagy: `restart` would imply that the run is restarted, that only happens when running as a daemon however.

I'll try to find a better name.

npwalker: `puppet_agent_end_run`

GabrielNagy:
I added a note in the readme to reflect this change in functionality. This should be ready to merge now.

lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
```diff
@@ -0,0 +1,33 @@
+Puppet::Type.type(:puppet_agent_end_run).provide :puppet_agent_end_run do
+  def end_run
+    false
+  end
+
+  def needs_upgrade
```
gimmyxd, line 6: you could rename this to `needs_upgrade?` as it is the convention for methods that return a boolean

GabrielNagy:
Hmm, ad-hoc is failing on Debian-based platforms. I'll check that.

PR #499, opened 2020-06-19
(MODULES-10713) Fix agent upgrade on Solaris 11

GabrielNagy:
Due to Solaris p5p limitations, we munge dev build versions by stripping all letters (i.e.: 5.5.20.102.g9e5216a6 turns to 5.5.20.102.952166).
This causes `Gem::Version` to wrongly interpret the desired version on Solaris 11, making upgrades not idempotent.

Use a regex instead of `Gem::Version` to make `package_version` more like the `aio_agent_version` fact. Move this logic from the provider to the Puppet manifests in order to be able to properly spec test this.


GabrielNagy:
image

PR #503, opened 2020-08-10
(MODULES-10768) Add task and plan for running Puppet agent

beechtom:
This adds a new `puppet_agent::run` task that runs the Puppet agent on a
list of targets, and returns the run report contents as well as the
exit code for the agent run.

This also adds a `puppet_agent::run` plan which checks if the agent is
installed on each target and runs the `puppet_agent::run` task on each
target that has the agent installed. The plan returns a `ResultSet` that
includes fail results for targets that do not have the agent installed,
as well as the results from the `puppet_agent::run` task.

### DIFF
--- a/plans/run.pp
+++ b/plans/run.pp
@@ -1,0 +1,70 @@
+# Starts a Puppet agent run on the specified targets.
+# Note: This plan may cause issues when run in Puppet Enterprise.
+# @param targets The targets to start a Puppet agent run on.
+plan puppet_agent::run (
+  TargetSpec $targets
+) {
+  # Check which targets have the agent installed by checking
+  # the version of the agent. No point in trying to run the
+  # agent if it's not installed.
+  $version_results = run_task(
+    'puppet_agent::version',
+    $targets,
+    'Check for Puppet agent',
+    '_catch_errors' => true
+  )
+
+  # Create results with more descriptive error messages for any targets
+  # where the version task failed.
+  $version_error_results = $version_results.error_set.map |$result| {
+    $err = {
+      '_error' => {
+        'msg'     => "The task puppet_agent::version failed: ${result.error.message}. Unable to determine if the Puppet agent is installed.",
+        'kind'    => 'puppet_agent/agent-version-error',
+        'details' => {}
+      }
+    }
+
+    Result.new($result.target, $err)
+  }
+
+  # Filter targets by those that have an agent installed and
+  # those that don't. The puppet_agent::version task will return
+  # version:null for any targets that don't have an agent.
+  $agentless_results = $version_results.ok_set.filter_set |$result| {
+    $result['version'] == undef
+  }
+
+  $agent_results = $version_results.ok_set.filter_set |$result| {
+    $result['version'] != undef
+  }
+
+  # Create fail results for agentless targets.
+  $agentless_error_results = $agentless_results.map |$result| {
+    $err = {
+      '_error' => {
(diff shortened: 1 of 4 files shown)


plans/run.pp
```diff
@@ -0,0 +1,56 @@
+# Starts a Puppet agent run on the specified targets.
+# Note: This plan may cause issues when run in Puppet Enterprise.
+# @param targets The targets to start a Puppet agent run on.
+plan puppet_agent::run (
+  TargetSpec $targets
+) {
+  # Check which targets have the agent installed by checking
+  # the agent of the version. No point in trying to run the
```
lucywyman, line 8:
```suggestion
  # the version of the agent. No point in trying to run the
```

plans/run.pp
```diff
@@ -0,0 +1,56 @@
...
+
+  $agent_results = $version_results.filter_set |$result| {
+    $result['version'] != undef
+  }
+
+  # Create fail results for agentless targets.
+  $error_results = $agentless_results.map |$result| {
+    $err = {
+      '_error' => {
+        'msg'     => 'Puppet agent is not installed on the target.',
```
lucywyman, line 32:
```suggestion
        'msg'     => 'Puppet agent is not installed on the target. Run the puppet_agent::install task on these targets to install the puppet agent.',
```

plans/run.pp
```diff
@@ -0,0 +1,56 @@
...
+    'puppet_agent::run',
+    $agent_results.targets,
+    'Run Puppet agent',
+    '_catch_errors' => true
+  )
+
+  # Merge all of the results into a single ResultSet so each
+  # target has a result.
+  return ResultSet.new(
+    $version_results.error_set.results +
```
lucywyman, line 52: This results in odd output when the version task fails:
```
lucy ~/githubs/bolt (GH-2051) $ be bolt plan run puppet_agent::run -t floaty
Starting: plan puppet_agent::run
Starting: Check for Puppet agent on simple-woodwind.delivery.puppetlabs.net
Finished: Check for Puppet agent with 1 failure in 0.99 sec
Finished: plan puppet_agent::run in 1.0 sec
Failed on simple-woodwind.delivery.puppetlabs.net:
  The task failed with exit code 1 and no output
Failed on simple-woodwind.delivery.puppetlabs.net:
  Puppet agent is not installed on the target.
Failed on 2 targets: simple-woodwind.delivery.puppetlabs.net,simple-woodwind.delivery.puppetlabs.net
Ran on 2 targets
```
I think we want to make `$agentless_result` just from results that succeeded the version task so that failures don't show up twice, and if you want to be really fancy modify the error message results to be more helpful.

plans/run.pp
beechtom: 😆 How did I miss that?

plans/run.pp
```diff
@@ -0,0 +1,70 @@
...
+  $version_results = run_task(
+    'puppet_agent::version',
+    $targets,
+    'Check for Puppet agent',
+    '_catch_errors' => true
+  )
+
+  # Create results with more descriptive error messages for any targets
+  # where the version task failed.
+  $fail_results = $version_results.error_set.map |$result| {
```
lucywyman, line 19: I think `version_error_results` might be less confusing here, mostly since it's not clear what the difference is between a fail result and error result (plus we don't fail the plan here)

beechtom: Updated! Also changed `error_results` to `agentless_error_results` for similar reasons.

PR #510, opened 2020-09-14
(MODULES-10780) PA 6 to PA 7 upgrade tests

Dorin-Pleava:
This PR tests the upgradability to puppet 7 using the
puppetlabs-puppet_agent module.

Currently it uses the puppet 7 nightly builds, as a puppet7 collection
does not exist.

### DIFF
--- a/acceptance/tests/test_upgrade_puppet6_to_puppet7.rb
+++ b/acceptance/tests/test_upgrade_puppet6_to_puppet7.rb
@@ -1,0 +1,44 @@
+require 'beaker-puppet'
+require_relative '../helpers'
+
+# Tests FOSS upgrades from the latest puppet 6 (the puppet6 collection) to puppet 7.0.0.
+test_name 'puppet_agent class: Upgrade agents from puppet6 to puppet7' do
+  # require_master_collection 'puppet7' #once puppet7 collection is available this can be uncommented
+  exclude_pe_upgrade_platforms
+  latest_version = `curl http://builds.delivery.puppetlabs.net/passing-agent-SHAs/puppet-agent-main-version`.gsub(/\d*\.\d*.\d*\./,"7.0.0.")
+  
+  puppet_testing_environment = new_puppet_testing_environment
+  step "Create new site.pp with upgrade manifest" do
+    manifest = <<-PP
+    node default {
+      class { puppet_agent:
+        package_version => '#{latest_version}',
+        apt_source => 'http://nightlies.puppet.com/apt',
+        yum_source => 'http://nightlies.puppet.com/yum',
+        collection      => 'puppet7-nightly'
+      }
+    }
+    PP
+    site_pp_path = File.join(environment_location(puppet_testing_environment), 'manifests', 'site.pp')
+    create_remote_file(master, site_pp_path, manifest)
+    on(master, %(chown #{puppet_user(master)} "#{site_pp_path}"))
+    on(master, %(chmod 755 "#{site_pp_path}"))
+  end
+
+  agents_only.each do |agent|
+    set_up_initial_agent_on(agent, 'puppet6') do
+      step '(Agent) Change agent environment to testing environment' do
+        on(agent, puppet("config --section agent set environment #{puppet_testing_environment}"))
+        on(agent, puppet("config --section user set environment production"))
+      end
+    end
+  end
+
+  step "Upgrade the agents from Puppet 6 to Puppet 7..." do
+    agents_only.each do |agent|
+      start_puppet_service_and_wait_for_puppet_run(agent)
+      wait_for_installation_pid(agent)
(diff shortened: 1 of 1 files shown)


ciprianbadescu:
Could we keep one test and just set different *_source based on platform?

acceptance/tests/test_upgrade_puppet6_to_puppet7_yum.rb
```diff
@@ -0,0 +1,43 @@
...
+  confine :to, :platform => /el-|fedora-/
+  # require_master_collection 'puppet7' #once puppet7 collection is available this can be uncommented
+  
+  exclude_pe_upgrade_platforms
+  puppet_testing_environment = new_puppet_testing_environment
+  step "Create new site.pp with upgrade manifest" do
+    manifest = <<-PP
+    node default {
+      class { puppet_agent:
+        package_version => '7.0.0.224.gaf2034af',
```
gimmyxd, line 15: Maybe we should upgrade to `latest` as nightly versions change quite often(and a cleanup is done for old versions) and we need to ensure that we can upgrade.

Dorin-Pleava: I think puppetlabs-puppet_agent does not support `latest` in its manifest. It can support `auto` but it will only update to the master server version.
Once puppet7 is released we won't have to specify a source for the package, but until then I can't think of a better workaround.

gimmyxd: The issue with fix version like `7.0.0.224.gaf2034af` is that old versions get removed from nightlies, so we might end-up with this test failing because that version is not found.

Also the test is confined only on debian based platforms, which is a bit restrictive.

If we want just to have a sanity check that the upgrade just work on some platforms, we can use osx/windows which are not relying on a package provider. There we have `puppet-agent-latest` packages.

You can also take a look on how the bolt task is testing the upgrade from puppet 6 to puppet 7 nightly,

Dorin-Pleava:
@ciprianbadescu I thought it could work, but that manifest is generated on the master node, and it only knows about the OS of the master. I am trying to create a manifest that has this logic, but don't think the syntax allows for something like this.

`      $source = case $facts['os']['name'] {
        /^(Debian|Ubuntu)$/: { "apt_source => 'http://nightlies.puppet.com/apt'" }
        /^(Redhat|CentOS|Fedora)$/: { "yum_source => 'http://nightlies.puppet.com/yum'" }
      }

      class { puppet_agent:
        package_version => '7.0.0.224.gaf2034af',
        $source,
        collection      => 'puppet7-nightly'
      }`

acceptance/tests/test_upgrade_puppet6_to_puppet7_yum.rb
GabrielNagy: :+1:, maybe we can programatically get the latest build available in a nightly collection. i.e. by `curl`-ing http://builds.delivery.puppetlabs.net/passing-agent-SHAs/puppet-agent-main-version and replacing `6.18.0` with `7.0.0`

acceptance/tests/test_upgrade_puppet6_to_puppet7.rb
```diff
@@ -0,0 +1,52 @@
...
+  exclude_pe_upgrade_platforms
+  latest_version = `curl http://builds.delivery.puppetlabs.net/passing-agent-SHAs/puppet-agent-main-version`.gsub(/\d*\.\d*.\d*\./,"7.0.0.")
+  
+  puppet_testing_environment = new_puppet_testing_environment
+  step "Create new site.pp with upgrade manifest" do
+    manifest = <<-PP
+    node default {
+      class { puppet_agent:
+        package_version => '#{latest_version}',
+        #{source},
```
GabrielNagy, line 25: Actually, I think you can have both `apt_source` and `yum_source` in the same class and each OS will use what it needs.

Dorin-Pleava: Yep, just tested and it worked :))

GabrielNagy (approved):

acceptance/tests/test_upgrade_puppet6_to_puppet7.rb
```diff
@@ -0,0 +1,44 @@
...
+        on(agent, puppet("config --section user set environment production"))
+      end
+    end
+  end
+
+  step "Upgrade the agents from Puppet 6 to Puppet 7..." do
+    agents_only.each do |agent|
+      start_puppet_service_and_wait_for_puppet_run(agent)
+      wait_for_installation_pid(agent)
+      assert_agent_version_on(agent, latest_version.scan(/\d*\.\d*\.\d*\.\d*/).first)
```
gimmyxd, line 41: The first digit should be 7 i guess, to make sure that the upgrade succeed.

PR #511, opened 2020-09-18
(MODULES-10799) Ensure upgradability from puppet 6 to 7 when remote filebuckets are enabled

GabrielNagy:
Puppet 7 changed the default digest algorithm to SHA256 in
https://github.com/puppetlabs/puppet/pull/8325. This could cause
failures when upgrading if remote filebuckets are also enabled.

Before performing the upgrade, check if remote filebuckets are enabled
and compare the digest algorithms of the agent/server. If they differ,
abort the upgrade.

### DIFF
--- a/lib/puppet/functions/any_resources_of_type.rb
+++ b/lib/puppet/functions/any_resources_of_type.rb
@@ -1,0 +1,25 @@
+Puppet::Functions.create_function(:any_resources_of_type, Puppet::Functions::InternalFunction) do 
+  dispatch :any_resources_of_type do
+    scope_param
+
+    required_param 'String', :resource_type
+    optional_param 'Hash[Any, Any]', :parameters
+  end
+
+  def any_resources_of_type(scope, resource_type, parameters = nil)
+    scope.catalog.resources.any? do |resource|
+      # We should always iterate over Puppet::Parser::Resource
+      # instances here, and documentation states that types can be
+      # strings or symbols.
+      # https://www.rubydoc.info/gems/puppet/Puppet/Resource#initialize-instance_method
+      if resource.type.to_s.casecmp(resource_type).zero? # String#casecmp? is Ruby 2.4+
+        if parameters
+          # If the resource matched, but any of the params didn't, go to the next one
+          next if parameters.any? { |k, v| resource.to_hash[k.to_sym] != v }
+        end
+        true
+      end
+    end
+  end
+end
+
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -168,6 +168,14 @@ class puppet_agent (
 
     $aio_upgrade_required = versioncmp($::aio_agent_version, $_expected_package_version) < 0
 
+    if $aio_upgrade_required {
+      if any_resources_of_type('filebucket', { path => false }) {
+        if $settings::digest_algorithm != $::puppet_digest_algorithm {
+          fail("Remote filebuckets are enabled, but there was a agent/server digest algorithm mismatch. Server: ${settings::digest_algorithm}, agent: ${::puppet_digest_algorithm}. Either ensure the algorithms are matching, or disable remote filebuckets during the upgrade.")
+        }
+      }
+    }
+
     if $::operatingsystem == 'Solaris' and $::operatingsystemmajrelease == '11' {
       # Strip letters from development builds. Unique to Solaris 11 packaging.
       $_version_without_letters = regsubst($master_or_package_version, '[a-zA-Z]', '', 'G')
--- a/spec/unit/functions/any_resources_of_type_spec.rb
+++ b/spec/unit/functions/any_resources_of_type_spec.rb
@@ -1,0 +1,92 @@
+require 'spec_helper'
+require 'puppet/pops'
+require 'puppet/loaders'
+
+describe "the 'any_resources_of_type' function" do
+  after(:all) { Puppet::Pops::Loaders.clear }
+
+  # This loads the function once and makes it easy to call it
+  # It does not matter that it is not bound to the env used later since the function
+  # looks up everything via the scope that is given to it.
+  # The individual tests needs to have a fresh env/catalog set up
+  #
+  let(:loaders) { Puppet::Pops::Loaders.new(Puppet::Node::Environment.create(:testing, [File.expand_path(fixtures('modules'))])) }
+  let(:func) { loaders.private_environment_loader.load(:function, 'any_resources_of_type') }
+
+  # A fresh environment is needed for each test since tests create resources
+  let(:environment) { Puppet::Node::Environment.create(:testing, [File.expand_path(fixtures('modules'))]) }
+  let(:node) { Puppet::Node.new('yaynode', :environment => environment) }
+  let(:compiler) { Puppet::Parser::Compiler.new(node) }
+  let(:scope) { Puppet::Parser::Scope.new(compiler) }
+
+  def newresource(type, title, parameters = {})
+    resource = Puppet::Resource.new(type, title, :parameters => parameters)
+    compiler.add_resource(scope, resource)
+    resource
+  end
+
+  it 'raises if called without the "resource_type" argument' do
+    expect { func.call(scope) }.to raise_error(ArgumentError, %r{expects between.*got none})
+  end
+
+  it 'allows an optional "parameters" argument' do
+    expect { func.call(scope, 'title', { :param1 => 'value1' }) }.not_to raise_error
+  end
+
+  it 'raises if the "parameters" argument is not a hash' do
+    expect { func.call(scope, 'title', 'not_a_hash') }.to raise_error(ArgumentError, %r{parameter 'parameters' expects a Hash value, got String})
+  end
+
+  context 'when resources are defined' do
+    context 'when there are multiple resources' do
+      it 'iterates until a parameter matches' do
+        newresource('filebucket', 'bucket_1')
+        newresource('filebucket', 'bucket_2')
+        newresource('filebucket', 'bucket_3')
+        expect(func.call(scope, 'filebucket', :name => 'bucket_3')).to be_truthy
+      end
+    end
+
+    it 'finds by type name' do
+      newresource('filebucket', 'my_filebucket')
+      expect(func.call(scope, 'filebucket')).to be_truthy
+    end
+
+    it 'finds by capitalized type name' do
+      newresource('filebucket', 'my_filebucket')
+      expect(func.call(scope, 'Filebucket')).to be_truthy
+    end
+
+    it 'finds by type name and a parameter'  do
+      newresource('filebucket', 'my_filebucket')
+      expect(func.call(scope, 'filebucket', :name => 'my_filebucket')).to be_truthy
+    end
+
+    it 'finds by type name and multiple parameters'  do
+      newresource('filebucket', 'my_filebucket', :path => false)
+      expect(func.call(scope, 'filebucket', :name => 'my_filebucket', :path => false)).to be_truthy
+    end
+
+    it 'finds by type name and multiple parameters in any order'  do
+      newresource('filebucket', 'my_filebucket', :path => false)
+      expect(func.call(scope, 'filebucket', :path => false, :name => 'my_filebucket')).to be_truthy
+    end
+  end
+
+  context 'when no resources are defined' do
+    it 'does not find by type name' do
+      expect(func.call(scope, 'filebucket')).to be_falsey
+    end
+
+    it 'matches the type but not the parameters' do
+      newresource 'filebucket', 'my_filebucket'
+      expect(func.call(scope, 'filebucket', :name => 'not_my_filebucket')).to be_falsey
+    end
+
+    it 'matches the type but not all parameters'  do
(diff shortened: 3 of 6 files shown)


lib/puppet/functions/any_resources_of_type.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    required_param 'String', :resource_type
+    optional_param 'Hash', :parameters
+  end
+
+  def any_resources_of_type(scope, resource_type, parameters = nil)
+    scope.catalog.resources.any? do |resource|
+      # For some reason, resources types in this catalog are capitalized strings
+      # When walking the catalog on the agent, types are lowercase symbols
+      if resource.type.casecmp?(resource_type.capitalize)
```
GabrielNagy, line 13: Figured out why this happens. When iterating over the resources on the server, we iterate over instances of `Puppet::Parser::Resource`. When doing the same thing on the agent, we get `Puppet::Type::SomeType` instances.

Aside from `resource.type` being capitalized in `Puppet::Parser::Resource` objects, parameters also differ.
`Puppet::Type` objects also have the `:name` parameter which can be accessed using `resource[:name]`, whereas `Resource` objects do not.

However, calling `Puppet::Parser::Resource#to_hash` produces a hash containing `:name` which gave me some headaches when testing this function with multiple parameters, as I was expecting `:name` to be accessible the same way as `:path`.

I wonder if there's any particular reason why we don't have a `:name` parameter on `Puppet::Resource` objects. Rediscovered this comment https://github.com/puppetlabs/puppet/pull/8174/files#r437060949 but it specifically talks about `Puppet::Type`.

GabrielNagy: For reference, this is where `:name` gets merged in with the rest of the parameters when calling `Puppet::Parser::Resource#to_hash`:
https://github.com/puppetlabs/puppet/blob/a54094d20a52f4b0a9f2317bfb67934f9d46f8c1/lib/puppet/resource.rb#L623

manifests/init.pp
```diff
@@ -116,6 +116,12 @@
   $wait_for_pxp_agent_exit = undef,
 ) inherits ::puppet_agent::params {
 
+  if any_resources_of_type('filebucket', { path => false }) {
+    if $settings::digest_algorithm != $::puppet_digest_algorithm {
```
GabrielNagy, line 120: I guess this condition will pass if agent nodes are upgraded prior to server nodes, so then we'd get to sha256 on agent and md5 on server which is still incompatible.

We also wouldn't know whether `digest_algorithm` is explicitly set in `puppet.conf` or the default value... which raises more questions.

GabrielNagy: [Official documentation](https://puppet.com/docs/puppet/latest/upgrade_minor.html) states that server nodes should always be upgraded first, so my previous concern is unfounded.

lib/puppet/functions/any_resources_of_type.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    required_param 'String', :resource_type
+    optional_param 'Hash', :parameters
+  end
+
+  def any_resources_of_type(scope, resource_type, parameters = nil)
+    scope.catalog.resources.any? do |resource|
+      # For some reason, resources types in this catalog are capitalized strings
+      # When walking the catalog on the agent, types are lowercase symbols
+      if resource.type.casecmp(resource_type.capitalize).zero? # String#casecmp? is Ruby 2.4+
```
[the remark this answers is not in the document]
GabrielNagy: Kinda circumvented the issue above by calling `#to_hash` on the resource before checking the parameters. This looks like the better way.

spec/unit/functions/any_resources_of_type_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  # The individual tests needs to have a fresh env/catalog set up
+  #
+  let(:loaders) { Puppet::Pops::Loaders.new(Puppet::Node::Environment.create(:testing, [File.expand_path(fixtures('modules'))])) }
+  let(:func) { loaders.private_environment_loader.load(:function, 'any_resources_of_type') }
+
+  before :each do
+    # A fresh environment is needed for each test since tests create resources
+    environment = Puppet::Node::Environment.create(:testing, [File.expand_path(fixtures('modules'))])
+    @node = Puppet::Node.new('yaynode', :environment => environment)
+    @known_resource_types = environment.known_resource_types
```
GabrielNagy, line 20: This is not used anywhere, so it can be removed

spec/unit/functions/any_resources_of_type_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      it 'by looking up an existing type name with no matching parameters' do
+        newresource 'filebucket', 'my_filebucket'
+        expect(func.call(@scope, 'filebucket', :name => 'not_my_filebucket')).to be_falsey
+      end
+
+      it 'by looking up an existing type name with some matching parameters'  do
+        newresource 'filebucket', 'my_filebucket', :path => false
+        expect(func.call(@scope, 'filebucket', :name => 'not_my_filebucket', :path => '/path/to/bucket')).to be_falsey
```
GabrielNagy, line 83: should modify this test to match either `:name` or `:path`

manifests/init.pp
GabrielNagy: Reference for the `$settings::{setting}` variables: https://puppet.com/docs/puppet/5.5/lang_facts_and_builtin_vars.html#puppet-master-variables

lib/puppet/functions/any_resources_of_type.rb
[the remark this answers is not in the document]
gimmyxd: Do you need `resource_type.capitalize` if using `String#casecmp`(which is case-insensitive) ?

lib/puppet/functions/any_resources_of_type.rb
```diff
@@ -0,0 +1,23 @@
+Puppet::Functions.create_function(:any_resources_of_type, Puppet::Functions::InternalFunction) do 
+  dispatch :any_resources_of_type do
+    scope_param
+
+    required_param 'String', :resource_type
+    optional_param 'Hash[Any, Any]', :parameters
+  end
+
+  def any_resources_of_type(scope, resource_type, parameters = nil)
```
gimmyxd, line 9: Seem that this function is quite generic and it could find usages outside of this module. Would be a better place to move this on?(like stdlib pr even Puppet) Maybe not right know but we could think about future usage.

spec/unit/functions/any_resources_of_type_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  # This loads the function once and makes it easy to call it
+  # It does not matter that it is not bound to the env used later since the function
+  # looks up everything via the scope that is given to it.
+  # The individual tests needs to have a fresh env/catalog set up
+  #
+  let(:loaders) { Puppet::Pops::Loaders.new(Puppet::Node::Environment.create(:testing, [File.expand_path(fixtures('modules'))])) }
+  let(:func) { loaders.private_environment_loader.load(:function, 'any_resources_of_type') }
+
+  before :each do
+    # A fresh environment is needed for each test since tests create resources
```
gimmyxd, line 17: i think this can be replaced using `let` https://www.betterspecs.org/#let

```ruby
  let(:environment) { Puppet::Node::Environment.create(:testing, [File.expand_path(fixtures('modules'))]) }
  let(:node) { Puppet::Node.new('yaynode', :environment => environment) }
  let(:compiler) { Puppet::Parser::Compiler.new(node) }
  let(:scope) { Puppet::Parser::Scope.new(compiler) }
```

gimmyxd: And update the test of the usage in examples

spec/unit/functions/any_resources_of_type_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    expect { func.call(@scope, 'title', { :param1 => 'value1' }) }.not_to raise_error
+  end
+
+  it 'raises if the "parameters" argument is not a hash' do
+    expect { func.call(@scope, 'title', 'not_a_hash') }.to raise_error(ArgumentError, %r{parameter 'parameters' expects a Hash value, got String})
+  end
+
+  context 'can determine if a resource' do
+    context 'is defined' do
+      it 'by looking up the type name' do
```
gimmyxd, line 44: [Optional]

I would restructure this to something like:
```ruby
  context 'when resource is defined' do
    it 'finds by type name' do
	...

  context 'when resource is not defined' do
	it 'does not find by type name'
```

lib/puppet/functions/any_resources_of_type.rb
GabrielNagy: @gimmyxd good catch about `capitalize`, I used it prior to adding `String#casecmp` and forgot to remove it

I also thought about having the function in a more general place. I think it would make sense to have it in Puppet, there's also the `defined` function there which serves a similar purpose.

spec/unit/functions/any_resources_of_type_spec.rb
GabrielNagy: Makes sense. I'll also add with multiple resources of the same type.

spec/unit/functions/any_resources_of_type_spec.rb
GabrielNagy: Tried this, but got:

> `let` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block)

gimmyxd: `let` should be defined outside `before` blocks, that should work.

GabrielNagy: hmm okay, didn't know `let` blocks refreshed before each test, but seems to be working :+1:

GabrielNagy:
Another valid point is that the code from `init.pp` always executes when the `puppet_agent` class is defined, regardless of whether we need to update or not. So we can end up failing the run if remote filebuckets are on and there's a digest mismatch, even if there isn't any upgrade to do.

I'm not sure how people use the puppet_agent module, i.e. if they only apply the class when they want to upgrade, or always.
In either case, having a digest algorithm mismatch between the agent and the server seems wrong and I think the current approach is correct.

lib/puppet/functions/any_resources_of_type.rb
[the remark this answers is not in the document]
GabrielNagy: > Figured out why this happens. When iterating over the resources on the server, we iterate over instances of `Puppet::Parser::Resource`. When doing the same thing on the agent, we get `Puppet::Type::SomeType` instances.

Actually I think this is related to catalog _compilation_, regardless of whether it happens on the server (puppet agent -t) or agent (puppet apply). We get `Puppet::Parser::Resource` objects in both cases.

When I got `resource.type` to give me symbols, it happened during catalog _application_ on the agent, and that's where I got `Puppet::Type` instances.

So I believe in the context of this function that `resource.type` will always be a capitalized string, not a symbol. I'm inclined to keep the `to_s` coercion and the case insensitive comparison though, to be on the safe side.

lib/puppet/functions/any_resources_of_type.rb
gimmyxd: I had a concern about calling `to_s` and increase memory usage if there are a lot of resources,  but i think Ruby is smart enough and if you call `to_s` on a String, it will not create a new object as it does when calling it on other Types.

lib/puppet/functions/any_resources_of_type.rb
```diff
@@ -0,0 +1,25 @@
...
+    scope.catalog.resources.any? do |resource|
+      # We should always iterate over Puppet::Parser::Resource
+      # instances here, and documentation states that types can be
+      # strings or symbols.
+      # https://www.rubydoc.info/gems/puppet/Puppet/Resource#initialize-instance_method
+      if resource.type.to_s.casecmp(resource_type).zero? # String#casecmp? is Ruby 2.4+
```
gimmyxd, line 15: I had a concern about calling `to_s` and increase memory usage if there are a lot of resources,  but i think Ruby is smart enough and if you call `to_s` on a String, it will not create a new object as it does when calling it on other Types.

manifests/init.pp
```diff
@@ -116,6 +116,12 @@
   $wait_for_pxp_agent_exit = undef,
 ) inherits ::puppet_agent::params {
 
+  if any_resources_of_type('filebucket', { path => false }) {
```
gimmyxd, line 119: Since this a bit of an expensive check, i think this should be done after validating that the requested version is valid and only if $aio_upgrade_require is true

Dorin-Pleava:
LGTM. Tested this with PE 2019.8 and with windows 10 and windows 2016. When remote filebuckets were enabled the error was shown, after removing the remote filebucket, I could successfully upgrade to puppet 7 on windows with the module.

Dorin-Pleava (approved):

gimmyxd (approved):

PR #512, opened 2020-09-21
(MODULES-10813) Mismatched versions stops install

Dorin-Pleava:
When performing a puppet-agent upgrade with PE, the .msi package
that will get downloaded by default is the puppet-agent version
of the server node.

If a different version than the one on the master is requested,
if no pe_repo is configured and available, it will install the version
on the master.

This PR fails if the desired version and version from the .msi match.
If they do not match the install_puppet.ps1 will fail before installing
an unwanted version.

### DIFF
--- a/files/prerequisites_check.ps1
+++ b/files/prerequisites_check.ps1
@@ -22,6 +22,7 @@ param (
 Write-Log "Checking puppet-agent.msi version and expected puppet-agent version.." $Logfile
 try{
   if (!(Test-Path $Msi.FullName)) {
+    Write-Error "ERROR: File '${Msi.FullName}' does not exist"
     Write-Log "ERROR: File '${Msi.FullName}' does not exist" $Logfile
     throw
   }
@@ -43,6 +44,7 @@ try{
     $Msi_version = $record.GetType().InvokeMember( "StringData", "GetProperty", $Null, $record, 1 )
 
   } catch {
+    Write-Error "ERROR: Failed to get MSI file version: ${_}."
     Write-Log "ERROR: Failed to get MSI file version: ${_}." $Logfile
     throw
   }
@@ -50,6 +52,7 @@ try{
     Write-Log ".msi file version and expected puppet-agent version match (${Msi_version})" $Logfile
     exit 0
   } else {
+    Write-Output "ERROR: The expected puppet-agent version(${RequiredVersion}) does NOT match the .msi version ${Msi_version}.  Installation will STOP!"
     Write-Log "ERROR: The expected puppet-agent version(${RequiredVersion}) does NOT match the .msi version ${Msi_version}. Installation will STOP!" $Logfile
     throw
   }
--- a/files/prerequisites_check.ps1
+++ b/files/prerequisites_check.ps1
@@ -1,0 +1,59 @@
+<#
+.Synopsis
+  Read the ProductVersion from a Windows Installer MSI file
+.Description
+  This script will check if the versions from the .msi file and the $RequiredVersion match, to ensure that only
+  the .msi with the wanted version is installed.
+.Parameter RequiredVersion
+  The version that is required to be installed
+.Parameter Msi
+  The file name of the MSI file, or full path
+.Parameter Logfile
+  File location where the installation will log output
+#>
+param (
+    [String] $RequiredVersion,
+    [IO.FileInfo] $Msi,
+    [String] $Logfile
+)
+
+. "$PSScriptRoot\helpers.ps1"
+
+Write-Log "Checking puppet-agent.msi version and expected puppet-agent version.." $Logfile
+try{
+  if (!(Test-Path $Msi.FullName)) {
+    Write-Log "ERROR: File '${Msi.FullName}' does not exist" $Logfile
+    throw
+  }
+  $Msi_version = ""
+  try {
+    $windowsInstaller = New-Object -com WindowsInstaller.Installer
+    $database = $windowsInstaller.GetType().InvokeMember(
+        "OpenDatabase", "InvokeMethod", $Null,
+        $windowsInstaller, @($Msi.FullName, 0)
+    )
+
+    $q = "SELECT Value FROM Property WHERE Property = 'ProductVersion'"
+    $View = $database.GetType().InvokeMember(
+        "OpenView", "InvokeMethod", $Null, $database, ($q)
+    )
+
+    $View.GetType().InvokeMember("Execute", "InvokeMethod", $Null, $View, $Null)
+    $record = $View.GetType().InvokeMember( "Fetch", "InvokeMethod", $Null, $View, $Null )
+    $Msi_version = $record.GetType().InvokeMember( "StringData", "GetProperty", $Null, $record, 1 )
+
+  } catch {
+    Write-Log "ERROR: Failed to get MSI file version: ${_}." $Logfile
+    throw
+  }
+  if ($Msi_version -eq $RequiredVersion) {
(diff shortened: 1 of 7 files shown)


manifests/install/windows.pp
```diff
@@ -78,10 +97,16 @@
                               exit 0; \
                             } \
                             exit 1; }.Invoke()",
+    onlyif  => "${::system32}\\WindowsPowerShell\\v1.0\\powershell.exe \
+                  -NoProfile \
+                  -NoLogo \
+                  -NonInteractive \
+                  ${_puppet_check} ${::puppet_agent::_expected_package_version} ${_msi_location}",
```
Dorin-Pleava, line 104: If the powershell script throws errors, the `exec` command will not be executed, and the errors will not show on puppet output.

manifests/install/windows.pp
```diff
@@ -49,6 +49,25 @@
...
+    command => "${::system32}\\WindowsPowerShell\\v1.0\\powershell.exe \
+                  -NoProfile \
+                  -NoLogo \
+                  -NonInteractive \
+                  ${_puppet_check} ${::puppet_agent::_expected_package_version} ${_msi_location}",
+    returns => 0,
+    require => [
+      File[$_puppet_check]
+    ]
+  }
```
Dorin-Pleava, line 69: I added this so that puppet runs will fail in case the powershell script does not return 0.
Without this check Puppet Enterprise will not detect any problems if the `onlyif` from `exec { 'install_puppet.ps1':` will fail.

manifests/install/windows.pp
```diff
@@ -43,12 +43,37 @@
...
+  exec { 'prerequisites_check.ps1':
+    command => "${::system32}\\WindowsPowerShell\\v1.0\\powershell.exe \
+                  -NoProfile \
+                  -NoLogo \
+                  -NonInteractive \
+                  ${_prerequisites_check} ${::puppet_agent::_expected_package_version} ${_msi_location} ${_logfile}",
+    returns => 0,
+    require => [
+      File[$_prerequisites_check]
+    ]
```
GabrielNagy, line 74: I think this can be shortened to `require => File[$_prerequisites_check]`

manifests/install/windows.pp
```diff
@@ -43,12 +43,37 @@
...
+    content => file('puppet_agent/prerequisites_check.ps1')
+  }
+
+  exec { 'prerequisites_check.ps1':
+    command => "${::system32}\\WindowsPowerShell\\v1.0\\powershell.exe \
+                  -NoProfile \
+                  -NoLogo \
+                  -NonInteractive \
+                  ${_prerequisites_check} ${::puppet_agent::_expected_package_version} ${_msi_location} ${_logfile}",
+    returns => 0,
```
GabrielNagy, line 71: Is this needed? Per https://puppet.com/docs/puppet/5.5/types/exec.html#exec-attribute-returns, this defaults to 0.

manifests/install/windows.pp
```diff
@@ -81,7 +106,9 @@
     path    => $::path,
     require => [
       Puppet_agent_upgrade_error['puppet_agent_upgrade_failure.log'],
-      File[$_installps1]
+      File[$_installps1],
+      File[$_prerequisites_check],
+      Exec['prerequisites_check.ps1']
```
GabrielNagy, line 111: Isn't it enough to specify the Exec resource here? Since the Exec itself depends on `File[$_prerequisites_check]`. Just a thought.

files/prerequisites_check.ps1
```diff
@@ -0,0 +1,59 @@
...
+    Write-Log ".msi file version and expected puppet-agent version match (${Msi_version})" $Logfile
+    exit 0
+  } else {
+    Write-Log "ERROR: The expected puppet-agent version(${RequiredVersion}) does NOT match the .msi version ${Msi_version}. Installation will STOP!" $Logfile
+    throw
+  }
+} catch {
+    Write-Log "ERROR: $_" $Logfile
+    exit 1
+}
```
GabrielNagy, line 59: nit: mixing 2/4 space indentation in the catch blocks

files/prerequisites_check.ps1
```diff
@@ -0,0 +1,59 @@
...
+#>
+param (
+    [String] $RequiredVersion,
+    [IO.FileInfo] $Msi,
+    [String] $Logfile
+)
+
+. "$PSScriptRoot\helpers.ps1"
+
+Write-Log "Checking puppet-agent.msi version downloaded from PE and expected puppet-agent version.." $Logfile
```
GabrielNagy, line 22: nit: this script is also executed on non-PE installations

files/prerequisites_check.ps1
```diff
@@ -0,0 +1,59 @@
...
+    $database = $windowsInstaller.GetType().InvokeMember(
+        "OpenDatabase", "InvokeMethod", $Null,
+        $windowsInstaller, @($Msi.FullName, 0)
+    )
+
+    $q = "SELECT Value FROM Property WHERE Property = 'ProductVersion'"
+    $View = $database.GetType().InvokeMember(
+        "OpenView", "InvokeMethod", $Null, $database, ($q)
+    )
+
```
GabrielNagy, line 40: nit: GitHub doesn't show it, but there's some empty whitespace on this line

GabrielNagy (commented):
Looks good @Dorin-Pleava, just had a few comments (aside from the code remarks):
- with this PR, when attempting an upgrade with mismatched versions, this is the error that we get on the CLI:

```
Error: 'C:\Windows\system32\WindowsPowerShell\v1.0\powershell.exe -NoProfile                   -NoLogo                   -NonInteractive C:\Users\ADMINI~1\AppData\Local\Temp\prerequisites_check.ps1 6.20.0 C:\ProgramData\Puppetlabs\packages\puppet-agent-6.17.0-x64.msi C:\Users\ADMINI~1\AppData\Local\Temp\puppet-2020_10_01-11_56-installer.log' returned 1 instead of one of [0]
Error: /Stage[main]/Puppet_agent::Install::Windows/Exec[prerequisites_check.ps1]/returns: change from 'notrun' to ['0'] failed: 'C:\Windows\system32\WindowsPowerShell\v1.0\powershell.exe                   -NoProfile                   -NoLogo                   -NonInteractive                   C:\Users\ADMINI~1\AppData\Local\Temp\prerequisites_check.ps1 6.20.0 C:\ProgramData\Puppetlabs\packages\puppet-agent-6.17.0-x64.msi C:\Users\ADMINI~1\AppData\Local\Temp\puppet-2020_10_01-11_56-installer.log' returned 1 instead of one of [0]
```

Which is not at all suggestive of why the script failed. I know the script logs to a file, but we should also have something printed in the console, if possible.

- in addition to fixing this bug, the PR also refactors some existing ps1 code, would have liked to see this in a separate commit
- the commit description mentions:
```
    This PR fails if the desired version and version from the .msi match.
    If they do not match the install_puppet.ps1 will fail before installing
    an unwanted version.
```
I think you meant `This PR verifies if the...`

Dorin-Pleava:
Good points @GabrielNagy, working on the CLI error now

GabrielNagy:
The error now shows on the CLI as well, so :+1: from me: 
image

GabrielNagy (approved):

gimmyxd:
> The error now shows on the CLI as well, so 👍 from me:
> image

Isn't a bit bothering that we print an error message as a notice?
image

Can that be changed to error?

Dorin-Pleava:
It seems that even when throwing an exception in the .ps1 script, it still prints the error as a notice
image

PR #515, opened 2020-10-14
(MODULES-10818) ignore `msi_move_locked_files` on newer puppet versions

ciprianbadescu:
Ignore `msi_move_locked_files` value for puppet-agent >= 5.5.21 since
it is no longer needed (in 5.5.17 another solution was implemented in
puppet-agent for puppetres.dll:
https://github.com/puppetlabs/puppet-agent/pull/1749/files)
and it interferes with restart protection implemented in puppet-agent
for nss,.exe (https://github.com/puppetlabs/puppet-agent/pull/1912)


ciprianbadescu:
do not merge yet, we have a similar issue ongoing, we may need to apply the workaround to earlier pupet-agent versions

PR #517, opened 2020-10-21
(MODULES-10818) update README with msi_move_locked_files updates/details

ciprianbadescu:


README.markdown
```diff
@@ -287,12 +287,17 @@ With Windows packages, note that file paths in `install_options` must use backsl
 
 ##### `msi_move_locked_files`
 
-This is only applicable for Windows operating systems. There may be instances where file locks cause unncessary service restarts.  By setting to true, the module will move files prior to installation that are known to cause file locks. By default this is set to false.
+This is only applicable for Windows operating systems and for Puppet 5 prior to 5.5.17 or Puppet 6 prior to 6.8.0. There may be instances where file locks cause unncessary service restarts.  By setting to true, the module will move files prior to installation that are known to cause file locks. By default this is set to false.
```
GabrielNagy, line 290: old typo: unncessary -> unnecessary

README.markdown
```diff
@@ -287,12 +287,17 @@ With Windows packages, note that file paths in `install_options` must use backsl
...
 
 ``` puppet
   msi_move_locked_files => true
 ```
 
+In case `msi_move_locked_files` is set to `true` while upgrading to Puppet 5 following 5.5.17 or Puppet 6 following 6.8.0, Puppet can get into a state where `puppet --versions` reports the older version (5.5.16) while the package reported by windows is the new version(5.5.17). To recoved from this case `ADDLOCAL=ALL` must be added to install_options
```
GabrielNagy, line 296: typo: --versions -> --version

GabrielNagy: capitalize Windows

GabrielNagy: typo: recoved -> recover

README.markdown
```diff
@@ -365,7 +370,12 @@ There are a few known issues on Windows platforms:
...
 
-* Upgrading on most \*NIX platforms (Linux, AIX, Solaris 11) will end the run after the puppet-agent upgrade finishes. This is to avoid unexpected behavior if already loaded Ruby code happens to interact with newer code that came with the upgrade, or viceversa. If run as a daemon, Puppet will automatically start a new agent run after the upgrade finishes.
+* If the upgrade is from Puppet 5 prior to 5.5.17 or Puppet 6 prior to 6.8.0 to newer version and `msi_move_locked_files` is set to `true`, Puppet can get into a state where `puppet --versions` reports the older version (5.5.16) while the package reported by windows is the new version(5.5.17). To recoved from this case `ADDLOCAL=ALL` must be added to install_options
+``` puppet
+  install_options => ['REINSTALLMODE="maus"', 'ADDLOCAL=ALL']
```
GabrielNagy, line 375: Think this is `amus` not `maus`. This is also on line 283 and 298.

Actually it should be order-independent (https://docs.microsoft.com/en-us/windows/win32/msi/reinstallmode), but in our code we hardcode `amus` so it may be good to have the same value in the documentation.

README.markdown
```diff
@@ -365,7 +370,12 @@ There are a few known issues on Windows platforms:
 
 * MSI installation failures do not produce any error. If the install fails, puppet_agent continues to be applied to the agent. If this happens, you'll need to examine the MSI log file to determine the failure's cause. You can find the location of the log file in the debug output from either a puppet apply or an agent run; the log file name follows the pattern `puppet-<timestamp>-installer.log`.
 
-* Upgrading on most \*NIX platforms (Linux, AIX, Solaris 11) will end the run after the puppet-agent upgrade finishes. This is to avoid unexpected behavior if already loaded Ruby code happens to interact with newer code that came with the upgrade, or viceversa. If run as a daemon, Puppet will automatically start a new agent run after the upgrade finishes.
+* If the upgrade is from Puppet 5 prior to 5.5.17 or Puppet 6 prior to 6.8.0 to newer version and `msi_move_locked_files` is set to `true`, Puppet can get into a state where `puppet --versions` reports the older version (5.5.16) while the package reported by windows is the new version(5.5.17). To recoved from this case `ADDLOCAL=ALL` must be added to install_options
```
GabrielNagy, line 373: Same comments as line 296

README.markdown
```diff
@@ -365,7 +370,12 @@ There are a few known issues on Windows platforms:
...
+``` puppet
+  install_options => ['REINSTALLMODE="maus"', 'ADDLOCAL=ALL']
+```
+
+Upgrading on most \*NIX platforms (Linux, AIX, Solaris 11) will end the run after the puppet-agent upgrade finishes. This is to avoid unexpected behavior if already loaded Ruby code happens to interact with newer code that came with the upgrade, or viceversa. If run as a daemon, Puppet will automatically start a new agent run after the upgrade finishes.
```
GabrielNagy, line 378: Can you add a * at the beginning of the line so it's formatted as before

ciprianbadescu: Not sure abut this one, as it will make this \*NIX related paragraph appear as part of a Windows enumeration: 
image

GabrielNagy: Oops, missed that. Can we update the section so for each bullet we specify the affected OS, and remove the `There are a few known issues...` line? Historically it looks like the Known issues section was added specifically for Windows, but then I updated it with the *NIX issue, and now it just looks weird.

GabrielNagy (approved):

PR #525, opened 2020-11-23
(MODULES-10879) Implement configuration management

reidmv:
I'm proposing this feature as a result of not being able to find a clear, supportable, and documented way to advise a PE customer how to manage agent configuration.

This commit introduces a parameter that can be used to manage agent configuration. This parameter is restricted to managing only specifically permitted settings. Initially, only runinterval.

Without this functionality, users can only manage agent configuration by independently writing their own ini_setting resources in their own code. This is inefficient and error-prone compared to supporting basic configuration management of the agent in the official Puppet agent supported module.

By supporting basic configuration management of settings in this module it will be easier and safer to configure the Puppet agent, as well as becoming easier to document a correct and maintainable way of doing so.

### DIFF
--- a/spec/classes/puppet_agent_configure_spec.rb
+++ b/spec/classes/puppet_agent_configure_spec.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'puppet_agent::configure' do
+  let(:pre_condition) do
+    [ 'function assert_private() { }',
+      'class puppet_agent::params { $config = "/etc/puppetlabs/puppet/puppet.conf" }',
+      'include puppet_agent::params',
+      'include puppet_agent',
+    ]
+  end
+
+  context 'with empty config input' do
+    let(:pre_condition) do
+      [ super(),
+        'class puppet_agent { $config = [ ] }',
+      ]
+    end
+
+    it { is_expected.to compile }
+  end
+
+  context 'with settings input' do
+    let(:pre_condition) do
+      [ super(),
+        'class puppet_agent { $config = [{ "section" => "agent",
+                                           "setting" => "runinterval",
+                                           "value"   => "30m",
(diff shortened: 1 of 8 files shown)


reidmv:
See also: [MODULES-9260](https://tickets.puppetlabs.com/browse/MODULES-9260)

kenyon:
Just yesterday I was looking for a way to do this, specifically because I want to enable splay on the agents in my Puppet Enterprise environment.

spec/classes/puppet_agent_configure_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    let(:pre_condition) do
+      [ super(),
+        'class puppet_agent { $config = [ ] }',
+      ]
+    end
+
+    it { is_expected.to compile }
+  end
+
+  context 'with settings input' do
```
GabrielNagy, line 24: Should we have some tests that assert the settings' absence when `ensure => absent` (regardless of whether  `value` is set)?

GabrielNagy (commented):
Thanks for the PR @reidmv! I had a comment about the tests, but from a code perspective this looks good to me.

The PR tests with nightly Puppet are expected to fail, but the tests with released Puppet gems seem to be valid: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/525/checks?check_run_id=2068620749. Maybe this is because of the dependency changes in init.pp.

Also, the commit check fails due to the following:

```
	This commit summary didn't match CONTRIBUTING.md guidelines:

		Update README high-level module description

	The commit summary (i.e. the first line of the commit message) should start with one of:
		(docs)
		(docs)(DOCUMENT-<digits>)
		(packaging)
```

GabrielNagy (approved):
Thanks for updating the tests! I also ran the acceptance suite pipeline with this PR, and everything passes.

PR #532, opened 2021-01-13
(MODULES-10897) Fix GPG key typo

GabrielNagy:
Accidentaly removed the RPM part of the filename when getting the GPG
key for SLES in the install task. This commit adds it back.

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -513,10 +513,12 @@ info "Downloading Puppet $version for ${platform}..."
 case $platform in
   "SLES")
     info "SLES platform! Lets get you an RPM..."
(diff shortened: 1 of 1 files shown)


tphoney (approved):

tphoney:
Thanks for the fix. @GabrielNagy  great catch

tphoney:
@GabrielNagy why are you installing 2 keys ? puppet and puppet-20250406 
just curious

GabrielNagy:
@tphoney I think packages are not yet signed with the new key, and that's why SLES 15 failed when only importing the 2025 one. By having both imported we avoid these types of failures.

PR #533, opened 2021-01-19
Add report parameter to bolt puppet_agent::run plan/task

scbjans:
This PR adds a boolean parameter `report` to the `puppet_agent::run` plan and task. This parameter can be used to either include (default `true` for backwards compatibility) or exclude (`false`) the last_run_report in the `puppet_agent::run` task output.

### DIFF
--- a/plans/run.pp
+++ b/plans/run.pp
@@ -4,7 +4,7 @@
 # @param report Whether to include the last run report in the task output.
 plan puppet_agent::run (
   TargetSpec $targets,
-  Optional[Boolean] $report = true,
+  Boolean $report = true,
 ) {
   # Check which targets have the agent installed by checking
   # the version of the agent. No point in trying to run the
--- a/plans/run.pp
+++ b/plans/run.pp
@@ -1,8 +1,10 @@
 # Starts a Puppet agent run on the specified targets.
 # Note: This plan may cause issues when run in Puppet Enterprise.
 # @param targets The targets to start a Puppet agent run on.
+# @param report Whether to include the last run report in the task output.
 plan puppet_agent::run (
-  TargetSpec $targets
+  TargetSpec $targets,
+  Optional[Boolean] $report = true,
 ) {
   # Check which targets have the agent installed by checking
   # the version of the agent. No point in trying to run the
@@ -57,6 +59,7 @@ plan puppet_agent::run (
     'puppet_agent::run',
     $agent_results.targets,
     'Run Puppet agent',
+    'report' => $report,
     '_catch_errors' => true
   )
 
--- a/tasks/run.json
+++ b/tasks/run.json
@@ -3,7 +3,7 @@
   "parameters": {
     "report": {
       "description": "Whether to include the last run report in the output.",
-      "type": "Optional[Boolean]",
+      "type": "Boolean",
       "default": true
     }
   },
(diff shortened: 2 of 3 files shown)


plans/run.pp
```diff
@@ -1,8 +1,10 @@
 # Starts a Puppet agent run on the specified targets.
 # Note: This plan may cause issues when run in Puppet Enterprise.
 # @param targets The targets to start a Puppet agent run on.
+# @param report Whether to include the last run report in the task output.
 plan puppet_agent::run (
-  TargetSpec $targets
+  TargetSpec $targets,
+  Optional[Boolean] $report = true,
```
beechtom, line 7:
```suggestion
  Boolean $report = true,
```

tasks/run.json
```diff
@@ -1,5 +1,11 @@
 {
   "description": "Run the Puppet agent. This task may cause problems if run in Puppet Enterprise.",
-  "parameters": {},
+  "parameters": {
+    "report": {
+      "description": "Whether to include the last run report in the output.",
+      "type": "Optional[Boolean]",
```
beechtom, line 6:
```suggestion
      "type": "Boolean",
```

tasks/run.rb
```diff
@@ -29,6 +29,10 @@ def disabled?(lockfile)
       File.exist?(lockfile)
     end
 
+    def include_report?
+      JSON.parse(STDIN.read)['report']
+    end
```
beechtom, line 34: This obfuscates where params are coming from a bit... in the interest of making it easier to add additional params in the future, let's parse params at the beginning of the `run` method. From there, we can pass `params['report']` to the `get_result_from_report` method.

beechtom (commented):
@scbjans Thanks for the PR! Overall this looks good, just a couple minor suggestions.

lucywyman:
@scbjans What are you trying to achieve this parameter? I think a plan that does something like this (pseudocode) would have the same effect:
```
plan bolt (
  TargetSpec $targets
){
  $result = run_plan('puppet_agent::run', $targets)
  return filter($result) |$key, $value| { $key != "report" }
}
```

Given we haven't heard other people asking for this, I'm inclined to think a more custom solution like the plan above might be more appropriate.

scbjans:
Hi @lucywyman , thanks for your suggestion, I haven't thought of that.

What I'm trying to achieve is the following:
As of now we still have a bit of mcollective/choria/NATS left; which we're migrating to Bolt. One of the tasks which is still left to migrate is the scheduling of Puppet runs across availability zones.
At defined times a Bolt Docker container is spun up which triggers a Puppet agent run on several hundreds of nodes simultaneously per AZ. We're not interested in the last run report (by default); only if the run was successful or not.
To me, it doesn't make sense to send data over the network from which you'll now upfront that it will be discarded; nor should it be cached in the Bolt container to be discarded when all Bolt tasks are completed.
Therefore I've introduced this flag to be able to exclude this data upfront.

lucywyman:
That use case makes total sense. My hesitation with adding this is that it adds complexity, and prevents us from adding a `report` flag with another meaning in the future (or what if we want to do other methods of filtering output later down the road, and need to account for this?). Another option if you're worried about sending a lot of data over the wire would be to fork and patch this, and either publish your own module or just keep it locally. We aren't anticipating changing either the plan or task very much, so I don't expect a fork would often get out of sync. Then we can open an issue in the Bolt repo with this use case and see if anyone has a similar problem and is looking for a solution.

Does that seem reasonable? I mostly don't want us to paint ourselves into a corner (as we've done in the past!).

scbjans:
Alright, I'll keep this to ourselves then, until (and if) this will be available in the future.
Thanks for thinking along!

PR #536, opened 2021-01-27
(MODULES-10909) Commands retry on network connectivity failures 

adrianiurca:
Before this commit, the install tasks (`install_shell.sh` and `install_powershell.ps1`) exited at the first network connectivity failure. This change adds a retry mechanism to the `wget`, `curl`, `fetch`, `perl`, `yum`, `zypper`, `apt-get` (Linux) and `System.Net.Webclient.DownloadFile` (Windows) commands which waits one second between attempts. The number of attempts is `5` by default but can be configured via the new `retry` setting.

### DIFF
--- a/tasks/install_powershell.json
+++ b/tasks/install_powershell.json
@@ -33,6 +33,11 @@
     "stop_service": {
       "description": "Whether to stop the puppet agent service after install",
       "type": "Optional[Boolean]"
+    },
+    "retry": {
+      "description": "The number of retries in case of network connectivity failures",
+      "type": "Optional[Integer]",
+      "default": 5
     }
   }
 }
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -4,7 +4,8 @@ Param(
   [String]$collection = 'puppet',
   [String]$windows_source = 'https://downloads.puppet.com',
   [String]$install_options = 'REINSTALLMODE="amus"',
-  [Bool]$stop_service = $False
+  [Bool]$stop_service = $False,
+  [Int]$retry = 5
 )
 # If an error is encountered, the script will stop instead of the default of "Continue"
 $ErrorActionPreference = "Stop"
@@ -104,20 +105,32 @@ $msi_dest = Join-Path ([System.IO.Path]::GetTempPath()) "puppet-agent-$arch.msi"
 $install_log = Join-Path ([System.IO.Path]::GetTempPath()) "$date_time_stamp-puppet-install.log"
 
 function DownloadPuppet {
-  Write-Verbose "Downloading the Puppet Agent for Puppet Enterprise on $env:COMPUTERNAME..."
-
+  Write-Output "Downloading the Puppet Agent installer on $env:COMPUTERNAME..."
   $webclient = New-Object system.net.webclient
 
   try {
     $webclient.DownloadFile($msi_source,$msi_dest)
   }
   catch [System.Net.WebException] {
-    # If we can't find the msi, then we may not be configured correctly
-    if($_.Exception.Response.StatusCode -eq [system.net.httpstatuscode]::NotFound) {
-        Throw "Failed to download the Puppet Agent installer: $msi_source"
+    For ($attempt_number = 1; $attempt_number -le $retry; $attempt_number++) {
+      try {
+        Write-Output "Retrying... [$attempt_number/$retry]"
+        $webclient.DownloadFile($msi_source,$msi_dest)
+        break
+      }
+      catch [System.Net.WebException] {
+        if($attempt_number -eq $retry) {
+          # If we can't find the msi, then we may not be configured correctly
+          if($_.Exception.Response.StatusCode -eq [system.net.httpstatuscode]::NotFound) {
+            Throw "Failed to download the Puppet Agent installer: $msi_source"
+          }
+
+          # Throw all other WebExceptions
+          Throw $_
+        }
+        Start-Sleep -s 1
+      }
     }
-    # Throw all other WebExceptions
-    Throw $_
   }
 }
 
--- a/tasks/install_shell.json
+++ b/tasks/install_shell.json
@@ -34,6 +34,11 @@
     "stop_service": {
       "description": "Whether to stop the puppet agent service after install",
       "type": "Optional[Boolean]"
+    },
+    "retry": {
+      "description": "The number of retries in case of network connectivity failures",
+      "type": "Optional[Integer]",
+      "default": 5
     }
   },
   "files": ["facts/tasks/bash.sh"]
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -122,6 +122,12 @@ else
   fi
 fi
 
+if [ -n "$PT_retry" ]; then
+  retry=$PT_retry
+else
+  retry=5
+fi
+
 # Track to handle puppet5 to puppet6
 if [ -f /opt/puppetlabs/puppet/VERSION ]; then
   installed_version=`cat /opt/puppetlabs/puppet/VERSION`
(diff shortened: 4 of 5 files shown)


lucywyman:
The ticket specifically calls out Windows hiccups, but this doesn't seem to modify the PowerShell task as well? It seems like it'd be useful, for us and users, to add it there too.

tasks/install_powershell.ps1
```diff
@@ -107,18 +107,25 @@ function DownloadPuppet {
...
-
-  try {
-    $webclient.DownloadFile($msi_source,$msi_dest)
-  }
-  catch [System.Net.WebException] {
-    # If we can't find the msi, then we may not be configured correctly
-    if($_.Exception.Response.StatusCode -eq [system.net.httpstatuscode]::NotFound) {
-        Throw "Failed to download the Puppet Agent installer: $msi_source"
+  
+  $a = 0
```
luchihoratiu, line 111: nit: Something like `$attempt_number` instead of `$a`? It fits better with the rest of the code and would be easier to read the code.

tasks/install_powershell.ps1
```diff
@@ -107,18 +107,25 @@ function DownloadPuppet {
...
+      if($_.Exception.Response.StatusCode -eq [system.net.httpstatuscode]::NotFound) {
+          Throw "Failed to download the Puppet Agent installer: $msi_source"
+      }
+      # Throw all other WebExceptions
+      if($a -eq $retry) {
+        Throw $_
+      }
+    }
+    $a++
+  } while($a -le $retry)
```
luchihoratiu, line 128: Would `$a` ever be equal to `$retry` and throw the exception?

tasks/install_powershell.json
```diff
@@ -33,6 +33,11 @@
     "stop_service": {
       "description": "Whether to stop the puppet agent service after install",
       "type": "Optional[Boolean]"
+    },
+    "retry": {
+      "description": "The number of download tries",
```
luchihoratiu, line 38: I would go with something more generic and explicit like: `The number of retries in case of network connectivity failures`.
With this in mind, I would also add retry logic in the `install_file` method to every `yum`, `zypper` and `apt-get` commands. Failures are seen there also in the collected data attached to the ticket.
Reminder: add this whole retry parameter to the `install.json` file also.

tasks/install_shell.json
```diff
@@ -34,6 +34,11 @@
     "stop_service": {
       "description": "Whether to stop the puppet agent service after install",
       "type": "Optional[Boolean]"
+    },
+    "retry": {
+      "description": "The number of download tries",
```
luchihoratiu, line 39: Same as `install_powershell.json`

tasks/install_shell.sh
```diff
@@ -122,6 +122,12 @@ else
   fi
 fi
 
+if [ -n "$PT_retry" ]; then
+  retry=$PT_retry
+else
+  retry=5
+fi
+
```
luchihoratiu, line 130: I would expect this to not be needed since there is a default value set in the `*.json` file (needs testing). If by any chance the `default` parameter is not being taken into consideration, then the powershell task should have this same logic in it and the `default` parameter from `*.json` files removed. It would only be confusing to have default value logic in tasks and `*.json` files.

tasks/install_shell.sh
```diff
@@ -315,6 +337,8 @@ do_wget() {
   if test $? -eq 0; then
     critical "ERROR 404"
     unable_to_retrieve_package
+  else
+    rc=retry_cmd "wget -O "$2" "$1" 2>$tmp_stderr"
```
luchihoratiu, line 341: shouldn't `"` be escaped inside the string?
`"wget -O \"$2\" \"$1\" 2>$tmp_stderr"`

tasks/install_shell.sh
```diff
@@ -304,6 +310,22 @@ capture_tmp_stderr() {
 
 trap "rm -f $tmp_stderr; rm -rf $tmp_dir; exit $1" 1 2 15
 
+retry_cmd() {
+  a=0
+  while test $a -lt $retry; do
+    log "Retrying... [$((a + 1))/$retry]"
+    eval $1
+    rc=$?
+    info $rc
```
luchihoratiu, line 319: How about something like `info "Return code: ${rc}"` and moved after checking if it's 0 so we get prints only about failing commands?

tasks/install_shell.sh
GabrielNagy: might be simpler to use simple quotes:

`"wget -O '$2' '$1' 2>$tmp_stderr"`

tasks/install_powershell.ps1
adrianiurca: I should because the while condition is `$a -le $retry`(less or equal). So when $a hits the same value as $retry, the loop will still iterate once again.

tasks/install_powershell.ps1
adrianiurca: good observation!

tasks/install_powershell.json
adrianiurca: good observation!

tasks/install_shell.json
adrianiurca: okay

tasks/install_shell.sh
adrianiurca: I removed the `default` value from `install_*json` files

tasks/install_shell.sh
adrianiurca: done

luchihoratiu:
Added a commit with some changes after discussing with @adrianiurca. Manually tested all the changes using Bolt on Microsoft Windows 2019, SLES 11, SLES 15 and Ubuntu 20.04 and seems to be working as expected.

lucywyman (commented):
Tests:
- [x] Accepts retry of 0 and does not retry
- [x] Accepts negative number and does not retry
- [x] Retries when running the PowerShell task
- [x] Retries when running the nix task

lucywyman (approved):

tasks/install_shell.sh
```diff
@@ -350,17 +383,31 @@ do_curl() {
...
+  fi
+
   return 0
 }
 
 # do_perl URL FILENAME
 do_perl() {
   info "Trying perl..."
-  perl -e 'use LWP::Simple; getprint($ARGV[0]);' "$1" > "$2" 2>$tmp_stderr
+  run_cmd "perl -e 'use LWP::Simple; getprint($ARGV[0]);' '$1' > '$2' 2>$tmp_stderr"
```
GabrielNagy, line 408: $ARGV might get interpolated by bash here since the outer quotes are now double

tasks/install_shell.sh
```diff
@@ -350,17 +383,31 @@ do_curl() {
...
-  test $? -ne 0 && return 1
+  run_cmd "fetch -o '$2' '$1' 2>$tmp_stderr"
+  rc=$?
+
+  # check for 404
+  grep "404 Not Found" $tmp_stderr 2>&1 >/dev/null
+  if test $? -eq 0; then
+    critical "ERROR 404"
+    unable_to_retrieve_package
+  fi
```
GabrielNagy, line 394: Curious why did we add a 404 check here? Previous code didn't seem to have it

luchihoratiu: Confirmation that it is a valid error:
```
> eval "perl -e 'use LWP::Simple; getprint(\$ARGV[0]);' 'http://nightlies.puppet.com/bad_package_name' > 'output_file' 2>errors"
> cat errors
404 Not Found <URL:http://nightlies.puppet.com/bad_package_name>
```
I didn't research why/if it was chosen this way but since it seems like a valid error and the task would fail anyway with a more generic error, then I think it would be nice to have them all aligned and have this explicit case covered here also.

tasks/install_shell.sh
luchihoratiu: Awesome catch! Tested and updated.

luchihoratiu:
Squashed the commits and updated the commit message/description.

tasks/install_shell.sh
GabrielNagy: You're breathtaking!

GabrielNagy (approved):

PR #537, opened 2021-02-03
(MODULES-9798) Add Timeout Parameter for the Current Puppet Run

phil4business:
- (Support-42796) Add ```wait_for_puppet_run``` parameter to configure the timeout value for the current puppet run.
Previously, if the puppet agent was to be upgraded but the current run took longer than 120000 ms (2 minutes) to complete, the agent upgrade script (install_puppet.ps1) would fail. This would happen even if the agent run would have otherwise completed without error.

- Fixed ```wait_for_pxp_agent_exit``` parameter tests. The parameter accepts an Int32, but the tests were passing booleans.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -46,6 +46,7 @@
         - [`install_options`](#install_options)
         - [`msi_move_locked_files`](#msi_move_locked_files)
         - [`wait_for_pxp_agent_exit`](#wait_for_pxp_agent_exit)
+        - [`wait_for_puppet_run`](#wait_for_puppet_run)
     - [Plans](#plans)
       - [`puppet_agent::run`](#puppet_agentrun)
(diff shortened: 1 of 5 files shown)


GabrielNagy (commented):
Hi @phil4business, thanks for your contribution! :tada:

The PR looks good, good catch on the spec tests. Also, we really appreciate the effort to keep the documentation up-to-date.

I found a JIRA ticket that relates to this: https://tickets.puppetlabs.com/browse/MODULES-9798

To get PR checks to pass, can you squash your commits and prepend the ticket ID to your commit? Something like `(MODULES-9798) Add Timeout Parameter for the Current Puppet Run`. Let me know if you need any help with this.

GabrielNagy (approved):

PR #542, opened 2021-02-23
(MODULES-10925) Added facts_diff task

gimmyxd:
This adds a new task `puppet_agent::facts_diff` that
can be used to display facts differences between
Facter 3 and Facter 4 using `puppet facts diff` action.

### DIFF
--- a/files/rb_task_helper.rb
+++ b/files/rb_task_helper.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module PuppetAgent
+  module RbTaskHelper
+    private
+
+    def error_result(error_type, error_message)
+      {
+        '_error' => {
+          'msg' => error_message,
+          'kind' => error_type,
+          'details' => {},
+        },
+      }
+    end
+
+    def puppet_bin_present?
+      File.exist?(puppet_bin)
+    end
+
+    # Returns the path to the Puppet agent executable
+    def puppet_bin
+      @puppet_bin ||= if Puppet.features.microsoft_windows?
+                        puppet_bin_windows
+                      else
+                        '/opt/puppetlabs/bin/puppet'
+                      end
+    end
+
+    # Returns the path to the Puppet agent executable on Windows
+    def puppet_bin_windows
+      require 'win32/registry'
+
+      install_dir = begin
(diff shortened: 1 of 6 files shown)


files/rb_task_helper.rb
```diff
@@ -0,0 +1,58 @@
...
+      {
+        '_error' => {
+          'msg' => error_message,
+          'kind' => error_type,
+          'details' => {},
+        },
+      }
+    end
+
+    def puppet_bin_present?
```
lucywyman, line 16: I see this is pulled from the [run task](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/tasks/run.rb#L20) - any chance we can use this file as a helper there too and remove the function definitions from there? That way if these change at all we don't have to change it in two places, or end up with divergent implementations.

tasks/facts_diff.rb
```diff
@@ -0,0 +1,37 @@
...
+
+params = JSON.parse(STDIN.read)
+require_relative File.join(params['_installdir'], 'puppet_agent', 'files', 'rb_task_helper.rb')
+
+module PuppetAgent
+  class FactsDiff
+    include PuppetAgent::RbTaskHelper
+
+    def run
+      unless puppet_bin_present?
```
lucywyman, line 15: This also seems like it should verify the agent version is `6.x` so that the command behaves as expected instead of returning facts for the target (which is what happens with 5.x versions). It's actually also unclear to me if this subcommand is supported in 7.x - it's only documented in [6.x documentation](https://puppet.com/docs/puppet/7.4/man/facts.html), but seems to work in 7.x?

gimmyxd: The `puppet facts diff` command is only available in puppet 6(>= 6.20.0). Added a check for that

gimmyxd:
tests are failing because of: https://tickets.puppetlabs.com/browse/FACT-2950

Dorin-Pleava (approved):
Tested with a puppet enterprise deployment and the puppet_agent::facts_diff task run successfully on a ubuntu node.
image

joshcooper:
I filed https://tickets.puppetlabs.com/browse/FACT-2952 about the lsbmajdistrelease fact

PR #544, opened 2021-03-09
Work with Puppet 7

MikaelSmith:
This module was updated to claim support for Puppet 7, but stdlib
doesn't support Puppet 7 until it's 7.0 release. That results in errors
with Puppetfile Resolver that can't be resolved for Puppet 7
```
Puppetfile Resolver could not find compatible versions for possibility named "stdlib":
  In Puppetfile:
    puppetlabs-stdlib =7.0.0

    puppetlabs-puppet_agent =4.4.0 was resolved to Forge puppetlabs-puppet_agent-4.4.0, which depends on
      puppetlabs-stdlib >= 5.1.0 < 7.0.0Puppet
```

Adds stdlib 7 to the dependency range.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -122,7 +122,7 @@
     }
   ],
   "dependencies": [
-    {"name":"puppetlabs-stdlib","version_requirement":">= 5.1.0 < 7.0.0"},
+    {"name":"puppetlabs-stdlib","version_requirement":">= 5.1.0 < 8.0.0"},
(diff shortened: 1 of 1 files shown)


GabrielNagy:
Thanks for the PR @MikaelSmith, this should be superseded by https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/545 which bumped more dependencies, so I'm closing this.

PR #547, opened 2021-03-23
(MODULES-10979) Release prep for 4.5.0

GabrielNagy:
Update supported platforms in metadata.json, bump version and add changelog.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
   "name": "puppetlabs-puppet_agent",
-  "version": "4.4.0",
+  "version": "4.5.0",
   "author": "puppetlabs",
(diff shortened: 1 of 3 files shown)


metadata.json
```diff
@@ -118,7 +116,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.7.0 < 8.0.0"
+      "version_requirement": ">= 6.0.0 < 8.0.0"
```
gimmyxd, line 119: We should still let people upgrade their older nodes using this module. So I think we should not update this.

mihaibuzgau (approved):

PR #550, opened 2021-04-15
(MODULES-11048) task to remove local filebucket

gimmyxd:
This adds a task that removes the local filebucket
which was disabled by default in Puppet 7.
The location of filebucket is determined using the
`clientbucketdir` puppet config.

```
❯ bolt task show puppet_agent::delete_local_filebucket
puppet_agent::delete_local_filebucket
  Removes the local filebucket

Usage
  bolt task run puppet_agent::delete_local_filebucket --targets <targets>
  [force=<value>]

Parameters
  force  Optional[Boolean]
    ignore nonexistent files and errors
```

Tested on Windows:
```
$ ls  /cygdrive/c/ProgramData/PuppetLabs/puppet/cache/clientbucket
ls: cannot access /cygdrive/c/ProgramData/PuppetLabs/puppet/cache/clientbucket: No such file or directory

❯ bolt task run puppet_agent::delete_local_filebucket
 {"_error"=>{"msg"=>"Errno::ENOENT: No such file or directory @ apply2files - C:/ProgramData/PuppetLabs/puppet/cache/clientbucket", "kind"=>"puppet_agent/cannot-remove-error", "details"=>{}}}

❯ bolt task run puppet_agent::delete_local_filebucket force=true
   {:success=>true}

$ mkdir -p /cygdrive/c/ProgramData/PuppetLabs/puppet/cache/clientbucket

❯ bolt task run puppet_agent::delete_local_filebucket
  {:success=>true}
  
$ ls  /cygdrive/c/ProgramData/PuppetLabs/puppet/cache/clientbucket
ls: cannot access /cygdrive/c/ProgramData/PuppetLabs/puppet/cache/clientbucket: No such file or directory

```

### DIFF
--- a/tasks/delete_local_filebucket.rb
+++ b/tasks/delete_local_filebucket.rb
@@ -1,0 +1,63 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+# frozen_string_literal: true
+
+require 'json'
+require 'puppet'
+
+params = JSON.parse(STDIN.read)
+force = params['force']
+require_relative File.join(params['_installdir'], 'puppet_agent', 'files', 'rb_task_helper.rb')
+
+module PuppetAgent
+  class DeleteLocalFilebucket
+    include PuppetAgent::RbTaskHelper
+
+    def initialize(force)
+      @force = force
+    end
+
+    def run
+      return error_result(
+        'puppet_agent/no-puppet-bin-error',
+        "Puppet executable '#{puppet_bin}' does not exist",
+      ) unless puppet_bin_present?
+
+      begin
+        path = clientbucketdir
+        if path && !path.empty? && (File.directory?(path) || force)
+          FileUtils.rm_r(Dir.glob("#{path}/*"), secure: true, force: force)
+          { "success": true }
+        else
+          error_result(
+          'puppet_agent/cannot-remove-error',
+          "clientbucketdir: '#{path}' does not exist or is not a directory"
+        )
+        end
+      rescue StandardError => e
+        error_result(
+          'puppet_agent/cannot-remove-error',
+          "#{e.class}: #{e.message}"
+        )
+      end
+    end
+
+    private
+
+    def clientbucketdir
+      options = {
+        failonfail: false,
+        override_locale: false,
+      }
+
+      command = "#{puppet_bin} config print clientbucketdir"
+      Puppet::Util::Execution.execute(command, options).strip
+    end
+
+    attr_reader :force
+  end
+end
+
+if __FILE__ == $PROGRAM_NAME
+  task = PuppetAgent::DeleteLocalFilebucket.new(force)
+  puts task.run
+end
--- a/tasks/delete_local_filebucket.json
+++ b/tasks/delete_local_filebucket.json
@@ -1,0 +1,10 @@
+{
+  "description": "Removes the local filebucket",
+  "parameters": {
+    "force": {
+      "description": "ignore nonexistent files and errors",
+      "type": "Optional[Boolean]"
+    }
+  },
+  "files": ["puppet_agent/files/rb_task_helper.rb"]
+}


tasks/delete_local_filebucket.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    private
+
+    def clientbucketdir
+      options = {
+        failonfail: false,
+        override_locale: false,
+      }
+
+      command = "#{puppet_bin} config print clientbucketdir"
```
gimmyxd, line 44: @joshcooper thoughts on this? Could this not return the correct path if `clientbucketdir` is set in a different section?

gimmyxd: `clientbucketdir` can also be found without shelling out to puppet config print.
```
Puppet.initialize_settings
Puppet[:clientbucketdir]
```

GabrielNagy: Looks like `clientbucketdir` is defined under the `agent` section in [defaults.rb](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/defaults.rb#L1734). However, the puppet filebucket application doesn't enforce a specific run_mode, so it will ignore the value of `clientbucketdir` if set under `[agent]`.

Using this logic, I think we can rely on the `initialize_settings` approach as it would be a bit faster.

joshcooper: One downside with `initialize_settings` is it can only be called once:

```shell
irb(main):001:0> require 'puppet'
=> true
irb(main):002:0> Puppet.initialize_settings
=> {}
irb(main):003:0> Puppet.initialize_settings
Traceback (most recent call last):
       16: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
       15: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/cli.rb:30:in `dispatch'
       14: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
       13: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
       12: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
       11: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/cli.rb:494:in `exec'
       10: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/cli/exec.rb:28:in `run'
        9: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/cli/exec.rb:63:in `kernel_load'
        8: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/bundler-2.2.16/lib/bundler/cli/exec.rb:63:in `load'
        7: from /home/josh/.rbenv/versions/2.7.3/bin/irb:23:in `<top (required)>'
        6: from /home/josh/.rbenv/versions/2.7.3/bin/irb:23:in `load'
        5: from /home/josh/.rbenv/versions/2.7.3/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        4: from (irb):3
        3: from /home/josh/work/puppet/lib/puppet.rb:153:in `initialize_settings'
        2: from /home/josh/work/puppet/lib/puppet.rb:178:in `do_initialize_settings_for_run_mode'
        1: from /home/josh/work/puppet/lib/puppet/settings.rb:305:in `initialize_global_settings'
Puppet::DevError (Attempting to initialize global default settings more than once!)
```

So it seems like there could be issues if this task was called twice in a plan (seems unlikely) or some other task also called `initialize_settings`?

I'd be inclined to just shell out.

tasks/delete_local_filebucket.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def run
+      return error_result(
+        'puppet_agent/no-puppet-bin-error',
+        "Puppet executable '#{puppet_bin}' does not exist",
+      ) unless puppet_bin_present?
+
+      begin
+        FileUtils.remove_dir(clientbucketdir, force)
```
ciprianbadescu, line 26: It may not worth fixing it, but if the `clientbucketdir` is removed, the puppet command listing local filebuckets may error:
```
scrawny-tumbler:~ # find /opt/puppetlabs/puppet/cache/clientbucket/                 
find: `/opt/puppetlabs/puppet/cache/clientbucket/': No such file or directory
scrawny-tumbler:~ # puppet filebucket -l list
Error: Could not run: File not found
scrawny-tumbler:~ # mkdir /opt/puppetlabs/puppet/cache/clientbucket/
scrawny-tumbler:~ # puppet filebucket -l list
scrawny-tumbler:~ # puppet filebucket backup /etc/passwd --local
scrawny-tumbler:~ #
```

ciprianbadescu (approved):
good to merge, one optional comment

tasks/delete_local_filebucket.rb
gimmyxd: Seems that puppet will create the directory again on the next run:
```
Debug: /File[/opt/puppetlabs/puppet/cache/clientbucket]: Adding autorequire relationship with File[/opt/puppetlabs/puppet/cache]
...
Debug: /File[/opt/puppetlabs/puppet/cache/clientbucket]/ensure: created
```

GabrielNagy: Maybe we should just remove the directory contents then?

PR #551, opened 2021-04-15
(MODULES-10996) Fix SLES 11 PE upgrades

GabrielNagy:
SLES 11 can no longer be upgraded in PE by installing from the repos. To work around this, if we're SLES 11 and PE, download the package and install it directly using rpm, regardless of the value of `manage_repo`. This is the same approach we take for AIX, macOS and Windows. Because zypper is left in a semi-broken state if our pe_repo is installed, make sure we remove it and don't install it again.

The GPG keys will continue to be imported.

For non-PE agents, installing from the FOSS repos should still be possible, so nothing should change on that part. Additional tests were added to assert the behavior of FOSS vs PE.

### DIFF
--- a/files/rpm_gpg_import_check.sh
+++ b/files/rpm_gpg_import_check.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# shellcheck disable=SC2086
+
+ACTION=$1
+GPG_HOMEDIR=$2
+GPG_KEY_PATH=$3
+
+GPG_ARGS="--homedir $GPG_HOMEDIR --with-colons"
+GPG_BIN=$(command -v gpg || command -v gpg2)
+
+if [ -z "${GPG_BIN}" ]; then
+  echo Could not find a suitable gpg command, exiting...
+  exit 1
+fi
+
+GPG_PUBKEY=gpg-pubkey-$("${GPG_BIN}" ${GPG_ARGS} "${GPG_KEY_PATH}" 2>&1 | grep ^pub | cut -d':' -f5 | cut --characters=9-16 | tr '[:upper:]' '[:lower:]')
+
+if [ "${ACTION}" = "check" ]; then
+  # This will return 1 if there are differences between the key imported in the
+  # RPM database and the local keyfile. This means we need to purge the key and
+  # reimport it.
+  diff <(rpm -qi "${GPG_PUBKEY}" | "${GPG_BIN}" ${GPG_ARGS}) <("${GPG_BIN}" ${GPG_ARGS} "${GPG_KEY_PATH}")
+elif [ "${ACTION}" = "import" ]; then
+  (rpm -q "${GPG_PUBKEY}" && rpm -e --allmatches "${GPG_PUBKEY}") || true
+  rpm --import "${GPG_KEY_PATH}"
+fi
--- a/manifests/osfamily/redhat.pp
+++ b/manifests/osfamily/redhat.pp
@@ -75,17 +75,11 @@ class puppet_agent::osfamily::redhat{
       $_sslclientkey_path = undef
     }
 
-    # Fedora doesn't ship with a gpg binary, only gpg2
-    if $::operatingsystem == 'Fedora' {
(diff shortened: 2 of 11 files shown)


GabrielNagy:
Updated the GPG key import checks, spec tests will also need to be updated.

files/gpg_import_check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+GPG_HOMEDIR=$1
+GPG_KEY_PATH=$2
+
+GPG_PUBKEY=gpg-pubkey-$(echo $(gpg --homedir "${GPG_HOMEDIR}" --with-colons "${GPG_KEY_PATH}" 2>&1 | grep ^pub | awk -F ':' '{print $5}' | cut --characters=9-16 | tr '[:upper:]' '[:lower:]'))
```
joshcooper, line 6: nit, since you're using `cut` to drop the first 8 characters, maybe use `cut` to print the 5th field instead of `awk`? 

```shell
 ... | grep ^pub | cut -d: -f5 | cut --characters=9-16 | ...
```

GabrielNagy: Yeah, I had to compute the pubkey here since the script didn't like being called with this interpolation. I would have liked to get rid of it in the puppet manifests, but we also need it there.

ciprianbadescu: `gpg --homedir "${GPG_HOMEDIR}" --with-colons` could be extracted to a variable for better readibility

manifests/osfamily/redhat.pp
```diff
@@ -123,11 +113,25 @@
...
+      unless    => "${::env_temp_variable}/gpg_import_check.sh ${gpg_homedir} ${legacy_gpg_path}",
+      require   => File[$legacy_gpg_path],
+      logoutput => 'on_failure',
+    }
+    -> exec { "import-${keyname}":
       path      => '/bin:/usr/bin:/sbin:/usr/sbin',
-      command   => "rpm --import ${gpg_path}",
-      unless    => "rpm -q ${gpg_pubkey}",
+      command   => "rpm -q ${gpg_pubkey} && rpm -e --allmatches ${gpg_pubkey}; rpm --import ${gpg_path}",
+      unless    => "${::env_temp_variable}/gpg_import_check.sh ${gpg_homedir} ${gpg_path}",
```
ciprianbadescu, line 134: Could we move all this gpg key mangling into a script `gpg_helper.sh`. If `unless` parameter is still useful, the same script can be called with arguments.

files/rpm_gpg_import_check.sh
```diff
@@ -0,0 +1,19 @@
...
+# shellcheck disable=SC2086
+
+ACTION=$1
+GPG_HOMEDIR=$2
+GPG_KEY_PATH=$3
+
+GPG_ARGS="--homedir $GPG_HOMEDIR --with-colons"
+GPG_PUBKEY=gpg-pubkey-$(gpg ${GPG_ARGS} "${GPG_KEY_PATH}" 2>&1 | grep ^pub | cut -d':' -f5 | cut --characters=9-16 | tr '[:upper:]' '[:lower:]')
+
+if [ "${ACTION}" = "check" ]; then
```
ciprianbadescu, line 11: nit: either `=` or `==` should be used in both action checks

ciprianbadescu (approved):
tested on sles11/pe 2019.8.5:
- verified that install shell script -> fails
- install puppet from rpm and try to upgrade using 4.5.0 version of pa_module -> fails
- upgrade using this PR version of pa_module -> success

Rewerson:
@GabrielNagy what about `/tmp` with `noexec` mount option? Suitable for shared hosting servers for security reasons.
So, got an error in `puppet_agent/manifests/osfamily/redhat.pp:113`:
`Could not evaluate: '/tmp/rpm_gpg_import_check.sh' is not executable`

ciprianbadescu:
It should be fixed by: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/557

PR #557, opened 2021-04-27
(MODULES-11057) avoid temporary file execution

ciprianbadescu:
Last module version saves a shell script to `/tmp` location
in order to execute it later.

This can interfere with various configurations (eg. files
from /tmp cannot be executed) leading to module failure.

With this change, the script will be executed dirrectly from
command line (like `/bin/bash -c "<script content>"`)


BobosilaVictor (approved):

gimmyxd (approved):
Tested this with subsequent upgrades to 6.21.1 and 6.22.1 and it works properly.
Tested upgrade to 6.21.1 with 4.6.0 and then to 6.22.1 with this ref.

PR #559, opened 2021-05-07
(MODULES-11066) Add support for running puppet_agent::install in noop mode

beechtom:
This enables running the `puppet_agent::install` task in noop mode. When
the task is run in noop mode, it executes up until the puppet-agent
package is downloaded and installed.

This work enables adding support for running plans in noop mode in Bolt,
which uses the `puppet_agent::install` task as the default
puppet_library plugin when running an `apply_prep`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [4.7.0] - 2021-05-12
+
+### Summary
+Support running `puppet_agent::install` task in no-operation mode.
+
+### Features
+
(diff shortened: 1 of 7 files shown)


lucywyman (changes requested):
There's a line to download the GPG key for SLES on line 563 of the shell task that needs to be wrapped in a noop check.

There are also some info messages that are printed that say the package is being downloaded even when it's not, which even though is still a 'noop' is now untrue. I think those can be moved to inside the noop check, and a `msg` variable can be set in the case statement to handle the custom messages (or you can try to just use the $platform variable, whichever seems better).

- [x] Does not install the agent on Windows
- [x] Does not install the agent on *nix
- [x] Still validates the provided URL and version
- [ ] No other operations in the task are at risk of executing that shouldn't be run in noop

lucywyman (approved):

GabrielNagy (approved):

PR #560, opened 2021-05-13
(PE-31118) add MacOS 11 support

ciprianbadescu:
Most of the changes are caused by the change in major product version.
Until now, the major product version was 10.15, 10.14, etc.

Starting with MacOS 11, the major version was incremented and we
started to use only the first number (eg. 11) as major version

### DIFF
--- a/manifests/osfamily/darwin.pp
+++ b/manifests/osfamily/darwin.pp
@@ -1,10 +1,11 @@
 class puppet_agent::osfamily::darwin{
   assert_private()
 
-  if $::macosx_productversion_major =~ /^11\./ {
-    $productversion_major = '11'
-  } else {
+  if $::macosx_productversion_major =~ /^10\./ {
     $productversion_major = $::macosx_productversion_major
+  } else {
+    $productversion_array = split($::macosx_productversion_major, '[.]')
+    $productversion_major = $productversion_array[0]
   }
 
   if $::puppet_agent::absolute_source {
--- a/manifests/osfamily/darwin.pp
+++ b/manifests/osfamily/darwin.pp
@@ -1,19 +1,25 @@
 class puppet_agent::osfamily::darwin{
   assert_private()
 
+  if $::macosx_productversion_major =~ /^11\./ {
+    $productversion_major = '11'
+  } else {
+    $productversion_major = $::macosx_productversion_major
+  }
+
   if $::puppet_agent::absolute_source {
     $source = $::puppet_agent::absolute_source
   } elsif ($::puppet_agent::is_pe and (!$::puppet_agent::use_alternate_sources)) {
     $pe_server_version = pe_build_version()
     if $::puppet_agent::alternate_pe_source {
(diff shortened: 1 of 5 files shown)


gimmyxd (approved):
Looks good!

Tested on a fresh node:
- installed latest nightly agent with the bolt task
- configured the module on PE and connected the agent
- managed agent version from PE using the PA module.

The commit check is failing because it accepts only `MODULES` as a project and we used to have a MODULES ticket for this.

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -37,8 +37,10 @@ def target_platform
                          '6.18.0'
                        when %r{osx-10.15}
                          '6.15.0'
+                       when %r{osx-11}
+                         '6.23.0'
                        else
-                         '6.17.0'
+                         '6.23.0'
```
gimmyxd, line 43: This is making the task spec fail on all the other OSes since we didn't release 6.23.0 yet

gimmyxd (changes requested):

manifests/osfamily/darwin.pp
```diff
@@ -1,19 +1,25 @@
 class puppet_agent::osfamily::darwin{
   assert_private()
 
+  if $::macosx_productversion_major =~ /^11\./ {
```
GabrielNagy, line 4: Assuming Apple will bump the major version on every macOS release, can we reverse the logic?
```suggestion
  if $::macosx_productversion_major =~ /^10\./ {
```

Also, `macosx_productversion_major` will be just `11` because of https://github.com/puppetlabs/facter/pull/2365. Since facter 3 doesn't have this change, maybe we can check based on `macosx_productversion` which should have the full version string everytime. Anyway if we reverse the logic it should be fine.

tasks/install_shell.sh
```diff
@@ -210,12 +210,17 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
     platform_version=`sw_vers | awk '/^ProductVersion:/ { print $2 }'`
 
     major_version=`echo $platform_version | cut -d. -f1,2`
+
+    # Starting with MacOS 11, the major version is the first number only
+    if echo "${major_version}" | grep -q '^11\.'; then major_version=11; fi
```
GabrielNagy, line 215: same concern as above

gimmyxd (approved):

GabrielNagy (approved):

Pigueiras:
This is going to be a random question, but where do I find big sur packages for puppet? If I check https://downloads.puppet.com/mac/puppet7/ https://downloads.puppet.com/mac/puppet6/ which are the urls you're pointing in your module, I can find only 10.15 dmgs.

GabrielNagy:
Hi @Pigueiras, macOS 11 support is set to go out in the next puppet-agent releases. In the meantime you can find nightly builds here: http://nightlies.puppet.com/downloads/mac/.

PR #565, opened 2021-06-04
(MODULES-11113) Allow present and latest as package version

relearnshuffle:
This pull requests intends to implement the package versions `present` and `latest` as valid package versions. According to #333, this was the case prior to commit c2206a7. The issue with #333 is that it is an outdated pull request that does not apply to the current state of  `main`. This, along with the fact that the author of #333 did not agree to the CLA in over 2 years now, is why I am raising this as a separate pull request.

Since there is no issues section in this project, I would also like to use this pull request as a discussion for this feature. The issue with the implementation as it is proposed in this pull request is that every puppet run will end with:
```
Notice: Stopping run after puppet-agent upgrade. Run puppet agent -t or apply your manifest again to finish the transaction.
```
when the version is set to `present` or `latest`. This is because the `needs_upgrade?` function in the `puppet_agent_end_run` type uses the `Puppet::Util::Package.versioncmp` function to compare the literal string `present` and `latest` with the actual version of the Puppet agent installed, e.g. `6.22.1`, which will never be equal.

I am not sure of a suitable fix for this and was hoping someone else could help.

I stumbled upon this issue when I realized the `pc_repo.repo` file was no longer being managed by this module after removing setting `puppet_agent::package_version` to `auto` which in itself led to an issue where the Puppet agent would also be downgraded (despite the server being the same version as the client) anytime the Puppet agent upgraded through the package manager. This is with `puppet_agent::collection` set to `puppet6` for both the main server and the agent, the `puppetserver` version being `6.15.3` and the `puppet-agent` version, prior to a catalogue pull, being `6.22.1` on both the main server and the agent.

### DIFF
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -62,7 +62,7 @@ describe 'puppet_agent' do
 
   context 'package version' do
     context 'valid' do
-      ['5.5.15-1.el7', '5.5.15.el7', '6.0.9.3.g886c5ab'].each do |version|
+      ['5.5.15-1.el7', '5.5.15.el7', '6.0.9.3.g886c5ab', 'present', 'latest'].each do |version|
 
         redhat_familly_supported_os.each do |os, facts|
           let(:facts) { global_facts(facts, os) }
@@ -88,6 +88,32 @@ describe 'puppet_agent' do
 
           context "on #{os}" do
             it { expect { catalogue }.to raise_error(%r{invalid version}) }
+          end
+        end
+      end
+    end
+
+    context 'latest' do
+      context 'on unsupported platform' do
+        on_supported_os.select { |platform, _| platform =~ /solaris|aix|windows|osx/ }.each do |os, facts|
+          let(:facts) { global_facts(facts, os) }
+          let(:params) { { package_version: 'latest' } }
+
+          context os do
+            it { is_expected.not_to compile }
+            it { expect { catalogue }.to raise_error(Puppet::Error, /Setting package_version to 'latest' is not supported/) }
+          end
+        end
+      end
+
+      context 'on supported platform' do
+        on_supported_os.reject { |platform, _| platform =~ /solaris|aix|windows|osx/ }.each do |os, facts|
+          let(:facts) { global_facts(facts, os) }
+          let(:params) { { package_version: 'latest' } }
+
+          context os do
+            it { is_expected.to compile.with_all_deps }
+            it { expect { catalogue }.not_to raise_error }
           end
         end
       end
--- a/Gemfile
+++ b/Gemfile
@@ -40,7 +40,7 @@ group :development do
   gem "rake", '~> 12',                                       require: false
   gem "parallel_tests", '>= 2.14.1', '< 2.14.3',             require: false
   gem "metadata-json-lint", '>= 2.0.2', '< 3.0.0',           require: false
-  gem "rspec-puppet-facts", '~> 1.10.0',                     require: false
+  gem "rspec-puppet-facts", '~> 2.0.1',                      require: false
   gem "rspec_junit_formatter", '~> 0.2',                     require: false
   gem "rubocop", '~> 0.49.0',                                require: false
   gem "rubocop-rspec", '~> 1.16.0',                          require: false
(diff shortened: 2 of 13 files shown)


GabrielNagy:
Thanks for your contribution @relearnshuffle!

> This is because the `needs_upgrade?` function in the `puppet_agent_end_run` type uses the `Puppet::Util::Package.versioncmp` function to compare the literal string `present` and `latest` with the actual version of the Puppet agent installed, e.g. `6.22.1`, which will never be equal.

After digging a bit into this, I found that we can search the catalog for the version we just upgraded to (in case `latest` is set in the manifest):

```
[73] pry(#<Puppet::Type::Puppet_agent_end_run::ProviderPuppet_agent_end_run>)> @resource.catalog.resource('package', 'puppet-agent').parameters[:ensure].latest                                                                    
=> "0:7.8.0.64.g6670bf40b-1.el8"
```

Since the version format is specific to yum-based packages, we have to normalize it so it looks like what `Facter.value(:aio_agent_version)` outputs. This regex appears to work for me: 
```
[26] pry(#<Puppet::Type::Puppet_agent_end_run::ProviderPuppet_agent_end_run>)> "0:7.8.0.64.g6670bf40b-1.el8".match(/^(?:[0-9]:)?(?<aio_version>\d+\.\d+\.\d+(?:\.\d+))?/)[:aio_version]
=> "7.8.0.64"
```

So the `needs_upgrade?` logic would become something like this:
```ruby
  def needs_upgrade?
    current_version = Facter.value('aio_agent_version')
    desired_version = @resource.name

    return false if desired_version == 'present'

    if desired_version == 'latest'
      latest_version = @resource.catalog.resource('package', 'puppet-agent').parameters[:ensure].latest
      desired_version = latest_version.match(/^(?:[0-9]:)?(?<aio_version>\d+\.\d+\.\d+(?:\.\d+))?/)[:aio_version]
    end

    Puppet::Util::Package.versioncmp(desired_version, current_version) != 0
  end
```

I assume `present` will be a noop everytime since the module requires puppet-agent to run, so the condition will always be satisfied.

Let me know if this works for you!

GabrielNagy:
@relearnshuffle I created the following JIRA ticket to track this issue: https://tickets.puppetlabs.com/browse/MODULES-11113

To get the PR commit check to pass, please prefix `(MODULES-11113)` to your commit message, something like: `(MODULES-11113) Allow present and latest as package version`

GabrielNagy:
I took it upon myself to work on this PR, since I believe this is an useful improvement.

This was also a good opportunity to add some test coverage, and bump rspec-puppet-facts which lowers CI duration with about ~5 minutes per run (due to the memoization of `on_supported_os`).

Setting `package_version` to `'latest'` should now be supported on all platforms for which we ship repos (EL, Fedora, Ubuntu, Debian, SLES), other platforms will raise an error.

luchihoratiu (approved):
LGTM! 👍 
Manually tested these changes on Ubuntu 20.04 and everything seems to be working as expected. Manifests used:
```puppet
class {'::puppet_agent':
  package_version => latest,
  collection => 'puppet6'
}
```
and
```puppet
class {'::puppet_agent':
  package_version => latest,
  apt_source => "http://nightlies.puppet.com/apt",
  collection => 'puppet6-nightly'
}
```
Output example:
```sh-session
➜ puppet apply manifest.pp
Notice: Compiled catalog for example.puppetlabs.net in environment production in 0.45 seconds
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Apt::Source[pc_repo]/Apt::Setting[list-pc_repo]/File[/etc/apt/sources.list.d/pc_repo.list]/content: content changed '{md5}88e1295cfdbda94bebe58c92215d90b7' to '{md5}935d4fe674b74276a929259e113d3e65'
Notice: /Stage[main]/Apt::Update/Exec[apt_update]: Triggered 'refresh' from 1 event
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Exec[pc_repo_force]/returns: forcing apt update for pc_repo puppet6-nightly
Notice: /Stage[main]/Puppet_agent::Osfamily::Debian/Exec[pc_repo_force]: Triggered 'refresh' from 2 events
Notice: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]/ensure: ensure changed '6.23.0-1bionic' to '6.24.0.28.g6b5947a6a-1bionic'
Notice: Stopping run after puppet-agent upgrade. Run puppet agent -t or apply your manifest again to finish the transaction.
Notice: Applied catalog in 12.81 seconds
➜ puppet apply manifest.pp
Notice: Compiled catalog for example.puppetlabs.net in environment production in 0.59 seconds
Notice: Applied catalog in 0.99 seconds
```

spec/classes/puppet_agent_spec.rb
```diff
@@ -92,6 +92,32 @@ def global_facts(facts, os)
         end
       end
     end
+
+    context 'latest' do
+      context 'valid' do
```
gimmyxd, line 97: I think this should be `context 'invalid'` as the catalog compilation raises error, the next one should be the `valid` context

PR #569, opened 2021-06-23
(MODULES-11077) Allow all settings to be managed

GabrielNagy:
Previously only a limited number of settings could be managed using the Puppet_agent::Config type. This commit allows all Puppet settings to be managed.


reidmv (approved):
LGTM 👍 

I'd idly considered trying to trim this list to settings which we think _should_ be managed on Puppet agents, but that's probably a lot of work without a lot of benefit. Just ensuring the setting is valid is probably best. This does that.

gimmyxd (approved):

PR #572, opened 2021-07-29
(MODULES-11060) Add Debian 11 testing support to install task

ciprianbadescu:
we add this earlier for testing and it will be updated when the official image is released

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -614,6 +614,7 @@ case $platform in
       "8") deb_codename="jessie";;
       "9") deb_codename="stretch";;
       "10") deb_codename="buster";;
+      "11"|"testing") deb_codename="bullseye";; # FIXME: testing to be removed when Debian is released and major_version is changed to 11
     esac
     filetype="deb"
     filename="${collection}-release-${deb_codename}.deb"


tasks/install_shell.sh
```diff
@@ -614,6 +614,7 @@ case $platform in
       "8") deb_codename="jessie";;
       "9") deb_codename="stretch";;
       "10") deb_codename="buster";;
+      "testing") deb_codename="bullseye";; # FIXME: to be fixed when Debian is released and major_version is changed to 11
```
david22swan, line 617: Unfortunately this won't work with the modules team setup as the matrix for it is automatically generated from the metadata file, with the resulting machine that is provisioned being clearly labelled as 11.

david22swan (changes requested):

tasks/install_shell.sh
ciprianbadescu: Debian 11 is planned to be released middle of August and then `major_version` will be updated to `11` instead of `testing`.
Do you have any proposal on how to fix this in the meantime?

david22swan: I've tested and for our setup 11 seems to work fine.

david22swan:
```
      "10") deb_codename="buster";;
      "11") deb_codename="bullseye";;
```
Adding the above to the code in my fixtures allowed me to install the agent on my docker machine and the acceptance tests ran fine.

david22swan: If testing is required for the setup that you are using, could we just include both? And then removed testing when it is no longer needed, it seems like the simplest option

GabrielNagy: We can do `11|testing)` until after the official release. The docker images return "11" indeed, but our vmpooler boxes return "testing".

ciprianbadescu: I updated PR to support both for now

david22swan (approved):
LGTM

Thanks for getting this done so fast :)

PR #581, opened 2021-09-19
Allow detection of non-AIO Puppet

smortex:
This PR add full support for FreeBSD.

If depends on this external PR:
* https://github.com/puppetlabs/puppetlabs-facts/pull/54

It is successfuly used for managing my configuration files on multiple operating systems:
https://github.com/smortex/dotfiles

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -133,6 +133,8 @@ fi
 # Track to handle puppet5 to puppet6
 if [ -f /opt/puppetlabs/puppet/VERSION ]; then
   installed_version=`cat /opt/puppetlabs/puppet/VERSION`
+elif which puppet >/dev/null 2>&1; then
+  installed_version=`puppet --version`
 else
   installed_version=uninstalled
 fi


smortex:
Please note that installing Puppet on FreeBSD is likely *not* to work on a fresh system because the script is not portable and depend on `bash` being installed.


I considered switching to bare bourne shell overkill for this PR but if it is something you want, I definitively can work on it :wink:.

smortex:
@GabrielNagy what I utterly need is only the first commit, which finds a non-AIO install of Puppet and use it if found, without the need to run as root user.  This is a requirement for things like `bolt apply ... -t remote-node`.

The second commit is just sugar for me, I do not really care about it, it just make things more consistent.

Would removing the second commit which manage package installation be OK for you?

GabrielNagy:
Sounds good! Do you have a quick repro case for me to test? I tried applying one of your profiles on a CentOS 8 machine with system puppet but I wasn't able to get past the `apply_helpers::custom_facts` task:

```sh-session
> bolt apply -e 'include profile::tmux' -t tasty-globulin.delivery.puppetlabs.net
Starting: install puppet and gather facts on tasty-globulin.delivery.puppetlabs.net
Plan aborted: run_task 'apply_helpers::custom_facts' failed on 1 target
[
  {
    "target": "tasty-globulin.delivery.puppetlabs.net",
    "action": "task",
    "object": "apply_helpers::custom_facts",
    "status": "failure",
    "value": {
      "_output": "",
      "_error": {
        "kind": "puppetlabs.tasks/task-error",
        "issue_code": "TASK_ERROR",
        "msg": "The task failed with exit code 126 and no stdout, but stderr contained:\nbash: /tmp/88693118-9e71-4413-b0c0-42dc559eadc1/custom_facts.rb: /opt/puppetlabs/puppet/bin/ruby: bad interpreter: No such file or directory\n\n\nThis might be caused by the default tmpdir being mounted using 'noexec'. See http://pup.pt/task-failure for details and workarounds.",
        "details": {
          "exit_code": 126
        }
      }
    }
  }
]
```

tasks/install_shell.sh
```diff
@@ -134,7 +134,12 @@ fi
 if [ -f /opt/puppetlabs/puppet/VERSION ]; then
   installed_version=`cat /opt/puppetlabs/puppet/VERSION`
 else
-  installed_version=uninstalled
+  if which puppet >/dev/null 2>&1
+  then
+    installed_version=`puppet --version`
+  else
+    installed_version=uninstalled
+  fi
```
GabrielNagy, line 142: We should be able to compact this to:

```sh
if [ -f /opt/puppetlabs/puppet/VERSION ]; then
  installed_version=`cat /opt/puppetlabs/puppet/VERSION`
elif which puppet >/dev/null 2>&1; then
  installed_version=`puppet --version`
else
  installed_version=uninstalled
fi
```

smortex:
I updated the PR to only detect and use non-AIO puppet when present and AIO package was not found.

---

Regarding the error you have when trying to use my dotfiles code, I use it the same way you do.

I often encounter similar errors with Bolt trying to use the path to a command on *my* machine on a remote one where the command is available somewhere else.  For example, on my FreeBSD machine and targeting a CentOS node, I got a similar error:

> The task failed with exit code 127 and no stdout, but stderr contained:\nzsh:1: /tmp/eeaa3ad6-b9be-458f-abe8-e7b8748fb1b0/custom_facts.rb: bad interpreter: /usr/local/bin/ruby: No such file or directory\n"

On the machine I run Bolt, ruby is `/usr/local/bin/ruby`, but on the remote node it's `/usr/bin/ruby` or better `/opt/puppetlabs/puppet/bin/ruby`…  So I had to edit inventory.yaml:

```yaml
targets:
  - uri: my-node-name
    config:
      ssh:
        interpreters:
          rb: "/opt/puppetlabs/puppet/bin/ruby"
```

My inventory.yaml is not under VCS and I think that I have different content for each machine where I clone my project because I have to adjust almost everything all the time.

If Bolt was honoring $PATH that could be ways better :grin:


But maybe your issue is more simple and is just that the AIO package is not installed on the remote node?  Since we do not `--run-as` an administrator, puppet must be installed on the remote node *before* managing the user configuration files (and thanks to this patch I can install it with `gem install --user-install puppet` on systems where I have not root access and still manage my configuration files).

tasks/install_shell.sh
smortex: Fixed!

GabrielNagy:
Works for me now, thanks @smortex, had to adjust the ruby path in inventory.yaml as you suggested. I see no reason not to merge this, then 😄.

GabrielNagy (approved):

donoghuc:
I know this is an older issue but when we try to take up this release (4.10.0) in bolt we encounter the case where there is only the ruby gem installed on a system and the agent package is not installed. I have some details in https://tickets.puppetlabs.com/browse/BOLT-1586

I really dont understand this https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/581#issuecomment-931766274 comment Bolt should not be overriding the shebang in custom facts https://github.com/puppetlabs/bolt/blob/f6c8cfe40a66d839ef52b56c93490b12c8096af9/libexec/custom_facts.rb#L1 on remote systems unless the interpreter is explicitly overridden in inventory. 

What this problem boils down to is that bolt's `apply_prep` (which uses the `puppet_library` hook to run the install task) relies on a system install (with a ruby interpreter at `/opt/puppetlabs/puppet/bin/ruby`). This commit breaks that assumption because systems where just the puppet gem are installed (using a different ruby interpreter) trigger the agent package to not be installed. The description of this task is https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/10e5a6b8552992e60a650d8b8c958e6be05e8ac1/tasks/install.json#L2 which to me indicates that packages should be installed even if there is just a puppet gem on the system. 

1. Can you provide more detail as to what the configuration looks like where the "wrong" ruby interpreter is used on the target
2. Was this change specifically for a use case around using bolt's apply/apply_prep? If so, can you configure your own `puppet_library` plugin hooks to do the proper detection of agents on your specific systems instead of changing the behavior here?

PR #582, opened 2021-09-24
(IAC-1751/IAC-1753) Add Rocky and AlmaLinux support to the install agent task

david22swan:


david22swan:
Have tested locally and everything works fine.

GabrielNagy (approved):

GabrielNagy:
Looks good, thanks @david22swan! We also have a ticket for adding this to CI, but no need to block this work.

PR #584, opened 2021-10-11
(maint). Add fact-limit configuration options to list.

Heidistein:
There are two fairly new options which go in puppet.conf.


Heidistein:
I added the other options as well. Perhaps I did it correctly.

GabrielNagy (approved):
Looks great! Thanks!

PR #585, opened 2021-11-05
Use Test-PuppetInstalledVersion instead of Test-PuppetInstalled

cgorshing:
I ran into an issue with Bolt running "puppet_agent::install" where we had uninstalled the Puppet Agent on a Win10 machine and Bolt would not install the agent again. The message was `Version parameter not defined and agent detected. Nothing to do.`

When the Puppet Agent is uninstalled, the registry keys (at least on the two Windows machines I've tested) are left in place. I believe the `Test-PuppetInstalled` method is a little lacking and should do further testing to check whether Puppet is installed or not. Similar to what the existing `Test-PuppetInstalledVersion` already does, which only has a boolean result.

I'm looking for suggestions related to this topic, because I can see a couple different options to tackle this problem.

I've also noticed there are a couple different places doing similar logic, reading the registry keys and the `VERSION` file. Another refactor might be to pull those pieces up to a "helper".

### DIFF
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -26,16 +26,6 @@ catch [System.Management.Automation.CommandNotFoundException] {
   }
 }
 
-function Test-PuppetInstalled {
-  $rootPath = 'HKLM:\SOFTWARE\Puppet Labs\Puppet'
-  try { 
-    if (Get-ItemProperty -Path $rootPath) { RETURN $true }
-  }
-  catch {
-    RETURN $false
-  }
-}
-
 function Test-PuppetInstalledVersion {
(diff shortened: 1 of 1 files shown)


joshcooper:
It makes sense to me to check for the version file and not just the presence of registry keys that may be left over after an uninstall. I'm closing and reopening this to trigger the GH checks.

joshcooper (approved):

joshcooper:
@donoghuc could you review?

joshcooper:
@beechtom since you worked on `Test-PuppetInstalled` in a91db25b394ae6fda29e59c447f3673b8ccd0adb and 6b8da89bea0327d11b6f860e9670b29097a54fbe, do you see any downside to removing that and calling `Test-PuppetInstalledVersion` instead?

PR #586, opened 2021-12-16
(MODULES-11214) Wrong url generated for darwin 11

BobosilaVictor:
Previously, any Darwin 11 machine that used a facter version older than
4.1.1, and tried to update the puppet-agent version using the
puppetlabs-puppet_agent module, would recieve a badly generated URL.

This happened because the `productversion_major` fact also included the
minor version.


ciprianbadescu:
@BobosilaVictor , could we also add osx platform to ad-hoc pipelines?

ciprianbadescu (approved):

PR #596, opened 2022-02-08
(MODULES-11276) Replace lsbdistcodename with os.disto.codename

jarretlavallee:
Prior to this PR, the debian package name used the lsbdistcodename
legacy fact to determine the package name. That legacy fact requires the
lsb-release package to determine, however in Puppet 7+, the
os.distro.codename fact can be determined without the lsb-release
package installed. This PR replaces the lsbdistcodename with
os.distro.codename for the Debian package name.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -70,7 +70,7 @@ class puppet_agent::install(
         if $package_version =~ /^latest$|^present$/ {
           $_package_version = $package_version
         } else {
-          $_package_version = "${package_version}-1${::lsbdistcodename}"
+          $_package_version = "${package_version}-1${facts['os']['distro']['codename']}"
         }
         $_provider = 'apt'
         $_source = undef
--- a/spec/classes/puppet_agent_osfamily_debian_spec.rb
+++ b/spec/classes/puppet_agent_osfamily_debian_spec.rb
@@ -7,6 +7,10 @@ describe 'puppet_agent' do
     lsbdistcodename: 'stretch',
     os: {
       'name'    => 'Debian',
+      'distro'  => {
+        'codename' => 'stretch',
+        'id'       => 'Debian'
+      },
       'release' => {
         'full'  => '9.0',
         'major' => '9',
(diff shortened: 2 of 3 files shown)


kenyon (approved):

manifests/install.pp
```diff
@@ -70,7 +70,7 @@
         if $package_version =~ /^latest$|^present$/ {
           $_package_version = $package_version
         } else {
-          $_package_version = "${package_version}-1${::lsbdistcodename}"
+          $_package_version = "${package_version}-1${facts['os']['distro']['codename']}"
```
joshcooper, line 73: Is the fact guaranteed to exist when upgrading from puppet-agent 5?

jarretlavallee: There are no guarantees for anything less than Puppet 7. Prior to that release, all of the LSB facts would be missing unless the `lsb-release` package was installed. If you were missing the `lsb-release` package, upgrading from Puppet 5 to 7 would still fail, but Puppet 7.a to 7.b would be successful after this change.

There are a couple of tickets open for this, but this was the most recent: https://tickets.puppetlabs.com/browse/FACT-3012

The `os.distro.codename` looks to have landed in [Facter 3.0+](https://github.com/puppetlabs/docs-archive/blob/main/facter/3.0/core_facts.md#os) 

Would it be worthwhile to add a note as they have in the [puppetlabs-apt](https://github.com/puppetlabs/puppetlabs-apt#module-description) module? Or I could fail/warn in a way that explains that the `lsb-release` package is missing.

> Note: Prior to Puppet 7, for this module to correctly autodetect which version of Debian/Ubuntu (or derivative) you're running, you need to make sure the lsb-release package is installed. With Puppet 7 the lsb-release package is not needed.

joshcooper:
```suggestion
  } elsif versioncmp($facts['facterversion'], '4.0.0') > 0
    $_package_version = "${package_version}-1${facts['os']['distro']['codename']}"
  } else
    $_package_version = "${package_version}-1${::lsbdistcodename}"
```

joshcooper: It'd be good not to break people upgrading from older versions, so maybe something like ^ ?

PR #599, opened 2022-03-03
(MODULES-11262) Update the version check for acceptance check

cthorn42:
When we test upgrades we should only check to look for the major version.
We use to test the latest known puppet-agent package was upgraded. Due
to a bug in our puppet-agent packaging there are times this test would
produce a failure, even though the agent was upgraded to the major version.

### DIFF
--- a/acceptance/tests/test_upgrade_puppet5_to_puppet6.rb
+++ b/acceptance/tests/test_upgrade_puppet5_to_puppet6.rb
@@ -48,7 +48,7 @@ node default {
     agents_only.each do |agent|
       on(agent, puppet('agent -t --debug'), acceptable_exit_codes: 2)
       wait_for_installation_pid(agent)
-      assert_agent_version_on(agent, latest_version.scan(%r{6\.\d*\.\d*\.\d*}).first)
+      assert(puppet_agent_version_on(agent) =~ %r{^6\.\d+\.\d+.*})
     end
   end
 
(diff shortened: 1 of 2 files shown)


joshcooper (commented):
Rubocop doesn't seem to like `/.../` I dunno if I agree with the cop, maybe disable `Style/RegexpLiteral` in .rubocop.yml? Or change the test?

Nit can we tighten up the regex? Right now when upgrading to `6.x` we'd match `5.16.x`. Similar for the upgrade to `7.x`. Can we anchor to the start of the string something like `^6\.[\d]+`?

cthorn42:
Doh, sorry good call out on the nit pick, for sure a mistake on my end. I updated the regex to also try to be more specific and also hopefully avoid the rubocop warning.

joshcooper:
looks like there's an extra backup file: `acceptance/tests/#test_upgrade_puppet6_to_puppet7.rb#`

acceptance/tests/test_upgrade_puppet5_to_puppet6.rb
```diff
@@ -48,7 +48,7 @@ class { puppet_agent:
     agents_only.each do |agent|
       on(agent, puppet('agent -t --debug'), acceptable_exit_codes: 2)
       wait_for_installation_pid(agent)
-      assert_agent_version_on(agent, latest_version.scan(%r{6\.\d*\.\d*\.\d*}).first)
+      assert(puppet_agent_version_on(agent) =~ %r{^6\.\d+.\d+.*$})
```
joshcooper, line 51: Nit we lost an escape on the dot, also the `.*$` is a little redundant since we don't care what comes after:

```suggestion
      assert(puppet_agent_version_on(agent) =~ %r{^6\.\d+\.\d+})
```

joshcooper (approved):

PR #602, opened 2022-05-02
(maint) Add macOS 12

joshcooper:
Adds macOS 12 as a supported platform.

Since macOS 12 has only been supported for one 6.27.1 release, don't try to test the upgrade from 6.27.1 -> latest and expect a new version to be installed.


joshcooper:
Blocked on https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/603

donoghuc (approved):

PR #606, opened 2022-05-13
(maint) Minor auto_release workflow fixes

mhashizume:
Updates commit summaries to include the `(packaging)` prefix
to appease the static code analysis PR check.

Updates link to to nightly tests.

### DIFF
--- a/.github/workflows/auto_release.yml
+++ b/.github/workflows/auto_release.yml
@@ -52,7 +52,7 @@ jobs:
         git config --local user.email "${{ github.repository_owner }}@users.noreply.github.com"
         git config --local user.name "GitHub Action"
         git add .
-        git commit -m "Release prep v${{ steps.gv.outputs.ver }}"
+        git commit -m "(packaging) Release prep v${{ steps.gv.outputs.ver }}"
 
     - name: Create Pull Request
       id: cpr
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_release.yml
```diff
@@ -52,22 +52,22 @@ jobs:
...
-        commit-message: "Release prep v${{ steps.gv.outputs.ver }}"
+        commit-message: "(packaging) Release prep v${{ steps.gv.outputs.ver }}"
         branch: "release-prep"
         delete-branch: true
         title: "Release prep v${{ steps.gv.outputs.ver }}"
         body: |
           Automated release-prep through [pdk-templates](https://github.com/puppetlabs/pdk-templates/blob/main/moduleroot/.github/workflows/auto_release.yml.erb) from commit ${{ github.sha }}. 
           Please verify before merging:
-          - [ ] last [nightly](https://github.com/${{ github.repository }}/actions/workflows/nightly.yml) run is green
+          - [ ] last [nightly](https://github.com/${{ github.repository }}/actions/workflows/daily_unit_tests_with_nightly_puppet_gem.yaml) run is green
```
joshcooper, line 70: I'm a little confused by the different workflows. Should we check that all of them are passing?

https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/.github/workflows/daily_unit_tests_with_nightly_puppet_gem.yaml
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/.github/workflows/unit_tests_with_nightly_puppet_gem.yaml
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/.github/workflows/unit_tests_with_released_puppet_gem.yaml

mhashizume: From what I can tell, the second two workflows (`unit_tests_with_nightly_puppet_gem` and `unit_tests_with_released_puppet_gem`) are kicked off when a PR is opened, so it's probably unnecessary to manually check them since they're running anyway and will block merge if they fail.

Given that, we probably don't need that manual check in the body of the PR at all since the scheduled, nightly unit tests and the unit tests that are kicked off when a PR is opened are identical except for their timing and a Slack notification.

mhashizume: These workflows are mostly taken from [the PDK templates repo](https://github.com/puppetlabs/pdk-templates/tree/main/moduleroot/.github/workflows), so I'm not sure if some of the decision-making baked into them necessarily make sense for us.

joshcooper: Ah yeah makes sense. I'm in favor of leaving the check you have in as we may release puppet-agent packages in between a puppet_agent PR and when the module is released.

PR #607, opened 2022-05-23
(maint) Adds cases for newly-supported OSes

mhashizume:
Fedora 34, Debian 11, and Red Hat 9 were all added after the 6
series had been released. As a result, we need special cases
for them in our install task test.


mhashizume:
I wasn't 100% sure on the syntax for the OS names used here, particularly with RHEL 9

cthorn42:
This looks good, what we are looking at is the `platform` that we set in beaker-hostgenerator: https://github.com/voxpupuli/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/data.rb
So for redhat9-64 the platform is set as `el-9-x86_64`.

PR #609, opened 2022-06-30
(FM-8943) Enable to install from nightly repo for Ubuntu 22.04 for task beaker tests

AriaXLi:
This PR is mostly to fix a typo. I have a very similar closed PR.

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -77,7 +77,7 @@ describe 'install task' do
 
     # Try to install an older puppet6 version
     results = if %r{ubuntu-22.04}.match?(target_platform)
-                run_task('puppet_agent::install', 'target', { 'collection' => 'puppet6_nightly',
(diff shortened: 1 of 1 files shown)


mhashizume:
Sorry, I missed an issue with your last PR. For the static code analysis check, it's only looking for `(FM)` right now, instead of the project name followed by the ticket number:

https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/21a148a9a13c0e89b43fc75594a5c2e32cb82ac4/rakelib/commits.rake#L10

In order for it to behave as expected, we need to do what was done for the MODULES project and add the regex `\d+`. `\d` will match any decimal digit, and the `+` will match that one or more times.

So line 10 should ultimately look like:

```
    if /^\((maint|packaging|doc|docs|fm-\d+|modules-\d+)\)|revert/i.match(commit_summary).nil?
```

Otherwise your PR looks good! I'll go ahead and approve and you can fix the static code analysis either in this PR or another.

PR #611, opened 2022-07-11
(MODULES-11334) Handle TLS 1.2 on older Windows systems

chelnak:
Most modern sites either require or are starting to require TLS 1.2. On older Windows systems the default security protocols for PowerShell 5 are SSL3 and TLS 1.0.

If using System.Net.WebClient, Invoke-WebRequest or Invoke-RestMethod to communicate with a site that has a higher TLS requirement requests will fail with the following error message:

`Could not create SSL/TLS secure channel.`

This PR adds a new function called `Set-Tls12` that will add TLS 1.2 to the list of active security protocols for the current session.

The above will only happen if the `Tls12` enum is not present in the current list.

This was causing failures on our nightly test runs for Windows Server 2016.

### DIFF
--- a/tasks/install_powershell.ps1
+++ b/tasks/install_powershell.ps1
@@ -105,8 +105,23 @@ $date_time_stamp = (Get-Date -format s) -replace ':', '-'
 $msi_dest = Join-Path ([System.IO.Path]::GetTempPath()) "puppet-agent-$arch.msi"
 $install_log = Join-Path ([System.IO.Path]::GetTempPath()) "$date_time_stamp-puppet-install.log"
 
+# Most modern sites either require or are starting to require TLS 1.2.
+# On older Windows systems the default security protocols for PowerShell 5 are SSL3 and TLS 1.0.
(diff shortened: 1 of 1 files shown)


tasks/install_powershell.ps1
```diff
@@ -105,8 +105,19 @@ $date_time_stamp = (Get-Date -format s) -replace ':', '-'
 $msi_dest = Join-Path ([System.IO.Path]::GetTempPath()) "puppet-agent-$arch.msi"
 $install_log = Join-Path ([System.IO.Path]::GetTempPath()) "$date_time_stamp-puppet-install.log"
 
+ # If TLS 1.2 is not configured for the current session add it to the list of defined Security Protocols
```
nicklewis, line 108: The commit message includes great context about why TLS 1.2 might not be configured and why it's necessary. Can you include a short version of that in this comment (ie. "old windows version, modern webserver")?

chelnak: Sure. I'll sort that out shortly.

chelnak: Updated!

PR #615, opened 2022-07-26
(FM-8969) Add support for macOS 12 ARM

joshcooper:
Emit task output when something goes wrong to ease debugging,

Distinguish between macOS 12 x86_64 and ARM in tests and install task.

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -32,6 +32,12 @@ describe 'install task' do
     hosts.first[:platform]
   end
 
+  def log_output_errors(result)
+    return if result['status'] == 'success'
+    out = result.dig('value', '_output') || 'Unknown result output'
+    puts logger.info(out)
+  end
+
   it 'works with version and install tasks' do
     puppet_6_version = case target_platform
                        when %r{debian-11}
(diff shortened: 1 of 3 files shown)


task_spec/spec/acceptance/init_spec.rb
```diff
@@ -32,7 +32,17 @@ def target_platform
     hosts.first[:platform]
   end
 
+  def log_result(result)
```
tvpartytonight, line 35: A little bit confusing when nothing happens for a successful `result`; maybe something like `log_output_errors` or something of that nature?

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -50,17 +60,20 @@ def target_platform
...
                        else
                          '6.17.0'
                        end
 
-    # platforms that only have nightly builds available
+    # platforms that only have nightly builds available. Once a platform
+    # is released, it should be removed from this list.
     case target_platform
-    when %r{ubuntu-22.04}
+    when %r{ubuntu-22.04}, %r{osx-12-arm}
```
tvpartytonight, line 76: Wow, I never knew you could just have comma separated values here. TIL.

tvpartytonight (approved):

task_spec/spec/acceptance/init_spec.rb
joshcooper: Updated

PR #624, opened 2022-09-29
(MODULES-11346) Update dependency for APT module

MartyEwings:
(MODULES-11346) Update dependency for APT module
Prior to this commit, the puppet_agent module requirements would put it at odds with other supported modules based on thee APT requirement of less than version 9.0

I also update the fixtures to that tests where carried out on the versions stated as supported in the metadata.yaml

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,13 +2,13 @@ fixtures:
   forge_modules:
     stdlib:
       repo: "puppetlabs/stdlib"
-      ref: "6.0.0"
+      ref: "8.4.0"
     inifile:
       repo: "puppetlabs/inifile"
-      ref: "2.4.0"
+      ref: "5.3.0"
     apt:
       repo: "puppetlabs/apt"
-      ref: "7.7.1"
+      ref: "9.0.0"
     translate:
       repo: "puppetlabs/translate"
       ref: "1.2.0"
@@ -16,6 +16,6 @@ fixtures:
       repo: "puppetlabs/yumrepo_core"
     facts:
       repo: "puppetlabs/facts"
-      ref: "0.5.1"
+      ref: "1.4.0"
   symlinks:
     puppet_agent: "#{source_dir}"
--- a/metadata.json
+++ b/metadata.json
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs-apt",
-      "version_requirement": ">= 7.7.1 < 9.0.0"
+      "version_requirement": ">= 7.7.1 < 10.0.0"
     },
     {
       "name": "puppetlabs-facts",
--- a/spec/default_facts.yml
+++ b/spec/default_facts.yml
@@ -1,0 +1,14 @@
+
+# Use default_module_facts.yml for module specific facts.
+#
+# Facts specified here will override the values provided by rspec-puppet-facts.
+---
+ipaddress: "172.16.254.254"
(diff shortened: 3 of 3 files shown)


.fixtures.yml
```diff
@@ -2,20 +2,20 @@ fixtures:
   forge_modules:
     stdlib:
       repo: "puppetlabs/stdlib"
-      ref: "6.0.0"
+      ref: "8.4.0"
```
bastelfreak, line 5: is there a reason for pinning? at vox pupuli we run rspec-puppet against the master branch of the dependencys. that enables us to catch breaking changes in deps before they are released. also we dont need to upgrade the fixtures file on a regular basis. for beaker tests, we run against the latest forge releases that the metadata.json allows.

MartyEwings: @bastelfreak I would agree but this isnt my project and i was copying the "status quo"

MartyEwings: @joshcooper what do you think? can we change all of these to main? it is best practice in most of the modules

joshcooper: Assuming this file is only used in CI, I'd be fine changing to `main`

metadata.json
```diff
@@ -18,7 +18,7 @@
     },
     {
       "name": "puppetlabs-apt",
-      "version_requirement": ">= 7.0.1 < 9.0.0"
+      "version_requirement": ">= 7.0.1 < 10.0.0"
```
joshcooper, line 21: From what I can tell the apt 9.0 release introduced a breaking change to the `apt::ppa` class: https://github.com/puppetlabs/puppetlabs-apt/commit/83082c088e6a6703a08071fbd5af0297beae62b9?w=1. However, this module doesn't use that class so this bump should be ok. That said, it would be good to test that a 5.5.x agent can upgrade to 7.x with the updated apt module.

MartyEwings: I'll give that a whirl in the lab tomorrow

MartyEwings: @joshcooper slight fly  in the ointment, i cant produce a puppet compatible CA in my puppet 7 labs in order to test this senario

vegaaz:
Are there any plans to merge this soon?

bastelfreak:
@MartyEwings any chance you can pick this up again? This would allow some customers to upgrade their modules.

MartyEwings:
@bastelfreak iv tested everything i can i think its a case of merging, @joshcooper you happy with this?

.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.3
```
joshcooper, line 1: I don't think we want this. I assume it's from rbenv or rvm?

AriaXLi:
Hello! Whenever you have time, please rebase your changes on top of main. Thank you!

MartyEwings:
@AriaXLi done

metadata.json.orig
```diff
@@ -0,0 +1,85 @@
...
+      "name": "puppetlabs-stdlib",
+      "version_requirement": ">= 5.1.0 < 9.0.0"
+    },
+    {
+      "name": "puppetlabs-inifile",
+      "version_requirement": ">= 2.4.0 < 6.0.0"
+    },
+    {
+      "name": "puppetlabs-apt",
+<<<<<<< HEAD
```
mhashizume, line 21: Looks like this file is an artifact from a rebase?

mhashizume (approved):

PR #627, opened 2022-11-23
(MODULES-11361) Remove legacy facts

joshcooper:
This isn't quite ready because we need to use legacy facts when older agents upgrade. Also Facter 3 doesn't support os.distro.codename if lsb_release is not installed (see MODULES-11348)


mhashizume:
Superceded by https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/636 and https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/637

PR #628, opened 2022-11-30
(maint) replace legacy validate function with datatype

bastelfreak:
This replaces the validate_absolute_path() function call with a datatype from puppetlabs/stdlib. The type is available since the 4.13.1 release of stdlib (and this module already depends on 5.1 and newer).

Some background: I'm currently debugging a slow pe-puppetserver metrics API. The puppet-profiler is enabled and provides metrics for all called functions. In our environment the puppet_agent module is the only one still using this legacy function. By replacing it with a datatype, it follows current best practice guides and also speeds up the API because it has to care about less metrics (in my specific case at least).

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -117,7 +117,7 @@ class puppet_agent (
   $aix_source              = 'puppet:///pe_packages',
   $use_alternate_sources   = false,
   $alternate_pe_source     = undef,
-  $install_dir             = undef,
+  Optional[Stdlib::Absolutepath] $install_dir = undef,
   $disable_proxy           = false,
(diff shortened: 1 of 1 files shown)


kenyon (approved):

bastelfreak:
Hey team @phoenix, can you please take a look? :)

mhashizume:
Thank you for this PR @bastelfreak , could you rebase and I can approve/merge?

bastelfreak:
@mhashizume done. Can you please also take a look at #632 ?

mhashizume (approved):

PR #631, opened 2022-12-16
allow puppet agent installation on fedora 33

ndelic0:


mhashizume:
Hi @ndelic0 

Thank you for your contribution, but Fedora 33 hit end-of-life on November 30, 2021 so we will not support it here.

Closing PR

PR #634, opened 2023-02-22
(MODULES-11348) Replace lsbdistcodename with os.distro.codename

AriaXLi:
This commit replaces references to `lsbdistcodename` with `os.distro.codename`.
`lsbdistcodename` is a legacy fact that is used to determine the package name but
it requires the lsb-release package. Additionally, for Puppet 8, we are moving
towards dropping legacy facts. Also, since Facter 3+ provides the structured
fact, `os.distro.codename`, users can safely upgrade from Puppet 6/7 to 8 or from
Puppet 8.x to 8.y when using `os.distro.codename`. Due to these reasons, 
`os.distro.codename`, will be used instead of `lsbdistcodename`.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -70,7 +70,7 @@ class puppet_agent::install(
         if $package_version =~ /^latest$|^present$/ {
           $_package_version = $package_version
         } else {
-          $_package_version = "${package_version}-1${::lsbdistcodename}"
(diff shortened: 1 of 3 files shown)


AriaXLi:
This PR is based off of #596

joshcooper (approved):

PR #643, opened 2023-03-17
(PA-5302) Remove puppet5 and MCollective

mhashizume:
Puppet 5 has long been end-of-life and we no longer support it in this module. This commit removes support and reference for Puppet 5 and MCollective, a service that was last provided in Puppet 5.


joshcooper (approved):
Yay spring cleaning!

PR #655, opened 2023-05-22
(PA-5287) Remove MAC OS 10.15 support

shubhamshinde360:


tasks/install_shell.sh
```diff
@@ -237,7 +237,6 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
       "10.12") platform_version="10.12";;
       "10.13") platform_version="10.13";;
       "10.14") platform_version="10.14";;
```
mhashizume, line 239: Could you remove these older versions as well the way we did in install-puppet?

shubhamshinde360: Thanks @mhashizume,

Removed all the 10.x lines and removed the if condition since it will always be true now after the removal.

joshcooper (approved):

PR #657, opened 2023-07-06
Implement stdlib 9 compatibility

bastelfreak:
stdlib 9 dropped the `is_function_available()` function. I reworked the code to check if `pe_anchor` is defined. The code now works with Puppet core functions and is compatible with Puppet 7/8, probably 6, and does not rely on stdlib.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -43,8 +43,9 @@ class puppet_agent::params {
   # The `is_pe` fact currently works by echoing out the puppet version
   # and greping for "puppet enterprise". With Puppet 4 and PE 2015.2, there
   # is no longer a "PE Puppet", and so that fact will no longer work.
-  # Instead check for the `is_pe` fact or if a PE provided function is available
-  $_is_pe = (getvar('::is_pe') or is_function_available('pe_compiling_server_version'))
+  # Instead check for the `is_pe` fact or if the `pe_anchor` resource type
+  # provided by the puppet_enterprise module exist.
+  $_is_pe = (getvar('::is_pe') or defined('pe_anchor'))
   if $_is_pe {
     # Calculate the default collection
     $_pe_version = pe_build_version()
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -69,17 +69,9 @@ class puppet_agent::params {
     else {
       $collection = 'puppet8'
     }
-    # The aio puppet-agent version currently installed on the compiling master
-    # (only used in PE)
-    if is_function_available('pe_compiling_server_aio_build') {
-      $master_agent_version = pe_compiling_server_aio_build()
-    } else {
-      $master_agent_version = undef
-    }
   } else {
     $_pe_version = undef
     $pe_repo_dir = undef
-    $master_agent_version = undef
     $collection = 'PC1'
   }
 }
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -237,7 +237,7 @@ describe 'puppet_agent' do
         before(:each) do
           Puppet::Parser::Functions.newfunction(:pe_build_version, type: :rvalue) { |_args| '2000.0.0' }
           Puppet::Parser::Functions.newfunction(:pe_compiling_server_aio_build, type: :rvalue) { |_args| '1.10.100' }
-          Puppet::Parser::Functions.newfunction(:pe_compiling_server_version, type: :rvalue) { |_args| '2.20.200' }
+          Puppet::Parser::Functions.newfunction(:defined, type: :rvalue) { |_args| true }
         end
 
         context 'package_version is initialized automatically' do
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -201,14 +201,6 @@ describe 'puppet_agent' do
           end
         end
 
-        context 'package_version is undef if pe_compiling_server_aio_build is not defined' do
-          let(:facts) do
-            global_facts(facts, os).merge(is_pe: true)
-          end
-
-          it { is_expected.to contain_class('puppet_agent').with_package_version(nil) }
-        end
-
         context 'package_version is same as master when set to auto' do
           let(:params) { { package_version: 'auto' } }
           let(:node_params) { { serverversion: '7.6.5' } }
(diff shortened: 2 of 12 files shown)


manifests/params.pp
```diff
@@ -40,12 +40,7 @@
   $ssldir = "${confdir}/ssl"
   $config = "${confdir}/puppet.conf"
 
-  # The `is_pe` fact currently works by echoing out the puppet version
-  # and greping for "puppet enterprise". With Puppet 4 and PE 2015.2, there
-  # is no longer a "PE Puppet", and so that fact will no longer work.
-  # Instead check for the `is_pe` fact or if a PE provided function is available
-  $_is_pe = (getvar('::is_pe') or is_function_available('pe_compiling_server_version'))
-  if $_is_pe {
+  if $facts['is_pe'] {
```
bastelfreak, line 43: ah derp, the fact is only available on PE primaries/compilers.

manifests/params.pp
```diff
@@ -40,11 +40,9 @@
...
   $config = "${confdir}/puppet.conf"
 
-  # The `is_pe` fact currently works by echoing out the puppet version
-  # and greping for "puppet enterprise". With Puppet 4 and PE 2015.2, there
-  # is no longer a "PE Puppet", and so that fact will no longer work.
-  # Instead check for the `is_pe` fact or if a PE provided function is available
-  $_is_pe = (getvar('::is_pe') or is_function_available('pe_compiling_server_version'))
+  # To determine if this is a PE installation or not, we check if the `pe_anchor` resource type exists
+  # it's provided by the puppet_enterprise module
+  $_is_pe = defined('pe_anchor')
```
smortex, line 45: It looks like the test suite rely on the `is_pe` variable and removing it from the condition break the test suite.  This PR being about updating dependencies, I think we can ignore this compatibility thing for now (assuming we trust the comments), and only focus on supporting recent stdlib?

https://github.com/puppetlabs/puppetlabs-stdlib/pull/1386 might be useful to keep the current behavior (see bellow).

spec/classes/puppet_agent_spec.rb
```diff
@@ -237,7 +237,7 @@ def global_facts(facts, os)
         before(:each) do
           Puppet::Parser::Functions.newfunction(:pe_build_version, type: :rvalue) { |_args| '2000.0.0' }
           Puppet::Parser::Functions.newfunction(:pe_compiling_server_aio_build, type: :rvalue) { |_args| '1.10.100' }
-          Puppet::Parser::Functions.newfunction(:pe_compiling_server_version, type: :rvalue) { |_args| '2.20.200' }
+          Puppet::Parser::Functions.newfunction(:defined, type: :rvalue) { |_args| true }
```
smortex, line 240: Mocking `defined` to return true might break things. According to the tests I did with the module it *seems* to be fine, but it feel fragile :fearful:.

smortex (commented):
Some in-line comments. I will push some local changes to your branch to see if it helps.

manifests/params.pp
```diff
@@ -44,7 +44,7 @@
   # and greping for "puppet enterprise". With Puppet 4 and PE 2015.2, there
   # is no longer a "PE Puppet", and so that fact will no longer work.
   # Instead check for the `is_pe` fact or if a PE provided function is available
-  $_is_pe = (getvar('::is_pe') or is_function_available('pe_compiling_server_version'))
+  $_is_pe = (getvar('::is_pe') or stdlib::has_function('pe_compiling_server_version'))
```
bastelfreak, line 47: I'm fine with this, but we need to keep in mind that it now requires stdlib 9.3. with my `defined()` it works on all stdlib versions

smortex: Yes, I am not happy with this neither… :sob:

I am currently fighting with bolt to use `apply` with modules that support stdlib ">= 9" but Bolt's `apply` rely on this module which currently only support stdlib "< 9".

bastelfreak: FYI I rolled out my code with `defined()` at a few customers and that works fine as expected

smortex: It is fragile but avoids the need for bumping the minimum version of stdlib.  I have no PE anywhere around, so I am not really concerned about this fragility, if it is okay, then it is fine!  I'll rework the commits I added to take this into account. If it fix CI and you can test this in the field to validate the change, I think we will be good to go.

smortex: > I'll rework the commits I added to take this into account. If it fix CI and you can test this in the field to validate the change, I think we will be good to go.

Code pushed. CI failures are related to my commit messages on the commits that need merging, so if you can confirm it fix your issue on PE, you can merge the commits (fixup/squash) and I think CI will be all green :grin:

smortex:
I have squashed my changes in your commits, and CI seems happy with it.

@bastelfreak  had you the opportunity to check that the behavior was not broken by my change in the real world with PE?  Is this PR ready for review :grin: :wink: :wink:?

bastelfreak:
@smortex yes it's working fine :)

smortex (approved):
This looks reasonable to me: the minor "issues" are not blockers and given the bad situation of the module (prevent usage of latest stdlib), this is a move in the right direction.  As it is now, it is still compatible with previous versions of stdlib so AFAIAC this is :100:!

:shipit:

joshcooper (commented):
Thank you @bastelfreak! We're heads down working on upcoming releases and can take a look at this next week. In the meantime could you update the summary to describe why this change is needed? I think I understand, but I admit I haven't been paying complete attention to the strict & stdlib deprecated functions issues.

bastelfreak:
@joshcooper I updated the git commit with the description and added it to the PR title as well.

I want to point out that puppetlabs/stdlib 9 was released almost 2 months ago. PE customers were not able to upgrade because this module isn't compatible. The CAT team worked hard to upgrade all of *their* modules. I still don't understand why the puppetdb and puppet_agent module are not owned by them / the CAT team isn't listed as code owners. From my point of view (also also from the PE customers I represent) this doesn't make sense. I pointed it out multiple times in the past years and never got a proper explanation for this. Also this wasn't changed. There was zero activity from the current owners about making puppet_agent compatible with stdlib 9. It took quite a long time to get a response here. I fail to understand why some modules are not maintained at all / maintained by other teams instead of the module with the main focus on maintaining modules. I would love to see a migration to the CAT team to ship enhancements faster to customers and to have a single source of configuration/guidelines/team for modules.

spec/classes/puppet_agent_spec.rb
tvpartytonight: I'm not sure I understand why this is getting mocked here; @bastelfreak can you shed some light?

joshcooper:
Hi @bastelfreak, I understand your frustration, the current approach is not working for anyone. I've been advocating for moving these and `*_core` modules to the ~Modules~, ~IAC~ and CAT teams for literally years. But as usual, there are too many things to do and not enough people. I am hopeful this will change in the near future though. I'll have a chat with folks on my end.

tvpartytonight (approved):

bastelfreak:
@tvpartytonight thanks for merging <3

PR #658, opened 2023-07-13
Update solaris to add pkg publisher

rstuart-indue:
The Solaris class only set up the repo, it didn't configure the pkg tool to use it. This adds the extra command to configure the "client" side of this to allow the pkg tool to see packages in the repo previously configured.

### DIFF
--- a/manifests/osfamily/solaris.pp
+++ b/manifests/osfamily/solaris.pp
@@ -107,6 +107,12 @@ class puppet_agent::osfamily::solaris {
           logoutput   => 'on_failure',
           refreshonly => true,
         }
+        ~> exec { 'puppet_agent add publisher to pkg':
+          command     => "pkg set-publisher -e -g ${pkgrepo_dir} ${publisher}",
+          path        => '/bin:/usr/bin:/sbin:/usr/sbin',
+          logoutput   => 'on_failure',
+          refreshonly => true,
+        }
         # Make sure the pkg publishers are all available.  Broken
         # publisher entries will stop the installation process.
         # This must happen before removing any packages.


joshcooper (commented):
Thanks @rstuart-indue! What was the error you were seeing prior to this PR? I'm also wondering if this ever worked in certain circumstances?

manifests/osfamily/solaris.pp
```diff
@@ -107,6 +107,12 @@
           logoutput   => 'on_failure',
           refreshonly => true,
         }
+        ~> exec { 'puppet_agent add publisher to pkg':
+          command     => "pkg set-publisher -e -g ${pkgrepo_dir} ${publisher}",
```
joshcooper, line 111: If the `publisher` already exists (as described in https://tickets.puppetlabs.com/browse/PA-5215) will this command still work?

joshcooper: ping @rstuart-indue

rstuart-indue: Sorry @joshcooper  it looks like I missed the previous notifications. 

I don't seem to be able to find the linked ticket above. The pkg man page says it will update or add. here is a system and executing this command when the publisher already exists:
`
[16:32:27] root@host
/root# pkg publisher
PUBLISHER                   TYPE     STATUS P LOCATION
solaris                     origin   online F http://172.20.98.16:8003/IPS/
puppetlabs.com              origin   online F file:///etc/puppetlabs/installer/solaris.repo/
[16:32:33] root@host
/root# pkg publisher puppetlabs.com

            Publisher: puppetlabs.com
                Alias:
           Origin URI: file:///etc/puppetlabs/installer/solaris.repo/
        Origin Status: Online
              SSL Key: None
             SSL Cert: None
          Client UUID: 6cfd453a-2141-11ee-9502-43f3b6da7299
      Catalog Updated: Thu Nov  9 03:44:37 2023
    Publisher enabled: Yes
[16:32:36] root@host
/root# pkg set-publisher -e -g /etc/puppetlabs/installer/solaris.repo/ puppetlabs.com
[16:32:44] root@host
/root# pkg publisher puppetlabs.com
            Publisher: puppetlabs.com
                Alias:
           Origin URI: file:///etc/puppetlabs/installer/solaris.repo/
        Origin Status: Online
              SSL Key: None
             SSL Cert: None
          Client UUID: 6cfd453a-2141-11ee-9502-43f3b6da7299
      Catalog Updated: Thu Nov  9 03:44:37 2023
    Publisher enabled: Yes
[16:32:51] root@host
`

This additional code is to have Solaris 11 behave the same was as redhat and configure the repo as appropriate. ref https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/manifests/osfamily/redhat.pp#L162 and this block mimics the yumrepo declaration of line 167.

joshcooper: For posterity, about PA-5215, the issue was

> When using puppet_agent to upgrade from PE 2016.4.6 to 2017.3.1 we were seeing issues when trying to retrieve the last version of puppet-agent package:

```
Error: Execution of '/usr/bin/pkg list -Hvfa puppet-agent@5.3.2' returned 1: pkg list: no packages matching 'puppet-agent@5.3.2' known
Error: /Stage[main]/Puppet_agent::Install/Package[puppet-agent]/ensure: change from 1.7.1,5.11-1:20161012T034616Z to 5.3.2 failed: Execution of '/usr/bin/pkg list -Hvfa puppet-agent@5.3.2' returned 1: pkg list: no packages matching 'puppet-agent@5.3.2' known
```

> The curl installer was working as expected.
> After some troubleshooting we discovered that customer had a custom publisher called puppetlabs.com that did not contained the puppet-agent version 5.3.2, making the pkg list command fail, and the puppet_agent module fail.
> Notice that this used to happen in the frictionless installer, but it was fixed.

@rstuart-indue this change makes sense to me, could you add a test?

PR #659, opened 2023-08-14
(PA-5309) Add macOS 13 (ARM) to the puppet_agent module task

skyamgarp:


joshcooper (approved):

joshcooper:
Ah @skyamgarp for next time, could you include a brief description in your commit message about why these changes were necessary? I understand this particular issue, but it's helpful to the next person who may not understand the significance of why we need to use nightly repos for new agent platforms.

PR #660, opened 2023-08-17
Support for debian absolute_path to be dynamic

pfo-netic:
For some reason absolute_path for Debian systems is statically set to 'present' in install.pp.
This seems to be because dpkg can't accept $package_version if it is set to a version or 'auto'.

This makes you unable to use a puppet installer locally placed on the system to upgrade your puppet agent on Debian.

Why not set $_package_version = $package_version and then just make sure that it is either 'present' or 'latest' and otherwise fail?

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -61,7 +61,10 @@ class puppet_agent::install (
       $_install_options = $install_options
       if $::puppet_agent::absolute_source {
         # absolute_source means we use dpkg on debian based platforms
-        $_package_version = 'present'
+        if ($package_version != 'present' and $package_version != 'latest') {
+          fail('When using $absolute_source on Debian, $package_version must be set to "present" or "latest"')
+        }
+        $_package_version = $package_version
         $_provider = 'dpkg'
         # The source package should have been downloaded by puppet_agent::prepare::package to the local_packages_dir
         $_source = "${::puppet_agent::params::local_packages_dir}/${::puppet_agent::prepare::package::package_file_name}"


joshcooper:
Closed and reopened to kick GH checks

joshcooper:
Thoughts @kenyon @bastelfreak ?

kenyon (approved):
> Thoughts @kenyon @bastelfreak ?

I don't use `absolute_source`, but I think the logic in this change makes sense.

I guess `absolute_path` in the PR title and description is a typo for `absolute_source`. Not sure what the "be dynamic" means. I'd title it "Fail with helpful error message on Debian if incorrect `package_version` is used with `absolute_source`".

kenyon:
> For some reason absolute_path for Debian systems is statically set to 'present' in install.pp.

I think this should read "For some reason `package_version` for Debian systems is statically set to `present` in `install.pp` when using `absolute_source`.

manifests/install.pp
```diff
@@ -61,7 +61,10 @@
       $_install_options = $install_options
       if $::puppet_agent::absolute_source {
         # absolute_source means we use dpkg on debian based platforms
-        $_package_version = 'present'
+        if ($package_version != 'present' and $package_version != 'latest') {
```
bastelfreak, line 64: isn't `installed` also allowed?

joshcooper:
@pfo-netic could you take a look at @bastelfreak and @kenyon's comments?

pfo-netic:
Hi,
It's such a long time since I was working with this, but it seems like you are getting the gist of what I was trying to say.
If i remember correctly I just wanted this part: 
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/manifests/install.pp#L68-L73
To work like this part:
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/manifests/install.pp#L86-L91

Which from my point of view it also should, which made me think it was an oversight in the first place.

PR #664, opened 2023-08-29
(maint) Delete duplicate test

joshcooper:
This code was last updated in Oct 2020 and all this time we've been testing "upgrades" from 7.x to 5.5.21. This only recently started to fail when stdlib v9 was released as the root_home fact uses Ruby syntax that's not valid in Puppet 5/Ruby 2.1.9.

It's not practical to update this test after each release. And we already test idempotency in the acceptance/tests/*, so delete this test.


joshcooper:
The `root_home` custom fact is defined here and uses syntax that's only supported in Ruby 2.3 and up:

https://github.com/puppetlabs/puppetlabs-stdlib/blob/abac8aeaa5d967691b25480a45394c2b38d447ce/lib/facter/root_home.rb#L6

tvpartytonight:
I am still seeing failures, though less, when I run the changes from this branch through `bundle exec rake beaker`. I think we should just wait and fix them all so we can get a clean run through CI.

bastelfreak (approved):

tvpartytonight (approved):

spec/acceptance/class_spec.rb
```diff
@@ -3,24 +3,12 @@
...
-    after(:all) { teardown_puppet_on default }
-
-    it 'works idempotently with no errors' do
-      pp = <<-EOS
-      class { 'puppet_agent': package_version => '5.5.21', collection => 'puppet5' }
-      EOS
-
+    before(:all) do
+      setup_puppet_on default
+      pp = "class { 'puppet_agent': package_version => 'auto'}"
```
joshcooper, line 8: `auto` causes the module to lookup the [`VERSION` file from the master](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/04bee65e4ea936faa2ed91872b24d26cbd44009e/manifests/init.pp#L164) and use that version when managing the package on the agent. But when running `puppet apply` it doesn't know to look in the Windows install directory. It's possible to get that using the `env_windows_installdir` fact, but wondering if maybe we can use `present` instead since we know the package is installed and we just want the service to be started?

joshcooper: Sadly `present` doesn't work:

```
Error: /Stage[main]/Puppet_agent::Prepare::Package/File[C:\ProgramData\Puppetlabs\packages\puppet-agent-present-x64.msi]: Could not evaluate: Could not retrieve information from environment production source(s) https://downloads.puppet.com/windows/puppet-agent-present-x64.msi
```

And if I lookup the version from the local `VERSION` file, then the module tries to prepare and downloads the MSI from the wrong URL (packages have been put in versioned directories for awhile now)

```
Could not retrieve information from environment production source(s) https://downloads.puppet.com/windows/puppet-agent-7.23.0-x64.msi
```

joshcooper: oh the [last passing run](https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/Acceptance%20Suites/view/CI%20Goalie/job/forge-module_puppetlabs-puppet-agent-module_intn-sys_smoke-main/301/PLATFORM=windows2012r2-64default.a-redhat7-64mdca,WORKER_LABEL=k8s-beaker/consoleFull) passed the `collection` explicitly and that seems to do the trick:

```
$ cat manifest.pp 
class { 'puppet_agent':
  package_version => '7.23.0',
  collection => 'puppet7',
}

$ cmd /c puppet apply manifest.pp
Notice: Compiled catalog for shy-mechanist.delivery.puppetlabs.net in environment production in 0.17 seconds
Warning: Private key for 'shy-mechanist.delivery.puppetlabs.net' does not exist
Warning: Client certificate for 'shy-mechanist.delivery.puppetlabs.net' does not exist
Notice: /Stage[main]/Puppet_agent::Prepare::Package/File[C:\ProgramData\Puppetlabs\packages\puppet-agent-7.23.0-x64.msi]/ensure: defined content as '{mtime}2023-02-08 18:35:06 UTC'
Notice: Applied catalog in 2.78 seconds
```

PR #665, opened 2023-08-31
(maint) Make spec tests pass on windows

tvpartytonight:
Windows requires the current puppet collection to be specified in the puppet_agent class, so that is refactored to be globally specified in the `spec_helper_acceptance`. Additionally, logic was added to specifiy default install paths for the VERSION file for both linux and windows.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -126,6 +126,7 @@ class puppet_agent (
   $wait_for_pxp_agent_exit = undef,
   $wait_for_puppet_run     = undef,
   Array[Puppet_agent::Config] $config = [],
+  $version_file_path       = $facts['os']['family'] ? { 'windows' => "${facts['env_windows_installdir']}\\VERSION", default => '/opt/puppetlabs/puppet/VERSION' }
 ) inherits ::puppet_agent::params {
   # The configure class uses $puppet_agent::config to manage settings in
   # puppet.conf, and will always be present. It does not require management of
@@ -161,7 +162,7 @@ class puppet_agent (
     # The AIO package version and Puppet version can, on rare occasion, diverge.
     # This logic checks for the AIO version of the server, since that's what the package manager cares about.
     if $package_version == 'auto' {
-      $master_or_package_version = chomp(file('/opt/puppetlabs/puppet/VERSION'))
+      $master_or_package_version = chomp(file($version_file_path))
     } else {
       $master_or_package_version = $package_version
     }
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -5,7 +5,7 @@ describe 'puppet_agent class' do
   context 'default parameters in apply' do
     before(:all) do
       setup_puppet_on default
-      pp = "class { 'puppet_agent': package_version => 'auto'}"
+      pp = "class { 'puppet_agent': package_version => 'auto', collection => #{PUPPET_COLLECTION}}"
       apply_manifest(pp, catch_failures: true)
     end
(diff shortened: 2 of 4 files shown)


kenyon:
I guess this explains why `auto` doesn't seem to be working on Windows for us, ending up with various agent versions on Windows.

tvpartytonight:
@joshcooper before I dig into the spec failures here, do you think this approach is correct? The `bundle exec rake beaker` passes on windows with these changes.

manifests/init.pp
```diff
@@ -126,6 +126,7 @@
   $wait_for_pxp_agent_exit = undef,
   $wait_for_puppet_run     = undef,
   Array[Puppet_agent::Config] $config = [],
+  $version_file_path       = $facts['os']['family'] ? { 'windows' => 'C:\Program Files\Puppet Labs\Puppet\VERSION', default => '/opt/puppetlabs/puppet/VERSION' }
```
joshcooper, line 129: On Windows the install dir is user-configurable/relocatable, but there's a core fact you can use https://www.puppet.com/docs/puppet/7/core_facts.html#env-windows-installdir

```suggestion
  $version_file_path       = $facts['os']['family'] ? { 'windows' => "${facts['env_windows_installdir']}\\VERSION", default => '/opt/puppetlabs/puppet/VERSION' }
```
One tricky thing, you'll need to double escape the backslash in the double quoted string (since we're interpolating). I don't know if that syntax works exactly, but something like it should.

joshcooper:
Yeah makes sense. I think the failures are because we have to stub out the VERSION file when testing the windows osfamily class.

manifests/init.pp
kenyon: You can also use forward slashes on Windows.

tvpartytonight:
I added a mock similar to the one already existing for `/opt/puppetlabs/puppet/VERSION`, but it doesn't fix the unit test failure.

joshcooper:
Maybe one option is to not try to fix auto on windows and puppet apply. Instead just modify the class_spec to specify the package_version and collection when applying the class.

tvpartytonight:
@joshcooper I was able to keep using the `auto` value, I realized I just wasn't mocking enough previously.

joshcooper (approved):
ah nice!

kenyon:
I'm testing puppet_agent 4.15.0. I think the change in this PR is causing catalog compilation to fail on Windows agents (PE 2019.8.11, Puppet 6.27.0). It appears to be trying to get the content of `"${facts['env_windows_installdir']}\\VERSION"` from the primary server, which of course is not running Windows.

joshcooper:
> It appears to be trying to get the content of "${facts['env_windows_installdir']}\\VERSION" from the primary server, which of course is not running Windows.

Gah... I forgot the module is using the `file` function... Right so when running `puppet apply` it will retrieve the VERSION file from the currently running agent. But when puppetserver is compiling a catalog for a windows node, the `file` function will fail. I think we need to conditional logic to the module.

We might be able to detect which mode we're in based on server variables https://www.puppet.com/docs/puppet/7/lang_facts_builtin_variables#server-variables I think they're omitted in `puppet apply`

Rewerson:
So, any news on it? Also getting errors on Windows hosts like:
`Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Could not find any files from C:\Program Files\Puppet Labs\Puppet\VERSION (file: /etc/puppetlabs/code/environments/production/modules/puppet_agent/manifests/init.pp, line: 170, column: 42)`

PR #670, opened 2023-09-22
(PA-5266) Change checksum algorithm to SHA256 

mhashizume:
Prior to this commit, the puppet_agent::prepare::package class used the sha256lite algorithm in the file resource for local installation packages.

In certain circumstances (e.g. when a download is interrupted), this checksum, which only checks the first N bytes, does not provide sufficient information to determine if the file is actually there.

This commit updates the file resource to use the more complete sha256 algorithm to handle cases like this.

### DIFF
--- a/manifests/prepare/package.pp
+++ b/manifests/prepare/package.pp
@@ -27,12 +27,11 @@ class puppet_agent::prepare::package (
   }
 
   file { $local_package_file_path:
-    ensure   => file,
-    owner    => $puppet_agent::params::user,
-    group    => $puppet_agent::params::group,
-    mode     => $mode,
-    source   => $source,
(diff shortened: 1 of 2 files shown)


manifests/prepare/package.pp
```diff
@@ -33,6 +33,6 @@
     mode     => $mode,
     source   => $source,
     require  => File[$puppet_agent::params::local_packages_dir],
-    checksum => sha256lite,
+    checksum => sha256,
```
bastelfreak, line 36: do we need to set the checksum at all? Isn't sha256 the default?

mhashizume: That's a good call, thanks Tim. I'll update the commit to remove the attribute entirely.

PR #663, opened 2023-08-24
(maint) PDK Update 2.7.5

AriaXLi:
This commit updates and applies the 2.7.5 PDK template to prepare for the v1.15.0 release

PR #672, opened 2023-09-27
Match allowed datatypes to yumrepo skip_if_unavailable support

seanmil:
In PR #663 an overly strict datatype was set for the skip_if_unavailable parameter. The yumrepo type supports a variety of string values and boolean values.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -357,9 +357,9 @@ Default value: `$facts['os']['family'] ? { 'windows' => "${facts['env_windows_in
 
 ##### <a name="-puppet_agent--skip_if_unavailable"></a>`skip_if_unavailable`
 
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

mhashizume:
Thanks for this @seanmil . Since this is just a minor correction, could you add "(maint)" to the beginning of your commit summary to get it to pass the static code analysis check?

mhashizume (approved):

mhashizume:
Ah sorry one more thing, could you regenerate REFERENCE.md with your changes? You should be able to run `bundle exec puppet strings generate --format markdown --out REFERENCE.md` then add the updated REFERENCE.md to your commit.

Thanks again!

PR #673, opened 2023-09-28
(PA-5820) Correct parameter types

mhashizume:
Previously, 74e022d erroneously typed the $package_version and $install_options parameters as Optional.

This commit updates those parameters to be required, and gives a more specific type to $install_options.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -14,9 +14,9 @@
 #   are in `file` resources.) Note also that backslashes in double-quoted strings _must_
 #   be escaped and backslashes in single-quoted strings _can_ be escaped.
 class puppet_agent::install (
-  Optional[String]                 $package_version = undef,
-  Optional[[Stdlib::Absolutepath]] $install_dir     = undef,
-  Optional[Array]                  $install_options = undef,
+  String                         $package_version = 'present',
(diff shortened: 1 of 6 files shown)


manifests/install.pp
```diff
@@ -14,9 +14,9 @@
 #   are in `file` resources.) Note also that backslashes in double-quoted strings _must_
 #   be escaped and backslashes in single-quoted strings _can_ be escaped.
 class puppet_agent::install (
-  Optional[String]                 $package_version = undef,
+  String                           $package_version = 'present',
   Optional[[Stdlib::Absolutepath]] $install_dir     = undef,
```
joshcooper, line 18:
```diff
-  Optional[[Stdlib::Absolutepath]] $install_dir     = undef,
+  Optional[Stdlib::Absolutepath]   $install_dir     = undef,
```
There's a subtle bug here.. apparently puppet accepts the double brackets but the resulting type will be nil.

joshcooper (approved):
I verified I could retrieve environment classes and their parameters using: 

```
curl -sf -k --cert cert.pem --key key.pem -X GET -H 'Accept: application/json' 'https://xxx:8140/puppet/v3/environment_classes?environment=production'
```

and all of the resulting class parameters have a "type"

PR #675, opened 2023-10-12
(PA-5013) Add Red Hat 9 (ARM64) to the puppet_agent module task

skyamgarp:
- Specified 'latest' to install el-9-aarch64 from **nightlies**, since official packages not available yet.
- Updated MacOS13 first available version and use the released agent build instead of nightlies.

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -46,7 +46,7 @@ describe 'install task' do
     puppet_7_version = case target_platform
                        when %r{debian-11}
                          '7.9.0'
-                       when %r{el-9}
+                       when %r{el-9-x86_64}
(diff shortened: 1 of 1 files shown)


amitkarsale (approved):
LGTM

skyamgarp:
@joshcooper Can you please review this?

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -69,7 +71,7 @@ def log_output_errors(result)
     # else
     # end
     case target_platform
-    when %r{osx-13}
+    when %r{el-9-aarch64}
```
bastelfreak, line 74: why is this changed / why do you remove osx13 here? That change isn't reflected in the PR title

donoghuc (approved):

PR #677, opened 2023-10-25
(PA-5826) Only read Windows VERSION file during puppet apply

joshcooper:
Previously, puppetserver attempted to load the wrong VERSION file when managing `package_version => auto` and compiling catalogs for Windows nodes as can be seen by the new test:

    Evaluation Error: Error while evaluating a Function Call, Could not find any files from C:\Program Files\Puppet Labs\Puppet\VERSION (file: path/to/init.pp, line: 170, column: 42)

This adds a `puppet_runmode` fact. When running as puppet agent, the run mode will be `agent`. When running as puppet apply, it will be `user`. If it's the latter and we're a Windows host, then look for the VERSION file locally.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -131,7 +131,7 @@ class puppet_agent (
   Optional                       $wait_for_pxp_agent_exit = undef,
   Optional                       $wait_for_puppet_run     = undef,
   Array[Puppet_agent::Config]    $config                  = [],
-  String                         $version_file_path       = $facts['os']['family'] ? { 'windows' => "${facts['env_windows_installdir']}\\VERSION", default => '/opt/puppetlabs/puppet/VERSION' }
+  Stdlib::Absolutepath           $version_file_path       = '/opt/puppetlabs/puppet/VERSION'
 ) inherits puppet_agent::params {
   # The configure class uses $puppet_agent::config to manage settings in
   # puppet.conf, and will always be present. It does not require management of
@@ -167,7 +167,11 @@ class puppet_agent (
     # The AIO package version and Puppet version can, on rare occasion, diverge.
     # This logic checks for the AIO version of the server, since that's what the package manager cares about.
     if $package_version == 'auto' {
-      $master_or_package_version = chomp(file($version_file_path))
+      if $facts['os']['family'] == 'windows' and $facts['puppet_runmode'] == 'user' {
+        $master_or_package_version = chomp(file("${facts['env_windows_installdir']}\\VERSION"))
+      } else {
+        $master_or_package_version = chomp(file($version_file_path))
+      }
     } else {
       $master_or_package_version = $package_version
     }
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -131,7 +131,7 @@
   Optional                       $wait_for_pxp_agent_exit = undef,
   Optional                       $wait_for_puppet_run     = undef,
   Array[Puppet_agent::Config]    $config                  = [],
-  String                         $version_file_path       = $facts['os']['family'] ? { 'windows' => "${facts['env_windows_installdir']}\\VERSION", default => '/opt/puppetlabs/puppet/VERSION' }
+  String                         $version_file_path       = '/opt/puppetlabs/puppet/VERSION'
```
bastelfreak, line 134: should this now be Stdlib::Absolutepath?

bastelfreak (approved):

kenyon:
Should fix #665.

manifests/init.pp
```diff
@@ -167,6 +167,9 @@
     # The AIO package version and Puppet version can, on rare occasion, diverge.
     # This logic checks for the AIO version of the server, since that's what the package manager cares about.
     if $package_version == 'auto' {
+      if $facts['os']['family'] == 'windows' and $facts['puppet_runmode'] == 'user' {
+        $version_file_path = "${facts['env_windows_installdir']}\\VERSION"
```
kenyon, line 171: Isn't this a reassignment of the `$version_file_path` variable, since it's already been assigned a value from the class parameter list?

joshcooper: Good catch, I’ll fix that. Weird that all the tests passed. Or would it only show up when running a specific puppet-lint check?

bastelfreak: puppet-lint does not catch that. you need an rspec-puppet test that hits this if condition and checks for `it should compile`.

joshcooper: ah yeah, my test coverage was lacking. I added a test for apply mode on Windows and it triggered the issue:

```
  1) puppet_agent supported Windows x64 environment with auto when applying is expected to contain File[C:\ProgramData\Puppetlabs]
     Failure/Error: it { is_expected.to contain_file("#{appdata}\\Puppetlabs") }
     
     Puppet::PreformattedError:
       Evaluation Error: Cannot reassign variable '$version_file_path' (file: /home/josh/work/modules/puppetlabs-puppet_agent/spec/fixtures/modules/puppet_agent/manifests/init.pp, line: 171, column: 28) on node xxx
     # /home/josh/.rbenv/versions/3.2.2/lib/ruby/3.2.0/benchmark.rb:311:in `realtime'
     # ./spec/classes/puppet_agent_osfamily_windows_spec.rb:106:in `block (5 levels) in <top (required)>'
```

kenyon (approved):

cthorn42 (approved):

PR #678, opened 2023-10-30
(maint) Remove .project file

mhashizume:
.project files were once used for the long-defunct Geppetto. This commit removes an old .project file in the root of this project.


bastelfreak (approved):

joshcooper (approved):
LGTM. There are some references to `.gitignore` and `.pdkignore` though I assume those are part of the module sync business and can be left as-is

PR #687, opened 2023-12-07
(maint) Unnest module and class names in Ruby tasks

MartyEwings:
Prior to this commit a bug that had previously been resolved in : 

https://github.com/puppetlabs/puppetlabs-puppet_agent/commit/2911a0a4af4255895647e20b6ca138eef93806b8

Had been reintroduced by application of rubocop recommendations.

This commit reverts this change, and also adds in the correct PDK compatible block for the cop in question so this wont happen automatically again

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,8 +37,7 @@ Style/BlockDelimiters:
     be consistent then.
   EnforcedStyle: braces_for_chaining
 Style/ClassAndModuleChildren:
-  Description: Compact style reduces the required amount of indentation.
-  EnforcedStyle: compact
+  Enabled: false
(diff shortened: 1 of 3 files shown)


donoghuc:
I'm confused about this, how are task acceptance tests passing? Also this only un-nests the module definitions in the task helper.

donoghuc:
Ah, it appears acceptance tests cover the shell/powershell tasks not the ruby ones.

donoghuc:
Ok, i see what is going on better now. This is a very annoying cop, thanks for disabling it. Seems like we could use some acceptance level testing for the ruby tasks here...

PR #689, opened 2023-12-11
(PE-37302) Update puppet_agent module for recent AIX changes in PE

amitkarsale:


manifests/osfamily/aix.pp
```diff
@@ -51,11 +51,11 @@
   if $puppet_agent::absolute_source {
     $source = $puppet_agent::absolute_source
   } elsif $puppet_agent::alternate_pe_source {
-    $source = "${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/aix-${aix_ver_number}-power/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.aix${aix_ver_number}.ppc.rpm"
+    $source = "${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/aix-power/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.aix${aix_ver_number}.ppc.rpm"
```
bastelfreak, line 54: Isn't that a breaking change? For which PE or for which AIX releases is that required? It would be helpful if you could make PE-37302 public or move it to a public project.

amitkarsale: Yes, I was moving it back to draft PR.

joshcooper: In order to preserve backwards compatibility, I believe we need to branch on the PE version. If the version is less than whatever the next version will be (either 2023.6.0 or 2021.7.7), then we need to include the `aix-${aix_ver_number}-ppc` in the URL. Otherwise, if it's equal or greater than, then use the generic `aix-ppc` This applies to `$puppet_agent::alternate_pe_source`, `$puppet_agent::source` and the `else` branches.

Side note, it

manifests/osfamily/aix.pp
```diff
@@ -41,21 +43,28 @@
...
       # 6.19.1 is the last puppet6 release that ships AIX 6.1 packages
+
       $aix_ver_number = versioncmp($puppet_agent::prepare::package_version, '6.19.1') ? {
         1       => '7.1',
         default => '6.1'
       }
     }
   }
 
+  if (versioncmp($pe_server_version, '2021.7.7') < 0) {
```
amitkarsale, line 54: @joshcooper, does this condition looks valid for backward compatibility?

manifests/osfamily/aix.pp
```diff
@@ -41,21 +43,28 @@
...
+    $aix_class_name = "aix-${aix_ver_number}-power"
+  } else {
+    $aix_class_name = 'aix-power'
+  }
```
kenyon, line 58: I like to avoid repeating the variable assignment for if/else statements like these (also removed the extra parentheses):

```suggestion
  $aix_class_name = if versioncmp($pe_server_version, '2021.7.7') < 0 {
    "aix-${aix_ver_number}-power"
  } else {
    'aix-power'
  }
```

manifests/osfamily/aix.pp
joshcooper: The `< 0` part looks correct:

```
❯ bx puppet apply -e '$pe_server_version = "2021.7.6"; notice(versioncmp($pe_server_version, "2021.7.7"))'
Notice: Scope(Class[main]): -1

❯ bx puppet apply -e '$pe_server_version = "2021.7.7"; notice(versioncmp($pe_server_version, "2021.7.7"))'
Notice: Scope(Class[main]): 0

❯ bx puppet apply -e '$pe_server_version = "2021.7.8"; notice(versioncmp($pe_server_version, "2021.7.7"))'
Notice: Scope(Class[main]): 1
```

But we have to support both 2021.7.x and 2023.x and we don't want to conflate the two streams. So I think you'll need to do something like:

```
if versioncmp($pe_server_version, '2021.7.7' < 0) || (versioncmp($pe_server_version, '2023.0') >= 0 && versioncmp($pe_server_version, '2023.6')) < 0) {
  # use the old way
} else {
  # use the new way
}
```

The disjoint ranges can be expressed more concisely using puppet's builtin types:

```
❯ bx puppet apply -e '$t = SemVer[SemVerRange(">= 2021.7.7 < 2023.0.0"), SemVerRange(">= 2023.6.0")]; notice(SemVer("2021.7.1") =~ $t)'
Notice: Scope(Class[main]): false

❯ bx puppet apply -e '$t = SemVer[SemVerRange(">= 2021.7.7 < 2023.0.0"), SemVerRange(">= 2023.6.0")]; notice(SemVer("2021.7.7") =~ $t)'
Notice: Scope(Class[main]): true

❯ bx puppet apply -e '$t = SemVer[SemVerRange(">= 2021.7.7 < 2023.0.0"), SemVerRange(">= 2023.6.0")]; notice(SemVer("2021.7.19") =~ $t)'
Notice: Scope(Class[main]): true

❯ bx puppet apply -e '$t = SemVer[SemVerRange(">= 2021.7.7 < 2023.0.0"), SemVerRange(">= 2023.6.0")]; notice(SemVer("2023.0.0") =~ $t)'
Notice: Scope(Class[main]): false

❯ bx puppet apply -e '$t = SemVer[SemVerRange(">= 2021.7.7 < 2023.0.0"), SemVerRange(">= 2023.6.0")]; notice(SemVer("2023.5.0") =~ $t)'
Notice: Scope(Class[main]): false

❯ bx puppet apply -e '$t = SemVer[SemVerRange(">= 2021.7.7 < 2023.0.0"), SemVerRange(">= 2023.6.0")]; notice(SemVer("2023.6.0") =~ $t)'
Notice: Scope(Class[main]): true
```

But I think there's a bug that `SemVer` requires all three components `x.y.z`. So if `pe_server_version` was just `2023.7` then it wouldn't work. Also not sure if `SemVer` handles all of the prerelease, `git describe -s`, etc in the same way as `versioncmp`?

joshcooper (approved):

joshcooper (approved):

PR #695, opened 2024-01-10
Update to PDK template 3.0.1

mhashizume:
This PR updates this module from PDK template 2.7.5 to 3.0.1.

In addition to the boilerplate changes, this PR also:

- Updates the "auto release" GitHub Action to "release prep," reflecting upstream change (puppetlabs/pdk-templates@c3f57b7).
- Removes the github_changelog_generator and concurrent-ruby gems, which were previously needed for the release process but are no longer necessary in newer release processes in the PDK template.
- Disables the puppet_url_without_modules lint check in sync.yml.
- Adds the Vox acceptance Rake task to .sync.yml.


spec/classes/puppet_agent_spec.rb
```diff
@@ -203,7 +203,9 @@ def global_facts(facts, os)
 
         context 'package_version is same as master when set to auto' do
           let(:params) { { package_version: 'auto' } }
-          let(:node_params) { { serverversion: '7.6.5' } }
+          let(:facts) do
+            global_facts(facts, os).merge(serverversion: '7.6.5')
+          end
```
mhashizume, line 208: I was getting test failures after the PDK update relating to this line and variable reassignment (see failed tests here: https://github.com/puppetlabs/puppetlabs-puppet_agent/actions/runs/7481505671/job/20363217135).

I assumed it was because we were treating this as a parameter instead of a fact. I changed it to what you see here and tests are passing again, but I'm not 100% certain that this is the right approach.

joshcooper: Looks to match the behavior below so :+1: to me: 

```ruby
        let(:facts) do
          global_facts(facts, os).merge(is_pe: true)
        end
```

joshcooper (approved):

PR #697, opened 2024-02-10
support download with perl file::fetch

h0tw1r3:
minimal OS installs without wget or curl may not have perl LWP::Simple either, but perl File::Fetch is often available.

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -67,13 +67,19 @@ assert_unmodified_apt_config() {
 }
 
 # Check whether perl and LWP::Simple module are installed
-exists_perl() {
-  if perl -e 'use LWP::Simple;' >/dev/null 2>&1
-  then
+exists_perl_lwp() {
+  if perl -e 'use LWP::Simple;' >/dev/null 2>&1 ; then
     return 0
-  else
-    return 1
   fi
+  return 1
+}
+
+# Check whether perl and File::Fetch module are installed
+exists_perl_ff() {
+  if perl -e 'use File::Fetch;' >/dev/null 2>&1 ; then
+    return 0
+  fi
+  return 1
 }
 
 # Get command line arguments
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

tasks/install_shell.sh
```diff
@@ -434,13 +440,33 @@ do_perl() {
...
+
+# do_perl_ff URL FILENAME
+do_perl_ff() {
+  info "Trying perl (File::Fetch)..."
+  run_cmd "perl -e 'use File::Fetch; use File::Copy; my \$ff = File::Fetch->new(uri => \$ARGV[0]); my \$outfile = \$ff->fetch() or die \$ff->server; copy(\$outfile, \$ARGV[1]) or die \"copy failed: \$!\"; unlink(\$outfile) or die \"delete failed: \$!\";' '$1' '$2' 2>>$tmp_stderr"
+  rc=$?
+
+  # check for 404
+  grep "HTTP response: 404" $tmp_stderr 2>&1 >/dev/null
+  if test $? -eq 0 ; then
```
kenyon, line 459:
```suggestion
  if grep -q "HTTP response: 404" $tmp_stderr ; then
```

tasks/install_shell.sh
```diff
@@ -67,13 +67,19 @@ assert_unmodified_apt_config() {
...
 
 # Check whether perl and LWP::Simple module are installed
-exists_perl() {
-  if perl -e 'use LWP::Simple;' >/dev/null 2>&1
-  then
+exists_perl_lwp() {
+  if perl -e 'use LWP::Simple;' >/dev/null 2>&1 ; then
+    return 0
+  fi
+  return 1
```
kenyon, line 74: These can be simplified:

```suggestion
  perl -e 'use LWP::Simple;' >/dev/null 2>&1
```

tasks/install_shell.sh
h0tw1r3: There are so many things that could be improved in this script.
I generally kept the same format as the existing perl function to make it easy to review and approve.

tasks/install_shell.sh
h0tw1r3: I generally kept the same format as the existing perl function to make it easy to review and approve.

h0tw1r3: The entire script is littered with this type of syntax. Happy to open a separate PR to "write" all the wrongs.

mhashizume (approved):
Thanks @h0tw1r3 !

PR #702, opened 2024-02-22
support download with python3 urllib.request

h0tw1r3:
Found minimal images that do not include perl, curl, or wget, but did have python3.

urllib.request is built-in to python3, reliable way to download for our purposes.


bastelfreak (approved):

mhashizume (approved):
Thanks @h0tw1r3 !

PR #704, opened 2024-02-27
[Fix #703] use 'type' instead of 'which' for increased portability

### LINKED ISSUE #703: `install` task incorrectly identifies Puppet as not installed on target hosts missing the `which` program.
## Describe the Bug

If a target of the `install` task does not have a `which` program available, but *does* have Puppet installed, the `install` task incorrectly attempts to re-install puppet.

## Expected Behavior

If Puppet and its dependencies are installed correctly on a Linux target, the `install` task in this module should not attempt to reinstall them.

## Steps to Reproduce

1. Install Arch Linux in a barebones/default config following the [docs](https://wiki.archlinux.org/title/installation_guide). In particular, install no additional packages beyond what is absolutely necessary.
2. Since Bolt cannot yet automate the installation of Puppet on Arch, install Puppet on Arch manually using the [suggested installation procedure](https://wiki.archlinux.org/title/puppet): `pacman -S puppet`. 
3. Make sure `which` is not installed on the Arch target (which is the default situation of fresh installs): `pacman -R which`.
4. From a system with Bolt installed, attempt to run the `install` task on the Arch target: `bolt task run puppet_agent::install --targets arch --log-level debug`. 
5. Observe that a failure installing puppet occurs despite Puppet already being correctly installed on the target host:
```
{"target":"arch","action":"task","object":"puppet_agent::install","status":"failure","value":{"_output":"20:08:19 +0000 INFO: Version parameter not defined and no agent detected. Assuming latest.\n20:08:19 +0000 INFO: Downloading Puppet latest for Arch...\n20:08:19 +0000 CRIT: Sorry Arch is not supported yet!\n","_error":{"kind":"puppetlabs.tasks/task-error","issue_code":"TASK_ERROR","msg":"The task failed with exit code 1","details":{"exit_code":1}}}}
```

That failure is expected: Arch is not supported by this module *for Puppet installation*, but there's no reason it can't be supported by this module *if Puppet is already correctly installed on the target*.

## Environment
 - Bolt 3.28, running from MacOS 14.
 - Puppet 7.27 on ruby 3
 -  Arch Linux installed according to the default install guide suggestions (kernel 6.7.6 as of this writing).

## Additional Context

The failure is occurring due to this line: https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/tasks/install_shell.sh#L147

`which` is not a shell builtin or a requirement of a POSIX-compliant environment. 

Surprisingly (at least to me), it's also not installed by default as a transitive or part of the core system packages on Arch, so this is less "my environment is weird for personal reasons" and more "a popular distro is ~weird~ missing a dependency of this module".

zbentley:
Fixes #703 by using the `type` shell builtin instead of the (maybe not installed) `which` binary.

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -144,7 +144,7 @@ fi
 # Track to handle puppet5 to puppet6
 if [ -f /opt/puppetlabs/puppet/VERSION ]; then
   installed_version=`cat /opt/puppetlabs/puppet/VERSION`
-elif which puppet >/dev/null 2>&1; then
+elif type -p puppet >/dev/null; then
   installed_version=`puppet --version`
 else
   installed_version=uninstalled


kenyon (approved):

mhashizume:
Zac! Thanks for the PR

mhashizume (approved):

zbentley:
From one Bearcat to another, thanks @mhashizume!

PR #706, opened 2024-03-04
[PA-5586]: Add debian-12-x86_64 to the puppet agent module task acceptance

span786:
Added debian-12 to the puppet agent module task acceptance

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -38,6 +38,19 @@ describe 'install task' do
     puts logger.info(out)
   end
 
+  # Added this method to simplify the 'case' condition
+  # used for target_platform, which will use latest puppet_agent
+  # in below mentioned test spec
+  def latest_platform_list
+    %r{
+      el-9-aarch64|
+      ubuntu-22\.04-aarch64|
+      amazon-2023|
+      osx-14|
+      debian-12
+    }x
+  end
+
   it 'works with version and install tasks' do
     # Specify the first released version for each target platform. When adding a new
     # OS, you'll typically want to specify 'latest' to install from nightlies, since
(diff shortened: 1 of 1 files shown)


task_spec/spec/acceptance/init_spec.rb
```diff
@@ -38,6 +38,19 @@ def log_output_errors(result)
     puts logger.info(out)
   end
 
+  # Added this method to simplify the 'case' condition
+  # used for target_platform, which will use latest puppet_agent
+  # in below mentioned test spec
+  def latest_platform_list
+    %r{
+      el-9-aarch64|
+      ubuntu-22.04-aarch64|
```
kenyon, line 47:
```suggestion
      ubuntu-22\.04-aarch64|
```

kenyon (approved):
Minor thing, since these are regexps, the dot can match any character, which shouldn't cause an unexpected match, but I always like to escape characters that should be literal instead of regexp metacharacters. You could also use Regexp.escape here instead.

task_spec/spec/acceptance/init_spec.rb
span786: Thank you @kenyon. Good catch.

I have modified the regex. Cheers!!

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -38,6 +38,19 @@ def log_output_errors(result)
...
+  # Added this method to simplify the 'case' condition
+  # used for target_platform, which will use latest puppet_agent
+  # in below mentioned test spec
+  def latest_platform_list
+    %r{
+      el-9-aarch64|
+      ubuntu-22\.04-aarch64|
+      amazon-2023|
+      osx-14|
+      debian-12
```
skyamgarp, line 50: Hello Saurabh,
I think we can remove above other platforms as those are released and versions for them can be updated. But we might need a different jira for same.

kenyon (approved):

kenyon (approved):

span786:
@bastelfreak can you please review this PR ?
It is not allowing to merge without code owner approval.

bastelfreak (approved):

PR #711, opened 2024-04-10
sh tasks: Run with set -e

bastelfreak:
This ensures that the task doesn't continue when an error happened. This is also best practice in bash.

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+set -e
+
 # Install puppet-agent as a task
 #
 # From https://github.com/petems/puppet-install-shell/blob/master/install_puppet_5_agent.sh
--- a/tasks/version_shell.sh
+++ b/tasks/version_shell.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
+
 loc=/opt/puppetlabs/puppet/VERSION
 if test -f $loc; then
   echo "{\"version\":\"$(cat $loc)\",\"source\":\"${loc}\"}"


tasks/install_shell.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+set -e
```
bastelfreak, line 3: this should be tested a bit, I don't know if this works out of the box on all operating systems

kenyon (approved):
I pasted these into https://www.shellcheck.net/ and didn't see any findings related to error exits. Lots of other findings in `install_shell.sh` though.

tasks/install_shell.sh
linuxdaemon: `set` is a bash builtin, so assuming the script is executed with Bash (as the shebang requires), then it should work fine.
ref: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

bastelfreak: That wasn't my point. I don't know if the whole code is compatible with `-e`.

smortex: Either the code does proper error checking, either the test suite needs some fixes :shrug:

From my point of view, the best test case is to release it to the public and if this cause trouble, fix the trouble.

smortex (approved):
+1 for good practice!

joshcooper:
I'm running this through our supported platforms to see if anything breaks.

joshcooper (approved):
Tests passed in https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/puppetlabs-puppet_agent%20module/view/ad-hoc/job/forge-module_puppetlabs-puppet-agent-module_intn-sys_pa-acceptance_7-nightly_to_8-nightly-adhoc/46/ except some newer platforms failed to install due to using an older beaker 3.0.x I think it's because this PR was created before we updated our modulesync dependencies.

tasks/version_shell.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
```
corporate-gadfly, line 2: Shebang does not require bash here, so open to interpretation (default `/bin/sh` is `dash` on Ubuntu), @bastelfreak .

PR #712, opened 2024-04-12
Fix strftime() deprecation warning

smortex:
When running on Puppet 8, a warning is produced:
```
The argument signature (String format, [String timezone]) is deprecated for #strftime. See #strftime documentation and Timespan type for more info
```

Pass a timestamp as the first parameter to use a non-deprecated
signature of the function `(Timespan $time_object, String $format)`.


bastelfreak (approved):

mhashizume (approved):

mhashizume:
Thanks @smortex !

PR #713, opened 2024-04-13
install support download with apt-helper

h0tw1r3:
recent ubuntu official docker images are unable to download with any of the other methods

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -66,6 +66,11 @@ assert_unmodified_apt_config() {
   fi
 }
 
+# Check whether apt-helper is available
+exists_apt_helper() {
+  test -x /usr/lib/apt/apt-helper
+}
+
 # Check whether python3 and urllib.request are available
 exists_python3_urllib() {
   python3 -c 'import urllib.request' >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


tasks/install_shell.sh
```diff
@@ -432,6 +437,19 @@ do_fetch() {
   return 0
 }
 
+do_apt_helper() {
+  info "Trying apt-helper..."
+  run_cmd "/usr/lib/apt/apt-helper download-file \"$1\" \"$2\"" 2>$tmp_stderr
```
bastelfreak, line 442: Do we need double quotes? 
```suggestion
  run_cmd "/usr/lib/apt/apt-helper download-file '$1' '$2'" 2>$tmp_stderr
```

h0tw1r3: updated

bastelfreak (approved):

tasks/install_shell.sh
```diff
@@ -525,6 +543,10 @@ do_download() {
     do_python3_urllib $1 $2 && return 0
   fi
 
+  if exists_apt_helper; then
+    do_apt_helper $1 $2 && return 0
+  fi
+
   critical "Cannot download package as none of wget/curl/fetch/perl-LWP-Simple/perl-File-Fetch/python3 is found"
```
kenyon, line 550:
```suggestion
  critical "Cannot download package as none of wget/curl/fetch/perl-LWP-Simple/perl-File-Fetch/python3/apt-helper is found"
```

kenyon (changes requested):

tasks/install_shell.sh
h0tw1r3: updated

kenyon (approved):

tasks/install_shell.sh
```diff
@@ -432,6 +437,19 @@ do_fetch() {
...
+  info "Trying apt-helper..."
+  run_cmd "/usr/lib/apt/apt-helper download-file '$1' '$2'" 2>$tmp_stderr
+  rc=$?
+
+  if test $rc -eq 0 && test -s "$2" ; then
+    return 0
+  fi
+
+  capture_tmp_stderr "apthelper"
+  return 1
```
joshcooper, line 450: All of the previous `do_wget/curl/etc` test for failures like

```
if test $rc -ne 0 || test ! -s "$2"; then
  return 1
fi
```
and then `return 0` at the end. Could you do the same here?

h0tw1r3: done, also added 404 test like other methods

joshcooper:
Closed and reopened to kick tests

PR #714, opened 2024-04-29
puppet_agent::run: properly validate environment parameter

bastelfreak:
previously we assumed that an environment exists in the params hash. And we didn't allow environment names with a single character. This PR fixes both.


tuxmea (approved):

joshcooper (approved):
Ideally would be great to have test coverage for the case where `environment` is absent versus a single character, but the task specs are lacking in general, so I'm ok merging.

PR #717, opened 2024-05-13
Support rpm platforms without yum

h0tw1r3:
observed the latest rockylinux 9 containers do not have a yum symlink to dnf anymore.

Related:
* https://github.com/puppetlabs/provision/pull/261
* https://github.com/puppetlabs/litmusimage/pull/88

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -554,10 +554,11 @@ install_file() {
       fi
 
       rpm -Uvh --oldpackage --replacepkgs "$2"
+      exists dnf && PKGCMD=dnf || PKGCMD=yum
       if test "$version" = 'latest'; then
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):

jordanbreen28 (approved):

donoghuc (commented):
Should the default remain yum in the case both are present?

h0tw1r3:
> Should the default remain yum in the case both are present?

No, technically yum has been deprecated since EL8. Anytime it's presence it should be preferred.

donoghuc (approved):
Looks like PR tests are unrelated. This seems like a good change. thanks

gavindidrichsen (approved):
Looks good to me!

PR #721, opened 2024-06-05
(PA-6276) Update puppet_agent module task acceptance for AL2

amitkarsale:


task_spec/spec/acceptance/init_spec.rb
```diff
@@ -46,6 +46,7 @@ def latest_platform_list
       el-9-aarch64|
       ubuntu-22\.04-aarch64|
       amazon-2023|
+      amazon-2-aarch64|
```
bastelfreak, line 49: does that add support for amazon-2 only on aarch64 or also for AMD? then you could maybe use `amazon-2-`?

tasks/install_shell.sh
```diff
@@ -663,7 +663,7 @@ case $platform in
     filetype="rpm"
     platform_package="el"
     # For Amazon Linux 2023 and onwards we can use the 'amazon' packages created instead of 'el' packages
-    if (( $platform_version >= 2023 )); then
+    if (( $platform_version == 2023 || $platform_version == 7 )); then
```
joshcooper, line 666: Should be this be
`== 2` since that's what is contained in the package name?

joshcooper (approved):

PR #722, opened 2024-06-10
(PA-6277) change manifests for amazon 2 repo

imaqsood:
Since we currently use the Amazon repository for Amazon Linux on the Yum repo instead of `el`, we are changing the manifest to fetch the correct package.


joshcooper (approved):

manifests/osfamily/redhat.pp
```diff
@@ -20,7 +20,6 @@
       'Amazon': {
         $major_version = $facts['os']['release']['major']
         $amz_el_version = "${major_version}" ? {
-          '2'              => '7',
```
joshcooper, line 23: Removing this is correct for amazon-2-aarch64, but breaks amazon-2-x86_64 (since we need to install RHEL 7 packages.

PR #723, opened 2024-06-18
(PA-6416) Update task acceptance for Ubuntu 24.04 amd64/aarch64

shubhamshinde360:


task_spec/spec/acceptance/init_spec.rb
```diff
@@ -45,6 +45,7 @@ def latest_platform_list
     %r{
       el-9-aarch64|
       ubuntu-22\.04-aarch64|
```
bastelfreak, line 47: here you match for aarch64, should that be cleaned up?

skyamgarp: I am not sure what you are asking here but cleanup for released platforms can be done with separate jira.
https://perforce.atlassian.net/browse/PA-6627

joshcooper: The reason we're not including aarch for Ubuntu 24 is because both x86_64 and aarch are newly supported platforms and haven't had public releases yet

joshcooper (approved):

donoghuc:
Did you do an ad-hoc test run to confirm this works?

skyamgarp:
> Did you do an ad-hoc test run to confirm this works?

We tried to but there isn't an option to add ubuntu-24 as test platform. 

On local it was successful

donoghuc (approved):

PR #726, opened 2024-06-26
(PA-6627) Update the first released version for released platforms in…

skyamgarp:
… spec

 We add a new platform in 'latest_platform_list' method so that it considers builds from nightlies with the 'latest' tag before they are released (since the versioned release package won't be available for new platforms). The platforms which are no more new and have been released are removed from 'latest_platform_list' here along with their first released versions specified as needed.

 We also test upgrade from the first released puppet7 version to the latest released puppet7 version. The new platforms only have one release, so this test is not applicable for them. Removed platforms which have had multiple puppet7 versions released as of now to make the test applicable for them. This is done by setting the 'multiple_puppet7_versions' parameter to true.

Tested this on local. Ad-hoc pipeline isn't available at this stage.

Updated release version according to https://perforce.atlassian.net/wiki/spaces/PUPPROD/pages/371851628/Currently+Supported+OS+Platforms

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -42,17 +42,7 @@ describe 'install task' do
   # used for target_platform, which will use latest puppet_agent
   # in below mentioned test spec
   def latest_platform_list
-    %r{
-      el-9-aarch64|
-      ubuntu-22\.04-aarch64|
-      ubuntu-24|
-      amazon-2023|
-      amazon-2|
-      osx-14|
-      debian-12|
-      debian-11-aarch64|
-      el-9-ppc64le
-    }x
+    %r{ubuntu-24}
   end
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
can you elaborate a bit why you do this change?

skyamgarp:
> can you elaborate a bit why you do this change?

We add a new platform in `latest_platform_list` method so it should consider build from nightlies as its not released yet. Now as they are released I have removed them from the method.

shubhamshinde360 (approved):

donoghuc:
That information belongs in the commit message. It is very unclear from the commit message what the refactor is doing.

shubhamshinde360:
@bastelfreak, @donoghuc Could you please take a look, added more info in the commit message. Thanks!

joshcooper (approved):

donoghuc (approved):

PR #732, opened 2024-08-07
support --tags for triggering puppet agent

bastelfreak:


tasks/run.json
```diff
@@ -9,6 +9,10 @@
     "environment": {
       "description": "The desired puppet code environment to use",
       "type": "Optional[String[1]]"
+    },
+    "tags": {
+      "description": "Optional string that will be passed to --tags",
+      "type": "Optional[String[1]]"
```
joshcooper, line 15: Puppet accepts a [comma delimited list of tags](https://github.com/puppetlabs/puppet/blob/1fba2505e6fd5c7ed17257c07667da98d6ed1dfb/lib/puppet/defaults.rb#L2084-L2086), so I'm thinking this should accept an array of non-empty strings

```ruby
"type": "Optional[Array[String[1]]]"
```

And then in the `run.rb` task below, join the tags

```ruby
"--tags=#{params['tags'].join(',')}"

**Comment by bastelfreak** [comment]
@joshcooper sorry for the delay, this is now fixed

File: tasks/run.json [hunk 2]
```diff
@@ -9,6 +9,10 @@
     "environment": {
       "description": "The desired puppet code environment to use",
       "type": "Optional[String[1]]"
+    },
+    "tags": {
+      "description": "Optional string that will be passed to --tags",
+      "type": "Optional[Array[String[1]]]"
```
joshcooper, line 15: Puppet accepts a [comma-separated list of tags](https://www.puppet.com/docs/puppet/8/configuration.html#tags), so wondering if this should this be?

```ruby
Optional[Variant[String[1],Array[String[1]]]
```

plans/run.pp
```diff
@@ -65,7 +67,11 @@
     true    => { 'noop' => true, },
     default => {},
   }
-  $args = $arg_env + $arg_noop + { '_catch_errors' => true }
+  $arg_tags = $tags ? {
+    true    => { 'tags' => $tags },
+    default => {},
+  }
```
smortex, line 73: This would only happen if `$tags == true` (boolean true, not any truthy value, this is not what you want).

```suggestion
  $arg_tags = if $tags {
    { 'tags' => $tags }
  } else {
    {}
  }
```

bastelfreak: on interesting

bastelfreak: to stay consistent with the above logic, I keep the selector but validate the `Undef` datatype now

joshcooper (approved):
macOS 12 failures are unrelated

PR #738, opened 2024-11-20
Remove Litmus from Gemfile

joshcooper:
We do not use Litmus for tests and its dependencies are causing issues in our CI pipeline. This commit removes Litmus from the Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,6 @@ group :development, :release_prep do
   gem "puppetlabs_spec_helper", '~> 7.0', require: false
 end
 group :system_tests do
-  gem "puppet_litmus", '~> 1.0',      require: false, platforms: [:ruby, :x64_mingw]
   gem "CFPropertyList", '< 3.0.7',    require: false, platforms: [:mswin, :mingw, :x64_mingw]
   gem "serverspec", '~> 2.41',        require: false
   gem "voxpupuli-acceptance", '~> 3', require: false


joshcooper:
We couldn't find a way to exclude just a single gem from pdk update, e.g. https://github.com/puppetlabs/puppetlabs-host_core/pull/56#issuecomment-2486726529

bastelfreak:
ah okay, makes sense.

About the CI issues: We fixed some CentOS 9 issues in https://github.com/voxpupuli/beaker-docker/pull/148 , but that broke other jobs. Current idea to fix it is https://github.com/voxpupuli/beaker-docker/pull/151

kenyon:
> Thanks for the pointers @bastelfreak but I think (?) these failures may be due to the apt module?
> 
> ```
>      # --- Caused by: ---
>      # Puppet::ParseError:
>      #   Apt::Setting[list-puppet-enterprise-installer]: parameter 'content' expects a value of type Undef or String[1], got String
> ```
> 
> Specifically this change [puppetlabs/puppetlabs-apt@1e1baad#diff-792ade83ea45b5c6696875678e6e7cc52f4a105db6ed4d8b6f2933ec5d7d6f3bR25-R26](https://github.com/puppetlabs/puppetlabs-apt/commit/1e1baadf8158a878562b8aab1924a3db5b777a0c#diff-792ade83ea45b5c6696875678e6e7cc52f4a105db6ed4d8b6f2933ec5d7d6f3bR25-R26)

Yes, that needs to be fixed at https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/26310e3028a3d54a603dc02c75019bd19c0e8b8a/spec/classes/puppet_agent_osfamily_debian_spec.rb#L61 and other places in that file, similar to this fix: https://github.com/voxpupuli/puppet-rabbitmq/pull/1025/files#diff-01e9ff00dcf670b76d375536642dcfc62cc0efa6d796820d20b87fe721d9ed7e

bastelfreak:
The bug in beaker-docker got fixed (see https://github.com/puppetlabs/puppetlabs-puppet_agent/actions/runs/11938696618/job/33277527843?pr=738 for the failures). We just published 2.5.2

joshcooper:
Superseded by https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/740

PR #752, opened 2025-01-20
(PA-6760) Update task_acceptance tests for MacOS15 ARM

skyamgarp:
For MacOS 15(ARM) - 
With this PR we are trying to only install puppet8. The upgrade part we will cover with another jira.
We don't support puppet7 for MacOS 15(ARM). So we don't need tests to upgrade the agent from puppet7 to puppet 8.

### DIFF
--- a/task_spec/spec/acceptance/init_spec.rb
+++ b/task_spec/spec/acceptance/init_spec.rb
@@ -42,206 +42,215 @@ describe 'install task' do
   # release of puppet-agent includes support for these platforms, they can be removed from this method and added to
   # the logic that determines the puppet_7_version variable below.
   def latest_platform_list
-    # %r{operatingsystem-version-architecture}
+    %r{osx-15-arm64}
   end
 
   it 'works with version and install tasks' do
-    # Specify the first released version for each target platform. When adding a new
-    # OS, you'll typically want to specify 'latest' to install from nightlies, since
-    # official packages won't be released until later. During the N+1 release, you'll
-    # want to change `target_platform` to be the first version (N) that added the OS.
-    puppet_7_version = case target_platform
-                       when %r{debian-11-amd64}
-                         '7.9.0'
-                       when %r{el-9-x86_64}
-                         '7.14.0'
-                       when %r{fedora-36}
-                         '7.19.0'
-                       when %r{osx-11}
-                         '7.7.0'
-                       when %r{osx-12}, %r{ubuntu-22.04-amd64}
-                         '7.18.0'
-                       when %r{osx-13}
-                         '7.26.0'
-                       when %r{el-9-aarch64}, %r{ubuntu-22.04-aarch64}
-                         '7.27.0'
-                       when %r{amazon-2023}, %r{osx-14}, %r{debian-11-aarch64}
-                         '7.28.0'
-                       when %r{debian-12}
-                         '7.29.0'
-                       when %r{el-9-ppc64le}, %r{amazon-2}, %r{fedora-40}
-                         '7.31.0'
-                       when %r{ubuntu-24.04}
-                         '7.32.1'
-                       when latest_platform_list
-                         'latest'
-                       else
-                         '7.18.0'
-                       end
-
-    # Use nightlies for unreleased platforms
     case target_platform
     when latest_platform_list
-      puppet_7_collection = 'puppet7-nightly'
-      puppet_8_collection = 'puppet8-nightly'
-    else
-      puppet_7_collection = 'puppet7'
-      puppet_8_collection = 'puppet8'
-    end
+      # Here we only install puppet-agent 8.x from nightlies as we don't support 7.x
+      # We have to consider tests to upgrade puppet 8.x to latest nightlies in future
 
-    # We can only test puppet 7 -> 7 upgrades if multiple Puppet releases
-    # have supported a given platform. Once a platform has been supported
-    # by multiple Puppet releases, it can be removed from this list.
-    # multiple_puppet7_versions = case target_platform
-    #                             when %r{osx-12-arm}
-    #                               false
-    #                             else
-    #                               true
-    #                             end
-    multiple_puppet7_versions = case target_platform
-                                when %r{el-9-ppc64le|amazon-2|ubuntu-24|fedora-40}
-                                  false
-                                else
-                                  true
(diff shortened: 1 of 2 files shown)


tasks/install_shell.sh
```diff
@@ -137,7 +137,7 @@ if [ -n "$PT_mac_source" ]; then
   mac_source=$PT_mac_source
 else
   if [ "$nightly" = true ]; then
-    mac_source='http://nightlies.puppet.com/downloads'
+    mac_source='https://artifactory.delivery.puppetlabs.net:443/artifactory/internal_nightly__local/downloads'
```
joshcooper, line 140: This is fine for testing, but it will need to use the new public API for downloading artifacts before merging.

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -42,206 +42,236 @@ def log_output_errors(result)
...
+                           '7.32.1'
+                         when latest_platform_list
+                           'latest'
+                         else
+                           '7.18.0'
+                         end
+
+      # Use nightlies for unreleased platforms
+      case target_platform
+      when latest_platform_list
```
joshcooper, line 113: I don't think this case is possible, otherwise we would have matched on line 50

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -42,206 +42,236 @@ def log_output_errors(result)
...
+                         when %r{amazon-2023}, %r{osx-14}, %r{debian-11-aarch64}
+                           '7.28.0'
+                         when %r{debian-12}
+                           '7.29.0'
+                         when %r{el-9-ppc64le}, %r{amazon-2}, %r{fedora-40}
+                           '7.31.0'
+                         when %r{ubuntu-24.04}
+                           '7.32.1'
+                         when latest_platform_list
+                           'latest'
```
joshcooper, line 106: Since `latest_platform_list` includes macOS 15, this seems to imply we should install puppet7 nightly on macOS 15, but we know that's not supported...

I'm wondering if it would be clearer to separate the different use cases as:

1. Happy Path: There are puppet7 and 8 released versions available. Install the earliest puppet7 version and upgrade to puppet 8.
2. New Agent hasn't been released yet (like macOS 15 is now). Install the latest dev version.
3. New Agent recently released (like macOS 15 will be in March). Install the earliest puppet8 version and upgrade to the latest dev version.
4. New Agent with multiple releases (like macOS 15 after 8.12.0 comes out). Install the earliest puppet8 version and upgrade to the latest released version.

All 4 cases should be idempotent, so after the step is finished, if you run `puppet agent -t` again, it shouldn't attempt to upgrade/downgrade.

skyamgarp: This won't be the applicable as this is in the `else` part of the `case` which won't be matched when we are running tests for macos15. I will remove this part to reduce ambiguity.

skyamgarp: This PR targets the 2nd case and 1st is existing behaviour, for 3rd case we have a separate jira in which maybe we can add 4th case as well or have new jira for same.

tasks/install_shell.sh
joshcooper: We talked about this and there's a separate ticket to add support for the new endpoint, so I'll resolve this.

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -42,206 +42,231 @@ def log_output_errors(result)
...
+                           '7.31.0'
+                         when %r{ubuntu-24.04}
+                           '7.32.1'
+                         else
+                           '7.18.0'
+                         end
+
+      # Use nightlies for unreleased platforms
+      # puppet_7_collection = 'puppet7-nightly'
+      # puppet_8_collection = 'puppet8-nightly'
```
joshcooper, line 111: Nit, these 3 lines aren't applicable anymore, could you remove them?

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -42,206 +42,231 @@ def log_output_errors(result)
...
-        expect(res['value']['version']).to eq(puppet_7_version)
+      # We can only test puppet 7 -> 7 upgrades if multiple Puppet releases
+      # have supported a given platform. Once a platform has been supported
+      # by multiple Puppet releases, it can be removed from this list.
+      # multiple_puppet7_versions = case target_platform
+      #                             when %r{osx-12-arm}
+      #                               false
+      #                             else
+      #                               true
+      #                             end
```
joshcooper, line 124: nit, could you remove the commented out block that starts with `multiple_puppet7_versions`

task_spec/spec/acceptance/init_spec.rb
```diff
@@ -42,206 +42,231 @@ def log_output_errors(result)
...
+      multiple_puppet7_versions = case target_platform
+                                  when %r{el-9-ppc64le|amazon-2|ubuntu-24|fedora-40}
```
joshcooper, line 126:
```suggestion
                                  when # %r{operatingsystem-version-architecture}
```

That line was added when those platforms were added in 7.31.0. See https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/752/files#diff-49e04ead14b575f80ad5dc0542c02aaabd03fffe27b8098d8afaec6659378544R101-R102

Since there have been multiple releases since 7.31.0, we should be able to test 7.x to 7.y upgrades on those platforms.

joshcooper:
Changes look good. Could you run adhoc pipeline against your branch on all supported platforms

https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/puppetlabs-puppet_agent%20module/view/ad-hoc/job/forge-module_puppetlabs-puppet-agent-module_init-manual-parameters_adhoc/build?delay=0sec

skyamgarp:
Successful ad-hoc execution: https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/puppetlabs-puppet_agent%20module/view/ad-hoc/job/forge-module_puppetlabs-puppet-agent-module_intn-sys_pa-acceptance_7-nightly_to_8-nightly-adhoc/57/

joshcooper (approved):

PR #757, opened 2025-02-12
Update module to support yum-puppetcore.puppet.com

joshcooper:
This updates the `puppet_agent::install` task so it is possible to install rpm-based packages from https://yum-puppetcore.puppet.com. Includes, amazon, fedora, rocky/alma/redhat and sles. Credentials are required when installing from yum-puppetcore. The username defaults to `forge-key` and the password must be set to your forge API token.

```
❯ export PUPPET_FORGE_TOKEN=...
❯ /opt/puppetlabs/bolt/bin/bolt task run puppet_agent::install version=8.11.0 collection=puppetcore8 version=8.11.0 password=${PUPPET_FORGE_TOKEN} --targets ...
...
16:27:49 -0800 INFO: Downloading https://yum-puppetcore.puppet.com/public/puppet8-release-el-8.noarch.rpm
```

This also updates the `puppet_agent` class so it's possible to manage agent versions over time:

```
class { 'puppet_agent':
  package_version => '8.11.0',
  collection => 'puppetcore8',
  password => Sensitive(...)
}
```
 
And it updates the Dockerfiles used to test install and upgrades. See `docker/README.md` for details.

I haven't changed the default behavior of the module if the collection is unspecified or is one of the existing `puppet7`, `puppet7-nightly`, `puppet8`, etc collections. If no collection is specified, then it continues to use the "unversioned" release package from `yum.puppet.com`. For example, on RHEL 9 https://yum.puppet.com/puppet-release-el-9.noarch.rpm.

Still to do:

- [x] update sles to use `/etc/zypp/credentials.d/`
- [x] fix `puppet-lint` check
- [x] fix rspec tests
- [x] fix acceptance tests


docker/install/dnf/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+  && dnf clean all
+
+# This is also duplicated in the docker/bin/helpers/run-upgrade.sh.
+ENV module_path=/tmp/modules
+WORKDIR "${module_path}/puppet_agent"
+COPY metadata.json ./
+
+# Installing dependencies from source. These versions should be within the range
+# of `dependencies` in metadata.json.
+RUN git clone --depth 1 https://github.com/puppetlabs/puppetlabs-stdlib ../stdlib --branch v9.7.0
```
bastelfreak, line 36: Is it a good idea to hardcode versions here? Or should it maybe checkout the latest tag?

tasks/install_shell.sh
```diff
@@ -119,7 +129,11 @@ else
   if [ "$nightly" = true ]; then
     yum_source='https://artifactory.delivery.puppetlabs.net:443/artifactory/internal_nightly__local/yum'
   else
-    yum_source='http://yum.puppet.com'
+    yum_source='https://yum-puppetcore.puppet.com/public'
```
joshcooper, line 132: Rather than change the default, we could introduce a new `puppetcore8` collection so that it's possible to incrementally move to puppetcore. This would complicated a few things as the old and new repos have different GPG keys and in the future could have diverging release package conventions.

docker/install/dnf/Dockerfile
joshcooper: I'm doing this for each module, let me know if there's a better way? 

```
RUN git clone --tags https://github.com/puppetlabs/puppetlabs-stdlib ../stdlib && \
    $(cd ../stdlib && git checkout $(git describe --tags $(git rev-list --tags --max-count=1)))
```

bastelfreak: I'm by far no git expert, but that looks good to me.

manifests/osfamily/redhat.pp
```diff
@@ -175,6 +175,14 @@
...
         sslclientkey        => $_sslclientkey_path,
         skip_if_unavailable => $puppet_agent::skip_if_unavailable,
+        username            => $puppet_agent::username,
+        password            => $puppet_agent::password,
+      }
+      file { '/etc/yum.repos.d/pc_repo.repo':
+        ensure => file,
+        owner  => 0,
+        group  => 0,
+        mode   => '0600',
```
joshcooper, line 185: Filed https://github.com/puppetlabs/puppetlabs-yumrepo_core/issues/77

mhashizume (approved):

PR #761, opened 2025-02-28
fix: Do not hard code package name in puppet_agent_end_run.rb

fetzerms:
When using this module to install puppet-agent compatible packages with a different name (such as `openvox-agent`) calling `puppet_agent_end_run` with `latest` will fail, as the package name is hard coded.

This patch replaces the hard coded `puppet-agent` with a dynamic lookup for the package name. It also considers that (during migration) the `latest_version` lookup will not work.

How to reproduce the error:
```
    class { '::puppet_agent':
      package_version => 'latest',
      package_name    => 'openvox-agent',
      manage_repo     => false,
      require         => [
        Apt::Source['openvox']
      ],
    }
```
Results in:

```
Error: undefined method `parameters' for nil:NilClass
Info: Unknown failure using insync_values? on type: Puppet_agent_end_run[latest] / property: end_run to compare values [true] and false
Error: /Stage[main]/Puppet_agent::Install/Puppet_agent_end_run[latest]/end_run: change from false to true failed: undefined method `parameters' for nil:NilClass

```


fetzerms:
Not sure how to handle the tests, as they do not seem to ensure that the puppet_agent class is in the catalog.

lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
```diff
@@ -32,8 +32,13 @@ def needs_upgrade?
...
     if desired_version == 'latest'
-      latest_version = @resource.catalog.resource('package', 'puppet-agent').parameters[:ensure].latest
-      desired_version = latest_version.match(%r{^(?:[0-9]:)?(\d+\.\d+(\.\d+)?(?:\.\d+))?}).captures.first
+      # Package name might be different to puppet-agent, hence we need to look it up.
+      package_name = @resource.catalog.resource('class', 'puppet_agent')[:package_name]
+
+      # Latest version might be undefined, e.G. if we're about to install a different named
+      # package  than the currently running one. In that case, we'll leave desired_version empty.
+      latest_version = @resource.catalog.resource('package', package_name).parameters[:ensure].latest
+      desired_version = latest_version.match(%r{^(?:[0-9]:)?(\d+\.\d+(\.\d+)?(?:\.\d+))?}).captures.first unless latest_version.nil?
```
joshcooper, line 41: If `latest_version` is nil, then `desired_version` will be too, and the `versioncmp` function will raise:

```
❯ bundle exec ruby -rpuppet -e 'Puppet::Util::Package.versioncmp(nil, "1.2.3")' 
/home/josh/work/puppet-private/lib/puppet/util/package.rb:12:in `versioncmp': undefined method `scan' for nil:NilClass (NoMethodError)
```

Maybe use empty string instead?

```
desired_version = if latest_version.nil?
                               ""
                             else
                               latest_version.match(%r{^(?:[0-9]:)?(\d+\.\d+(\.\d+)?(?:\.\d+))?}).captures.first
                             end
```

lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
```diff
@@ -32,8 +32,13 @@ def needs_upgrade?
     return false if desired_version == 'present'
 
     if desired_version == 'latest'
-      latest_version = @resource.catalog.resource('package', 'puppet-agent').parameters[:ensure].latest
-      desired_version = latest_version.match(%r{^(?:[0-9]:)?(\d+\.\d+(\.\d+)?(?:\.\d+))?}).captures.first
+      # Package name might be different to puppet-agent, hence we need to look it up.
+      package_name = @resource.catalog.resource('class', 'puppet_agent')[:package_name]
```
joshcooper, line 36: If `package_name` can be nil (not sure that's possible?), then it'd be good to set a default:

```suggestion
      package_name = @resource.catalog.resource('class', 'puppet_agent')[:package_name] || 'puppet-agent'
```

joshcooper:
The test are failing because the test "manually" creates a catalog and adds the `agent_latest_package` to it. You could add the class too, it's a little gross. Note the difference with dashes and underscores.

```diff
       Puppet::Type.type(:package).new(name: 'puppet-agent', ensure: 'latest', provider: :yum)
     end
 
+    let(:puppet_agent_class) do
+      Puppet::Type.type(:component).new(name: 'puppet_agent', package_name: 'puppet-agent')
+    end
+
     before(:each) do
+      catalog.add_resource(puppet_agent_class)
       catalog.add_resource(agent_latest_package)
       resource.catalog = catalog
     end
```

lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
fetzerms: I don't think it can (realistically) be nil - but having a sane default still seems like a  good idea.

lib/puppet/provider/puppet_agent_end_run/puppet_agent_end_run.rb
fetzerms: If latest_version is nil then desired_version will still be 'latest' - But I think it makes sense to be more explicit here.

fetzerms:
> The tests are failing because they "manually" create a catalog and add the `agent_latest_package` to it. You could add the class too, it's a little gross. Note the differences with dashes and underscores.
> 
> ```diff
>        Puppet::Type.type(:package).new(name: 'puppet-agent', ensure: 'latest', provider: :yum)
>      end
>  
> +    let(:puppet_agent_class) do
> +      Puppet::Type.type(:component).new(name: 'puppet_agent', package_name: 'puppet-agent')
> +    end
> +
>      before(:each) do
> +      catalog.add_resource(puppet_agent_class)
>        catalog.add_resource(agent_latest_package)
>        resource.catalog = catalog
>      end
> ```

Thanks a lot! I'll enhance/fix the tests and will include your previous review comments early next week.

PR #762, opened 2025-03-03
Fix breaking changes in non-core repos

joshcooper:
Commit e6f817cba5057870c29b9dd220a048914d0c33a7 introduced a breaking change when downloading the GPG key from non-core repos on SLES.

It also incorrectly attempted to manage credentials for non-core repos for all rpm-based platforms.

This commit restores the GPG location on yum.puppet.com and guards the sed commands to only run when the repo collection contains "core"

This should fix https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/Acceptance%20Suites/view/CI%20Goalie/job/forge-module_puppetlabs-puppet-agent-module_intn-run_task_acceptance-main/591/

### DIFF
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -602,11 +602,13 @@ install_file() {
 
       repo="/etc/yum.repos.d/${collection/core/}-release.repo"
       rpm -Uvh --oldpackage --replacepkgs "$2"
-      if [[ -n $username ]]; then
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):
https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/762/files?diff=split&w=1 makes it way more readable.

mhashizume (approved):

PR #766, opened 2025-03-20
Update module to support puppetcore on Windows

joshcooper:
This is analogous to https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/757 but for windows.

Updates the `puppet_agent::install_task` so it's possible to install msi-based packages from https://artifacts-puppetcore.puppet.com. Credentials are required when installing from puppetcore. The username defaults to `forge-key` and the password must be set to your forge API token.

When using WinRM, you must specify Windows credentials (using `--user` and `--password` arguments) to log into the host and puppetcore credentials to download the MSI (as `username` and `password` task parameters)

```
$ /opt/puppetlabs/bolt/bin/bolt task run puppet_agent::install \
  collection=puppetcore8 \
  version=8.11.0 \
  username=forge-key \
  password=... \
  --targets 'winrm://HOST' \
  --user Administrator \
  --password ...
```

Also updates the `puppet_agent` class so it's possible to manage agent versions over time:

```puppet
class { 'puppet_agent':
  package_version => '8.11.0',
  collection => 'puppetcore8',
  password => Sensitive(...)
}
```

The `puppet_agent::prepare::package` class on Windows normally uses a `file` resource to download the MSI from the `source`. However, due to a bug in puppet, it is not possible to pass credentials as the `userinfo` component of the URL, e.g. `https://forge-id:TOKEN@artifacts-puppetcore.puppet.com` So instead use a powershell script to download the MSI. That code is borrowed from https://github.com/klab-systems/puppet_core_agent, credit to @klab-systems

PR #769, opened 2025-04-01
Add puppetcore macos support

cthorn42:
This is a work in progress, currently based on Josh's WIP: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/766

### DIFF
--- a/manifests/osfamily/darwin.pp
+++ b/manifests/osfamily/darwin.pp
@@ -8,24 +8,36 @@ class puppet_agent::osfamily::darwin {
     $productversion_array = split($facts['os']['macosx']['version']['major'], '[.]')
     $productversion_major = $productversion_array[0]
   }
-
-  if $puppet_agent::absolute_source {
-    $source = $puppet_agent::absolute_source
+  $source = if $puppet_agent::absolute_source {
+    $puppet_agent::absolute_source
   } elsif ($puppet_agent::is_pe and (!$puppet_agent::use_alternate_sources)) {
     $pe_server_version = pe_build_version()
     if $puppet_agent::alternate_pe_source {
-      $source = "${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
     } elsif $puppet_agent::source {
-      $source = "${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
     } else {
-      $source = "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
+    }
+  } elsif $puppet_agent::collection =~ /core/ {
+    if count(split($puppet_agent::prepare::package_version, '\.')) > 3 {
+      "https://artifacts-puppetcore.puppet.com/v1/download?type=native&version=${puppet_agent::prepare::package_version}&os_name=osx&os_version=${productversion_major}&os_arch=${puppet_agent::arch}&dev=true"
+    } else {
+      "https://artifacts-puppetcore.puppet.com/v1/download?type=native&version=${puppet_agent::prepare::package_version}&os_name=osx&os_version=${productversion_major}&os_arch=${puppet_agent::arch}"
     }
   } else {
-    $source = "${puppet_agent::mac_source}/mac/${puppet_agent::collection}/${productversion_major}/${puppet_agent::arch}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+    "${puppet_agent::mac_source}/mac/${puppet_agent::collection}/${productversion_major}/${puppet_agent::arch}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
+  }
+
+  $destination_name = if $puppet_agent::collection =~ /core/ {
+    "${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
+  } else {
+    undef
   }
 
   class { 'puppet_agent::prepare::package':
-    source => $source,
+    source           => $source,
+    destination_name => $destination_name,
   }
 
   contain puppet_agent::prepare::package
--- a/manifests/prepare/package.pp
+++ b/manifests/prepare/package.pp
@@ -56,6 +56,45 @@ class puppet_agent::prepare::package (
       creates => $local_package_file_path,
       require => File[$puppet_agent::params::local_packages_dir],
     }
+  } elsif $puppet_agent::collection =~ /core/ and $facts['os']['family'] =~ /Darwin/ {
+    $download_username = getvar('puppet_agent::username', 'forge-key')
+    $download_password = unwrap(getvar('puppet_agent::password'))
+
+    $response_file = "${local_package_file_path}.response"
+    $netrc_file = "${facts['env_temp_variable']}/.netrc"
+    file { $netrc_file:
+      ensure    => file,
+      content   => "machine artifacts-puppetcore.puppet.com\nlogin ${download_username}\npassword ${download_password}\n",
+      mode      => '0600',
+      show_diff => false,
+    }
+
+    $curl_command = "curl --fail -1 -sL --netrc-file '${netrc_file}' -w '%{http_code}' -o '${local_package_file_path}' '${source}' > '${response_file}'"
+    exec { 'Download Puppet Agent for Darwin':
+      command => $curl_command,
+      creates => $local_package_file_path,
+      path    => ['/usr/bin', '/usr/sbin', '/bin', '/sbin'],
+    }
+
+    exec { 'Remove .netrc file':
+      command => "rm -f '${netrc_file}'",
+      path    => ['/usr/bin', '/bin'],
+      onlyif  => "test -f '${netrc_file}'",
+      require => Exec['Download Puppet Agent for Darwin'],
+    }
+    #
+    # TODO: This is a temporary workaround to get the HTTP response code from the curl command.
+    #       For now just outputting the response is good enough.
+    #       We need to find a way to interspect this value and fail the catalog if the response
+    #       code is not 200, and then logging the output wont be as important.
+    #
+    exec { 'Read HTTP Response Code':
+      command   => "cat '${response_file}'",
+      path      => ['/usr/bin', '/bin'],
+      onlyif    => "test -f '${response_file}'",
+      logoutput => true,
+      require   => Exec['Download Puppet Agent for Darwin'],
+    }
   } else {
     file { $local_package_file_path:
       ensure  => file,
--- a/tasks/install_shell.sh
+++ b/tasks/install_shell.sh
@@ -26,6 +26,14 @@ warn () {
     log "WARN: ${1}"
 }
 
+url_parameters() {
+  if [[ "$version" =~ ^([0-9]+)\.([0-9]+)\.([0-9]+)\.([0-9]+)\.g([a-f0-9]+)$ ]]; then
+    echo "&dev=true"
+  else
+    echo ""
+  fi
+}
+
 critical () {
     log "CRIT: ${1}"
 }
@@ -162,10 +170,18 @@ fi
 if [ -n "$PT_mac_source" ]; then
   mac_source=$PT_mac_source
 else
-  if [ "$nightly" = true ]; then
-    mac_source='http://nightlies.puppet.com/downloads'
-  else
-    mac_source='http://downloads.puppet.com'
+ if [[ "$PT_collection" =~ core ]]; then
+    if [ -z "$password" ]; then
+      echo "A password parameter is required to install with puppetcore"
+      exit 1
+    fi
+  mac_source='https://artifacts-puppetcore.puppet.com/v1/download'
+ else
+    if [ "$nightly" = true ]; then
+      mac_source='http://nightlies.puppet.com/downloads'
+    else
+      mac_source='http://downloads.puppet.com'
+    fi
   fi
 fi
 
@@ -421,13 +437,23 @@ do_wget() {
 # do_curl URL FILENAME
 do_curl() {
   info "Trying curl..."
-  run_cmd "curl -1 -sL -D $tmp_stderr '$1' > '$2'"
+  if [[ -n "$3" && -n "$4" ]]; then
+    run_cmd "curl -1 -sL -u '$3:$4' -D $tmp_stderr '$1' > '$2'"
+  else
+    run_cmd "curl -1 -sL -D $tmp_stderr '$1' > '$2'"
+  fi
   rc=$?
 
   # check for 404
   grep "404 Not Found" $tmp_stderr 2>&1 >/dev/null
   if test $? -eq 0; then
     critical "ERROR 404"
+    unable_to_retrieve_package
+  fi
+
+  grep "401 Unauthorized" $tmp_stderr 2>&1 >/dev/null
+  if test $? -eq 0; then
+    critical "ERROR 401: Unauthorized access"
     unable_to_retrieve_package
   fi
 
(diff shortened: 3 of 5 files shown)


manifests/osfamily/darwin.pp
```diff
@@ -20,12 +20,20 @@
     } else {
       $source = "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
     }
-  } else {
+  } elsif $puppet_agent::collection and $puppet_agent::collection =~ /core/ {
+    if $puppet_agent::prepare::package_version =~ /^\d+\.\d+\.\d+\.\d+\.g([a-f0-9]+)+$/ {
```
cthorn42, line 24: Is there a better way to determine if we need to set `dev=true` parameter?

manifests/prepare/package.pp
```diff
@@ -28,12 +38,79 @@
...
+    #       We need to find a way to interspect this value and fail the catalog if the response
+    #       code is not 200, and then logging the output wont be as important.
+    #
+    exec { 'Read HTTP Response Code':
+      command => "cat '${response_file}'",
+      path    => ['/usr/bin', '/bin'],
+      onlyif  => "test -f '${response_file}'",
+      logoutput => true,
+      require => Exec['Download Puppet Agent for Darwin'],
+    }
```
cthorn42, line 105: This is what the output looks like for setting up the netrc file, reading the output file and then executing the install script:
```
Started on jovial-duck.delivery.puppetlabs.net...
Finished on jovial-duck.delivery.puppetlabs.net:
  Notice: /Stage[main]/Puppet_agent::Prepare::Package/File[/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/.netrc]/ensure: defined content as '{sha256}4f3d0ec5c7fa14041867e7009e7924c13c15c54cc2fbfc52362708a88964bf86'
  Notice: /Stage[main]/Puppet_agent::Prepare::Package/Exec[Remove .netrc file]/returns: executed successfully
  Notice: /Stage[main]/Puppet_agent::Prepare::Package/Exec[Read HTTP Response Code]/returns: 200
  Notice: /Stage[main]/Puppet_agent::Prepare::Package/Exec[Read HTTP Response Code]/returns: executed successfully
  Notice: /Stage[main]/Puppet_agent::Install::Darwin/Exec[osx_install script]/returns: executed successfully
  changed: 4, failed: 0, unchanged: 26 skipped: 0, noop: 0
Finished: apply catalog with 0 failures in 15.54 sec
```
For now while working on this I'm mainly looking to get quick feedback on the curl HTTP response. But this is only a work in progress, and I do plan on spending time to improve this.

manifests/prepare/package.pp
```diff
@@ -28,12 +38,79 @@
...
+
+    $response_file = "${local_package_file_path}.response"
+    $netrc_file = "${facts['env_temp_variable']}/.netrc"
+    file { $netrc_file:
+      ensure  => file,
+      content => "machine artifacts-puppetcore.puppet.com\nlogin ${download_username}\npassword ${download_password}\n",
+      mode    => '0600',
+    }
+
+    $curl_command = "curl -1 -sL --netrc-file '${netrc_file}' -w '%{http_code}' -o '${local_package_file_path}' '${source}' > '${response_file}'"
```
joshcooper, line 80: suggest adding `--fail` so it returns non-zero exit on HTTP error

manifests/prepare/package.pp
```diff
@@ -28,12 +38,79 @@
...
+      creates => $local_package_file_path,
+      path    => ['/usr/bin', '/usr/sbin', '/bin', '/sbin'],
+    }
+
+    exec { 'Remove .netrc file':
+      command => "rm -f '${netrc_file}'",
+      path    => ['/usr/bin', '/bin'],
+      onlyif  => "test -f '${netrc_file}'",
+      require => Exec['Download Puppet Agent for Darwin'],
+    }
```
joshcooper, line 92: may produce a "changed" event each time the agent runs?

manifests/prepare/package.pp
```diff
@@ -5,21 +5,31 @@
 # @param source
 #   The source file for the puppet-agent package. Can use any of the data types
 #   and protocols that the File resource's source attribute can.
+# @param destination_name
```
bastelfreak, line 8: since the whole class is private because of `assert_private()`, you can just use `# @api private` , then you don't need any further puppet strings docs.

manifests/prepare/package.pp
```diff
@@ -5,21 +5,31 @@
...
+  if $destination_name {
+    $package_file_name = $destination_name
+  } else {
+    # In order for the 'basename' function to work correctly we need to change
+    # any \s to /s (even for windows UNC paths) so that it will correctly pull off
+    # the filename. Since this operation is only grabbing the base filename and not
+    # any part of the path this should be safe, since the source will simply remain
+    # what it was before and we can still pull off the filename.
+    $package_file_name = basename(regsubst($source, "\\\\", '/', 'G'))
+  }
```
bastelfreak, line 31: my impression is that the current best practice looks like this:
```suggestion
  $package_file_name = if $destination_name {
    $destination_name
  } else {
    # In order for the 'basename' function to work correctly we need to change
    # any \s to /s (even for windows UNC paths) so that it will correctly pull off
    # the filename. Since this operation is only grabbing the base filename and not
    # any part of the path this should be safe, since the source will simply remain
    # what it was before and we can still pull off the filename.
    basename(regsubst($source, "\\\\", '/', 'G'))
  }
```

manifests/prepare/package.pp
```diff
@@ -28,12 +38,79 @@
...
+    $download_username = getvar('puppet_agent::username', 'forge-key')
+    $download_password = unwrap(getvar('puppet_agent::password'))
+
+    $_download_puppet = windows_native_path("${facts['env_temp_variable']}/download_puppet.ps1")
+    file { $_download_puppet:
+      ensure  => file,
+      content => Sensitive(epp('puppet_agent/download_puppet.ps1.epp')),
+    }
+
+    exec { 'Download Puppet Agent':
```
bastelfreak, line 58: the file resource can also download via http(s).Can you use that instead of an exec?

manifests/prepare/package.pp
```diff
@@ -5,21 +5,31 @@
 # @param source
 #   The source file for the puppet-agent package. Can use any of the data types
 #   and protocols that the File resource's source attribute can.
+# @param destination_name
+#   The destination file name for the puppet-agent package. If no destination
+#   is given, then the basename component of the source will be used as the
+#   destination name.
 class puppet_agent::prepare::package (
   Variant[String, Array] $source,
+  Optional[String] $destination_name = undef
```
bastelfreak, line 14: Are empty strings allowed here, or should be prohibit them?
```suggestion
  Optional[String[1]] $destination_name = undef
```

manifests/prepare/package.pp
bastelfreak: Isn't that the kind of issue https://forge.puppet.com/modules/puppetlabs/transition/readme exists for?

tasks/install_powershell.json
```diff
@@ -42,6 +42,14 @@
...
       "type": "Optional[Integer]",
       "default": 5
+    },
+    "username": {
+      "description": "The username to use when downloading from a source location requiring authentication",
+      "type": "Optional[String]"
+    },
+    "password": {
+      "description": "The password to use when downloading from a source location requiring authentication",
+      "type": "Optional[String]"
```
bastelfreak, line 52: Same here, are empty strings allowed, or should it be updated to `Optional[String[1]]`? And can we use Sensitive as well for the password?

manifests/prepare/package.pp
```diff
@@ -28,12 +38,79 @@
...
+  } elsif $puppet_agent::collection and $puppet_agent::collection =~ /core/ and $facts['os']['family'] =~ /Darwin/ {
+    $download_username = getvar('puppet_agent::username', 'forge-key')
+    $download_password = unwrap(getvar('puppet_agent::password'))
+
+    $response_file = "${local_package_file_path}.response"
+    $netrc_file = "${facts['env_temp_variable']}/.netrc"
+    file { $netrc_file:
+      ensure  => file,
+      content => "machine artifacts-puppetcore.puppet.com\nlogin ${download_username}\npassword ${download_password}\n",
+      mode    => '0600',
```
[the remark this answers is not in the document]
cthorn42: 👍 Good catch.

manifests/prepare/package.pp
joshcooper: I tried that originally and discovered there's a bug in puppet if the URL's userinfo component contain special characters, as they end up being double encoded, e.g. `source => 'https://forge-key:CREDS@artifacts-puppetcore.puppet.com/...'` Hence the powershell route

manifests/prepare/package.pp
joshcooper: I'll fix this in the windows PR

manifests/prepare/package.pp
joshcooper: I'll fix this in the Windows PR

tasks/install_powershell.json
joshcooper: Moved discussion to Windows PR

manifests/prepare/package.pp
joshcooper: Moved to Windows PR

manifests/osfamily/darwin.pp
```diff
@@ -8,7 +8,7 @@
     $productversion_array = split($facts['os']['macosx']['version']['major'], '[.]')
     $productversion_major = $productversion_array[0]
   }
-
+  $destination_name = undef
```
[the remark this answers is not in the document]
cthorn42: The variable needs to get set for the scenario if puppetcore isn't being used. The way I had it set here is a work in progress, I'm about to push up changes that reflect how Josh fixed this with his Windows commit. I think it will be a bit cleaner in what is going on. Thanks for the comment I had forgotten to update this.

manifests/osfamily/darwin.pp
```diff
@@ -20,12 +19,25 @@
     } else {
       $source = "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
     }
+  } elsif $puppet_agent::collection and $puppet_agent::collection =~ /core/ {
```
bastelfreak, line 22:
```suggestion
  } elsif $puppet_agent::collection =~ /core/ {
```

manifests/osfamily/darwin.pp
```diff
@@ -8,7 +8,6 @@
     $productversion_array = split($facts['os']['macosx']['version']['major'], '[.]')
     $productversion_major = $productversion_array[0]
   }
-
   if $puppet_agent::absolute_source {
```
bastelfreak, line 11: I think this whole block becomes a bit more readable and shorter, if you do something like:

```puppet
$source = if $puppet_agent::absolute_source {
...
```

Which enables you to get rid of all the following `$source =` in the logic block.

manifests/prepare/package.pp
```diff
@@ -56,6 +56,45 @@
       creates => $local_package_file_path,
       require => File[$puppet_agent::params::local_packages_dir],
     }
+  } elsif $puppet_agent::collection and $puppet_agent::collection =~ /core/ and $facts['os']['family'] =~ /Darwin/ {
```
bastelfreak, line 59:
```suggestion
  } elsif $puppet_agent::collection =~ /core/ and $facts['os']['family'] =~ /Darwin/ {
```

manifests/osfamily/darwin.pp
```diff
@@ -8,24 +8,36 @@
...
-
-  if $puppet_agent::absolute_source {
-    $source = $puppet_agent::absolute_source
+  $source = if $puppet_agent::absolute_source {
+    $puppet_agent::absolute_source
   } elsif ($puppet_agent::is_pe and (!$puppet_agent::use_alternate_sources)) {
     $pe_server_version = pe_build_version()
     if $puppet_agent::alternate_pe_source {
-      $source = "${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
```
joshcooper, line 16: We lost the inner '$'

```suggestion
"${puppet_agent::alternate_pe_source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
```

manifests/osfamily/darwin.pp
```diff
@@ -8,24 +8,36 @@
...
     } elsif $puppet_agent::source {
-      $source = "${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
```
joshcooper, line 18: Same

```suggestion
"${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
```

manifests/osfamily/darwin.pp
```diff
@@ -8,24 +8,36 @@
...
+    if $puppet_agent::prepare::package_version =~ /^\d+\.\d+\.\d+\.\d+\.g([a-f0-9]+)+$/ {
+      "https://artifacts-puppetcore.puppet.com/v1/download?type=native&version=${puppet_agent::prepare::package_version}&os_name=osx&os_version=${productversion_major}&os_arch=${puppet_agent::arch}&dev=true"
     } else {
-      $source = "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "https://artifacts-puppetcore.puppet.com/v1/download?type=native&version=${puppet_agent::prepare::package_version}&os_name=osx&os_version=${productversion_major}&os_arch=${puppet_agent::arch}"
     }
   } else {
-    $source = "${puppet_agent::mac_source}/mac/${puppet_agent::collection}/${productversion_major}/${puppet_agent::arch}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+    "${puppet_agent::mac_source}/mac/${puppet_agent::collection}/${productversion_major}/${puppet_agent::arch}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
```
joshcooper, line 29: Same here, also indentation is off

```suggestion
"${puppet_agent::mac_source}/mac/${puppet_agent::collection}/${productversion_major}/${puppet_agent::arch}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
```

manifests/osfamily/darwin.pp
```diff
@@ -8,24 +8,36 @@
...
-      $source = "${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
+      "${puppet_agent::source}/packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
+    } else {
+      "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${productversion_major}.dmg"
+    }
+  } elsif $puppet_agent::collection =~ /core/ {
+    if $puppet_agent::prepare::package_version =~ /^\d+\.\d+\.\d+\.\d+\.g([a-f0-9]+)+$/ {
+      "https://artifacts-puppetcore.puppet.com/v1/download?type=native&version=${puppet_agent::prepare::package_version}&os_name=osx&os_version=${productversion_major}&os_arch=${puppet_agent::arch}&dev=true"
     } else {
-      $source = "puppet:///pe_packages/${pe_server_version}/${facts['platform_tag']}/${puppet_agent::package_name}-${puppet_agent::prepare::package_version}-1.osx${$productversion_major}.dmg"
```
joshcooper, line 21: I think we want to keep `puppet:///pe_packages/${pe_server_version}/....` as a valid source.

manifests/osfamily/darwin.pp
bastelfreak: But that's not required, is it? We already have `$` before the `{`

tasks/install_shell.sh
```diff
@@ -810,19 +840,29 @@ case $platform in
...
     if test "$version" = "latest"; then
       filename="puppet-agent-latest.dmg"
     else
       filename="puppet-agent-${version}-1.osx${platform_version}.dmg"
     fi
+    info "Mac platform! Lets get you a DMG...!!"
+    if [[ "$collection" =~ core ]]; then
+      if [ -z "$password" ]; then
+      echo "A password parameter is required to install"
+      exit 1
```
joshcooper, line 856: Since we check for an empty password earlier, is the `if [ -z "$password ]` check needed?

tasks/install_shell.sh
```diff
@@ -431,6 +451,12 @@ do_curl() {
     unable_to_retrieve_package
   fi
 
+  grep "HTTP/2 401" $tmp_stderr 2>&1 >/dev/null
```
joshcooper, line 454: I don't know that we can safely assume the server replied with HTTP 2. Probably better to just check for:

```suggestion
grep "401 Unauthorized" $tmp_stderr 2>&1 >/dev/null
```

PR #775, opened 2025-05-06
bug: mark password as sensitive in puppet_agent::install task

jordanbreen28:
This commit marks `password` as sensitive in the install.json bolt task param metadata.

This prevents password from being shown in plaintext, in the bolt logs.

before:
```
Running task puppet_agent::install with '{"retry":5,"collection":"puppetcore8","version":"latest","password":"1234","_task":"puppet_agent::install"}' on [".."]
```

after:
```
Running task puppet_agent::install with '{"retry":5,"collection":"puppetcore8","version":"latest","password":"Sensitive [value redacted]","_task":"puppet_agent::install"}' on [".."]
```

### DIFF
--- a/tasks/install.json
+++ b/tasks/install.json
@@ -48,6 +48,7 @@
     },
     "password": {
       "description": "The password to use when downloading from a source location requiring authentication",
+      "sensitive": true,
       "type": "Optional[String]"
     }
   },
--- a/tasks/install_powershell.json
+++ b/tasks/install_powershell.json
@@ -49,6 +49,7 @@
     },
     "password": {
       "description": "The password to use when downloading from a source location requiring authentication",
+      "sensitive": true,
       "type": "Optional[Sensitive[String[1]]]"
     }
   },
--- a/tasks/install_shell.json
+++ b/tasks/install_shell.json
@@ -50,6 +50,7 @@
     },
     "password": {
       "description": "The password to use when downloading from a source location requiring authentication",
+      "sensitive": true,
       "type": "Optional[Sensitive[String[1]]]"
     }
   },


tasks/install.json
```diff
@@ -48,6 +48,7 @@
     },
     "password": {
       "description": "The password to use when downloading from a source location requiring authentication",
+      "sensitive": true,
```
joshcooper, line 51: Thanks @jordanbreen28 could you do the same for the windows install_powershell.json?

jordanbreen28: hey josh, so [install_powershell.json](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/tasks/install_powershell.json#L52) & [install_shell.json](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/tasks/install_shell.json#L53) already have password defined as `Optional[Sensitive[String[1]]]`, so wasn't sure if that was necessary. Is that the case? (Forgive me, I am not that versed when it comes to bolt..) 

[install.json](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/tasks/install.json#L51) seemed to be the only place where password was not marked as sensitive using the meta param, or datatype.

jordanbreen28: hey @joshcooper - I updated all three instances of password. Thanks for pointing that out!

jordanbreen28: hey @joshcooper - I updated all three instances of the password param to be sensitive. Thanks for pointing that out!

joshcooper (approved):
Verified both tasks redact the password:

### ssh
```
Running task puppet_agent::install with '{"retry":5,"collection":"puppetcore8","version":"8.12.0","username":"forge-key","password":"Sensitive [value redacted]","_task":"puppet_agent::install"}' on ["shy-forerunner.delivery.puppetlabs.net"]
```

### winrm

```
Running task puppet_agent::install with '{"retry":5,"collection":"puppetcore8","version":"8.11.0.60.g0b032a12d","username":"forge-key","password":"Sensitive [value redacted]","_task":"puppet_agent::install"}' on ["winrm://dramatical-pill.delivery.puppetlabs.net"]
```

PR #781, opened 2025-06-16
(PA-7433) Update task_spec to upgrade from puppet7 to latest puppet8 dev build

AriaXLi:
Previously, the puppet_agent::install task to install or upgrade to puppet8 used puppet8 builds from downloads.puppet.com. Since puppet8 builds are no longer published there, this commit updates the puppet_agent::install task to install or upgrade to puppet8 using the puppet8 dev builds from artifactory instead.

This commit also adds some logger.info statements so it's easier to track what specific version of puppet-agent is being upgraded to or installed.

### DIFF
--- a/.github/workflows/task_acceptance_tests.yaml
+++ b/.github/workflows/task_acceptance_tests.yaml
@@ -22,6 +22,12 @@ jobs:
 
     runs-on: 'ubuntu-latest'
     steps:
+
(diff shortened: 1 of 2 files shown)


AriaXLi:
Closing this in favor of #783 which can access the `TWINGATE` secret

PR #795, opened 2025-07-22
(PA-7608) Update artifactory URLs for puppetcore nightlies

skyamgarp:


joshcooper:
@skyamgarp one git process thing, in cases where you want to revert something and take a different path, it's better to `git revert c3bc4959a1a3246d198c622f2bbb2e6aaf619087` and then add your new changes in a different commit. That way it's clear the old behavior was reverted. If you combine the revert with new code, it's not always clear if everything was reverted or just partially.

Could you add `puppetcore8-nightly` support to the windows task too? See https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/0d52b3ddb5072f945bff3514899934fb9bafc624/tasks/install_powershell.ps1#L119-L127

Also could you add `puppetcore8-nightly` to the task metadata for the bash and powershell tasks
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/0d52b3ddb5072f945bff3514899934fb9bafc624/tasks/install_shell.json#L12
https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/0d52b3ddb5072f945bff3514899934fb9bafc624/tasks/install_powershell.json#L11

It looks like I forgot to add `puppetcore7` and `puppetcore8` to the latter. Could you add those too?

david22swan (approved):

PR #796, opened 2025-07-24
(CAT-2413) Update `install_powershell.ps1` to generate the correct link

david22swan:
As it stands the download link generated is missing the type variable and as such is considered invalid. This is likely a remnant of when the agent was the only product hosted on the site.


gavindidrichsen (approved):
Great fix!

PR #802, opened 2025-07-30
(maint) Update Gemfile with pdk 3.5.0 template

tvpartytonight:


tvpartytonight:
The four pending checks here are for tests that we are no longer running as a part of our updates to the test matrices.

joshcooper (approved):

PR #806, opened 2025-09-29
(PE-42492) Use dnf on fedora >= 41

joshcooper:
On Fedora 41, `yum` is a symlink to dnf5, which doesn't accept -d:

    Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y update puppet-agent-8.15.0.9.g6d2fce25b' returned 2: Unknown argument "-d" for command "dnf5". Add "--help" for more information about the arguments.

Puppet's dnf provider was updated to support dnf5 in PA-7414, see puppetlabs/puppet-private@0d932d327ef, but that doesn't help if the provider is explicitly set to `yum`.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,6 +2,8 @@ fixtures:
   repositories:
     stdlib: https://github.com/puppetlabs/puppetlabs-stdlib.git
     apt: https://github.com/puppetlabs/puppetlabs-apt.git
-    inifile: https://github.com/puppetlabs/puppetlabs-inifile.git
+    inifile:
+      repo: https://github.com/puppetlabs/puppetlabs-inifile.git
+      ref: "v6.2.0"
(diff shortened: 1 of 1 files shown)


manifests/install.pp
```diff
@@ -93,7 +93,11 @@
         # any other type of source means we use a package manager (yum) with no 'source' parameter in the
         # package resource below
         $_package_version = $package_version
-        $_provider = 'yum'
```
bastelfreak, line 96: Do we even have to set the provider? Can't we use undef?

joshcooper: Leaving it unspecified makes sense in general.. but it's been this way for so long, I didn't want to break anything. This is especially true because when doing an agent upgrade from 6.x to 8.x, we run the 6.x version of the provider to perform the upgrade. If we switch to dnf, then we have to be sure dnf in puppet 6.x works when upgrading to 8.x.

cthorn42 (approved):

PR #808, opened 2025-10-10
puppet-lint fixes

pearsondavid:
tidy manifests to comply with puppet-lint


joshcooper (commented):
Hi @pearsondavid, it seems puppet-lint is confused, as the output looks worse, none of the keys are aligned.

PR #834, opened 2026-05-08
Restrict bolt gem to development group

joshcooper:
The layout of rake tasks and Gemfiles in this repo is tightly coupled with CJC. The latter relies on bundle installing from the top-level sometimes to run `beaker-hostgenerator`, and sometimes from the  `acceptance` directory when running beaker tests. Rather than fight CJC, revert the previous change and restrict the bolt dependency to the development group and guarded like GEM_BOLT. Make the Gemfile unmanaged until this can be sorted out in the pdk templates.

This resolves the original problem of GH release prep trying to install patron and failing due to lack of libcurl headers.

I was able to run the release prep workflow using the version in this branch: https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/837


.sync.yml
```diff
@@ -11,6 +11,33 @@
...
+      - gem: beaker-module_install_helper
+      - gem: beaker-puppet_install_helper
+      - gem: nokogiri
+      - gem: 'beaker-task_helper'
+        version: '~> 1.9'
+        condition: 'ENV["GEM_BOLT"]'
+    ":system_tests":
+      - gem: voxpupuli-acceptance
+        version: '~> 3'
+
```
joshcooper, line 40: These are strictly not needed (since the Gemfile is unmanaged), but I left them for when we re-manage it.

tvpartytonight (approved):

PR #842, opened 2026-07-01
(PA-8924) Remove macOS 11, 12, and 13 support

mhashizume:
## Summary

Removes macOS 11, 12, and 13 support from the `puppet_agent` module as part of platform end-of-life cleanup.

Jira: [PA-8924](https://perforce.atlassian.net/browse/PA-8924) (epic [PA-8920](https://perforce.atlassian.net/browse/PA-8920), "Remove macOS 13 FOSS support"). Scope was extended to also drop macOS 11 and 12.

## Changes

- **`tasks/install_shell.sh`** — remove the `"11"`, `"12"`, `"13"` cases from the Darwin `major_version` dispatch. These now fall through to the existing `*)` "No builds for platform" branch and exit non-zero, matching their removal from the supported set. macOS 14/15/26 remain.
- **`spec/classes/puppet_agent_osfamily_darwin_spec.rb`** — retarget the supported-macOS catalog-compile matrix (`osx-11`/`osx-12` → `osx-14`/`osx-15`) and the two productversion-parsing tests (macOS 11 → 14) to supported versions. The `10.13` tests exercise a separate `/^10\./` branch in `manifests/osfamily/darwin.pp` and are intentionally unchanged.

`metadata.json` lists `OSX` with no per-release enumeration, and `manifests/osfamily/darwin.pp` derives the macOS major version dynamically, so no changes are required there.

## Testing

- `bundle exec rake spec_standalone` — 1134 examples, 0 failures
- `bundle exec rake lint` (puppet-lint) — clean
- `bundle exec rubocop` — clean

### DIFF
--- a/spec/classes/puppet_agent_osfamily_darwin_spec.rb
+++ b/spec/classes/puppet_agent_osfamily_darwin_spec.rb
@@ -37,7 +37,7 @@ describe 'puppet_agent' do
     let(:params) { { package_version: } }
 
     context 'when running a supported macOS' do
-      ['osx-11-x86_64', 'osx-12-x86_64'].each do |tag|
+      ['osx-14-x86_64', 'osx-15-x86_64'].each do |tag|
         context "on #{tag} with no aio_version" do
           let(:osmajor) { tag.split('-')[1] }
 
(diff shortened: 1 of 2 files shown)


spec/classes/puppet_agent_osfamily_darwin_spec.rb
```diff
@@ -37,7 +37,7 @@
     let(:params) { { package_version: } }
 
     context 'when running a supported macOS' do
-      ['osx-11-x86_64', 'osx-12-x86_64'].each do |tag|
+      ['osx-14-x86_64', 'osx-15-x86_64'].each do |tag|
```
bastelfreak, line 40: oh we really need more MacOS factsets in FacterDB

bastelfreak (approved):

spec/classes/puppet_agent_osfamily_darwin_spec.rb
mhashizume: Agreed — broader macOS FacterDB coverage would let us assert against real factsets here rather than hand-built ones. That's out of scope for this platform-removal change, but a good follow-up for the test fixtures. Thanks for the review!

joshcooper (approved):

PR #838, opened 2026-05-22
(PE-43839) Add acceptance test for puppet8 to puppet9 agent upgrade

span786:
## Summary

- Adds `acceptance/tests/test_upgrade_puppet8_to_puppet9.rb`, mirroring the existing 7→8 test, with adjustments for the puppet9-nightly track (pre-release `8.99.99.x.gSHA` versions, dev_builds routing for newer agent platforms like macOS 26 Intel that have no build at `downloads.puppet.com/mac/puppet8/<os>`).
- `acceptance/Rakefile`: new `UPGRADE_TARGET_COLLECTION` env-var dispatch so the test selected is decoupled from `MASTER_COLLECTION`. Needed because puppetserver-9 nightlies aren't yet published, so the master has to stay on puppet8-nightly while still exercising the 8→9 upgrade.
- `acceptance/helpers.rb`: fix dormant bug in `collection_cmp` — `Symbol#casecmp(String)` returns `nil` on Ruby 3.2, crashing the `min:`/`max:` form of `require_master_collection`. The 7→8 test only uses the `String` form so this code path was never exercised.

Related: [PE-43839](https://perforce.atlassian.net/browse/PE-43839)

## Test plan

- [x] Validated end-to-end via `rake ci` against ABS-provisioned puppet8-nightly master + macOS 26 Intel agent — `Passed: 1, Failed: 0, Errored: 0`
- [x] Confirmed `acceptance/tests/test_upgrade_puppet7_to_puppet8.rb` still works (Rakefile dispatch defaults to `MASTER_COLLECTION` when `UPGRADE_TARGET_COLLECTION` is unset, preserving existing behaviour)
- [x] `ruby -c` syntax-check on the new test file
- [ ] Reviewer: when puppetserver-9 nightlies land at `internal_nightly__local/yum`, drop `UPGRADE_TARGET_COLLECTION` and switch `MASTER_COLLECTION=puppet9-nightly` to test the full puppet9 master path

### How to run

```
cd acceptance
ssh-add ~/.ssh/id_rsa_jenkins   # one-time per session
MASTER_COLLECTION=puppet8-nightly \
UPGRADE_TARGET_COLLECTION=puppet9-nightly \
HOSTS=hosts.yaml \
  bundle exec rake ci
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #844, opened 2026-08-03
(PA-9081) Guard apt-signing sed for ubuntu-24.04/26.04 when no repo file exists

SugatD:
## Summary

- Fixes a failure in the puppet8→9 PA acceptance test (added in #838), specific to Ubuntu 26.04.
- The initial puppet8 agent install uses an explicit `puppet_agent_version`, which routes through `dev_builds` and installs the `.deb` directly rather than via an apt repo. On `ubuntu-26.04` (no published puppet8 apt repo yet), this leaves no `/etc/apt/sources.list.d/*puppet*.list` file, so the unconditional apt-signing `sed` in `acceptance/helpers.rb` failed:
  ```
  sed: can't read /etc/apt/sources.list.d/*puppet*.list: No such file or directory
  ```
- Confirmed via the puppet8→9 experimental Jenkins pipeline (puppetlabs/ci-job-configs#10694, run against this repo's `main` at `61cbd149`) and reproduced/fixed locally with a full end-to-end Beaker run (puppet 8.20.0.34 → puppet9 8.99.99.x, idempotent re-run, `Passed: 1, Errored: 0`).
- **Scoped to `ubuntu-26.04` only** — every other Ubuntu release (18.04/20.04/22.04/24.04) and Debian keep their current unconditional `sed` behavior unchanged.

Related: puppetlabs/ci-job-configs#10694, puppetlabs/ci-job-configs#10727 (enables `ubuntu2604-64` in CI once this lands)

Opened as draft while we re-verify against the ci-job-configs experimental pipeline.

## Test plan

- [x] Ran the full `acceptance/tests/test_upgrade_puppet8_to_puppet9.rb` suite locally against a real Ubuntu 26.04 amd64 agent + RHEL8 master, with this fix applied — passed end-to-end
- [x] Re-run the ci-job-configs experimental pipeline against this branch to confirm in CI

### DIFF
--- a/acceptance/helpers.rb
+++ b/acceptance/helpers.rb
@@ -229,18 +229,14 @@ module Beaker::DSL
           # This discrepancy causes apt to error, so we manually add signing info.
           if %r{debian|ubuntu}.match?(host['platform'])
             step '(Agent) Add apt signing information' do
-              sed_command = "sed -e 's/^deb http/deb [signed-by=\\/etc\\/apt\\/keyrings\\/GPG-KEY-puppet.asc] http/' /etc/apt/sources.list.d/*puppet*.list -i"
-              if host['platform'].include?('ubuntu-24.04') || host['platform'].include?('ubuntu-26.04')
-                # Confirmed via the puppet8-to-puppet9 experimental Jenkins pipeline that
-                # ubuntu-24.04 and ubuntu-26.04 install the initial puppet8 agent via a
-                # direct dev_builds .deb download rather than an apt repo (no sources.list.d
-                # file to sign). ubuntu-22.04 is confirmed passing with the unconditional
-                # sed, so it (and debian) are left on that path.
-                result = on(host, 'ls /etc/apt/sources.list.d/*puppet*.list', acceptable_exit_codes: [0, 2])
-                on(host, sed_command) if result.exit_code.zero?
-              else
-                on(host, sed_command)
-              end
+              # On some platform/release combos beaker-puppet's dev_builds install path
+              # dpkg-installs the artifact directly rather than through an apt repo, leaving
+              # no *puppet*.list file to sign. Skip signing when no such file exists instead
+              # of hardcoding the affected platform/version strings.
+              on(host, 'for f in /etc/apt/sources.list.d/*puppet*.list; do ' \
+                       '[ -e "$f" ] || continue; ' \
+                       "sed -e 's/^deb http/deb [signed-by=\\/etc\\/apt\\/keyrings\\/GPG-KEY-puppet.asc] http/' -i \"$f\"; " \
+                       'done')
             end
           end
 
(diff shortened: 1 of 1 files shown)


SugatD:
CI passed for ubuntu 26 and 24 (aarch64 and 64)
https://jenkins-platform.delivery.puppetlabs.net/view/__experimental%20automatic/job/experimental_auto_puppetlabs-puppet-agent-module_intn-sys_pa-acceptance_8-nightly_to_9-nightly-adhoc/2/

acceptance/helpers.rb
```diff
@@ -229,7 +229,18 @@ def set_up_initial_agent_on(host, options)
...
+                # Confirmed via the puppet8-to-puppet9 experimental Jenkins pipeline that
+                # ubuntu-24.04 and ubuntu-26.04 install the initial puppet8 agent via a
+                # direct dev_builds .deb download rather than an apt repo (no sources.list.d
+                # file to sign). ubuntu-22.04 is confirmed passing with the unconditional
+                # sed, so it (and debian) are left on that path.
+                result = on(host, 'ls /etc/apt/sources.list.d/*puppet*.list', acceptable_exit_codes: [0, 2])
+                on(host, sed_command) if result.exit_code.zero?
+              else
+                on(host, sed_command)
+              end
```
span786: Consider generalizing this instead of hardcoding the affected platform/version strings.

The underlying condition isn't really "is this ubuntu-24.04 or ubuntu-26.04" — it's "did beaker-puppet's `dev_builds` install path skip apt repo configuration and dpkg-install the artifact directly," which happens whenever there's no established public repo for that platform/release (verified this independently against `puppet5.rb#install_from_build_data_url` in beaker-puppet, and reproduced the exact failure live on an ubuntu-26.04 host — no `*puppet*.list` file is ever created there).

Hardcoding `'ubuntu-24.04'` / `'ubuntu-26.04'` means the next release that hits this path (e.g. a future Ubuntu LTS, or Debian once it loses its public repo for a given collection) reproduces this exact CI failure again and needs another platform-string patch here. An unconditional existence check avoids the whack-a-mole and works for every platform/version, present and future, without needing a separate `ls` probe call:

```suggestion
              on(host, 'for f in /etc/apt/sources.list.d/*puppet*.list; do ' \
                       '[ -e "$f" ] || continue; ' \
                       "sed -e 's/^deb http/deb [signed-by=\\/etc\\/apt\\/keyrings\\/GPG-KEY-puppet.asc] http/' -i \"$f\"; " \
                       'done')
```

This drops the platform-version branching and the extra `ls`/`acceptable_exit_codes` round-trip in favor of a single command: the loop is a no-op when no file matches (as on ubuntu-24.04/26.04 today) and still rewrites the file correctly when one exists (every other current debian/ubuntu platform). I verified both branches directly on a live ubuntu-26.04 host while debugging this same failure.

span786 (approved):

PR #847, opened 2026-08-06
(PA-8998) Require a puppet9 master for the puppet8->9 upgrade test

span786:
## Summary

- Resolves [PA-8998](https://perforce.atlassian.net/browse/PA-8998).
- `test_upgrade_puppet8_to_puppet9.rb` used `require_master_collection min: 'puppet8'` because puppet9-nightly puppetserver builds weren't reliably published yet, so the test accepted a puppet8-or-newer master while agents were upgraded to puppet9.
- Per [Josh Cooper's comment on PA-8998](https://perforce.atlassian.net/browse/PA-8998?focusedCommentId=4179028), puppetserver 9-nightly builds are now available. Verified directly against artifactory: `puppetserver9-nightly` is published for `el/9/x86_64` (our master's platform).
- Tightens the guard to `require_master_collection min: 'puppet9'` so the test properly exercises the upgrade against a real puppet9 master instead of silently falling back to puppet8.
- Also updates the Rakefile's `UPGRADE_TARGET_COLLECTION` comment, which cited the same now-stale "puppetserver-9 nightlies aren't yet published" rationale.
- Note (per the same comment): puppetserver 9-nightly platform coverage differs from 8.x (some platforms dropped, others added) — flagged in the updated code comment so this is checked again if the master platform in `hosts.yaml` ever changes.

## Test plan

- [x] Confirmed `puppetserver9-nightly` exists for el-9-x86_64 (current master platform) on artifactory
- [x] `ruby -c` syntax check
- [ ] Full `test_upgrade_puppet8_to_puppet9.rb` acceptance run with `MASTER_COLLECTION=puppet9-nightly`

PR #851, opened 2026-08-18
(PA-9119) Diagnose dnf5 metadata staleness and fix versioncmp() truthiness bug

span786:
## Problem

`acceptance/tests/test_upgrade_puppet8_to_puppet9.rb` failed deterministically on Fedora 41: `Package[puppet-agent]` (`ensure => 'latest'`) found no update immediately after `Yumrepo['pc_repo']` created the repo pointing at the newer puppet9-nightly build, in the same catalog run.

```
Notice: .../Yumrepo[pc_repo]/ensure: created
...
Debug: Package[puppet-agent](provider=dnf): Yum didn't find updates, current version (8.21.0.13...) is the latest
...
Minitest::Assertion: Expected nil to be truthy.
```

Confirmed via a live failing build: [Jenkins build #1, PLATFORM=fedora41-64default](https://jenkins-platform.delivery.puppetlabs.net/view/__experimental%20automatic/job/experimental_auto_puppetlabs-puppet-agent-module_intn-sys_pa-acceptance_8-nightly_to_9-nightly-main/1/PLATFORM=fedora41-64default.a-redhat9-64mdca,WORKER_LABEL=k8s-beaker/consoleFull).

Tracked in [PA-9119](https://perforce.atlassian.net/browse/PA-9119) and [PUP-12151](https://perforce.atlassian.net/browse/PUP-12151).

## Root cause

A dnf5 metadata-cache staleness bug: dnf5's on-disk repo metadata cache can lag a repo that was created or changed earlier in the same Puppet transaction. PA-9119's investigation deterministically reproduced this (100% correlation across 6 trials tied to cache-warm-before-repo-exists), ruled out publish-lag, and pinned it to Puppet's `dnf`/`yum` provider never forcing a metadata refresh when checking for updates.

**The actual fix lives in `puppet-private`**: puppetlabs/puppet-private#217 (adds a dnf5-gated `--refresh` to `check_updates`/`available_versions`). This PR is the module-side half: better failure diagnostics for this class of bug, plus an unrelated bug found while investigating.

## What's in this PR

1. **Acceptance test diagnostics** (`test_upgrade_puppet8_to_puppet9.rb`): on a post-upgrade version-assertion failure, log the repo/package state (the real `--debug` decision lines, `dnf repolist`, `dnf list --showduplicates`) before teardown destroys the evidence, so a recurrence is diagnosable from CI logs alone. Guarded so it only runs on RPM-family agents (checked with a non-raising `test -d /etc/yum.repos.d`, not `facter`/`fact_on` — this branch exists specifically for the case where the agent's facter is broken, and `fact_on` has no `accept_all_exit_codes` equivalent, so it would raise in exactly that scenario). Also fixed the pre-existing `nil.match?` crash risk: `puppet_agent_version_on` returns `nil` (not a non-matching string) when `facter aio_agent_version` itself fails, which would otherwise raise before the diagnostics or even the `assert` could run.

2. **`manifests/install.pp` bug fix**: found during PA-9119's investigation —
   ```puppet
   if $facts['os']['name'] == 'Fedora' and versioncmp($facts['os']['release']['major'], '41') {
   ```
   `versioncmp()` returns an Integer, and Puppet treats every Integer (including `0` and negatives) as truthy, so this was unconditionally true for every Fedora release, not just `>= 41`. Fixed to `versioncmp(...) >= 0`. This restores the pre-existing behavior for Fedora < 41 (explicit `'yum'` provider) that a recent change (`d058ce3`, introducing dnf5 handling for Fedora >= 41) unintentionally dropped for all Fedora releases. Added a regression test (`spec/classes/puppet_agent_spec.rb`) confirmed to fail against the buggy condition.

## Testing performed

- `bundle exec rake spec` — 1135 examples, 0 failures
- `bundle exec rake lint` / `rake syntax` / `rake metadata_lint` — clean
- `bundle exec rubocop` on both modified Ruby files — clean
- Manually verified the new `install.pp` regression test fails against the pre-fix `if versioncmp(...)` condition (reverted, confirmed red, restored)

## Related

- puppetlabs/puppet-private#217 (the actual provider fix this test diagnoses)
- [PA-9119](https://perforce.atlassian.net/browse/PA-9119)
- [PUP-12151](https://perforce.atlassian.net/browse/PUP-12151)

Base branch note: stacked on #847 (PA-8998, already open) since this test file's current state depends on it; the diff will narrow to just this PR's changes once #847 merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/acceptance/tests/test_upgrade_puppet8_to_puppet9.rb
+++ b/acceptance/tests/test_upgrade_puppet8_to_puppet9.rb
@@ -120,7 +120,47 @@ node default {
       # Pre-release puppet9 nightlies report as 8.99.99.<build>.g<sha> (Puppet's
       # next-major pre-release convention); accept those alongside the eventual
       # stable 9.x.y form.
-      assert(puppet_agent_version_on(agent) =~ %r{^(?:9\.\d+\.\d+|8\.99\.99\.\d+).*})
+      installed_version = puppet_agent_version_on(agent)
+      # puppet_agent_version_on returns nil (rather than a non-matching
+      # string) when `facter aio_agent_version` itself fails to run -- e.g.
+      # exactly the "package removed, reinstall failed" failure mode these
+      # diagnostics exist for -- so guard before calling String methods on it.
+      is_upgraded = !installed_version.nil? && installed_version.match?(%r{^(?:9\.\d+\.\d+|8\.99\.99\.\d+).*})
+      unless is_upgraded
+        # Checked with a plain, non-raising shell test rather than
+        # fact_on/facter: this branch exists specifically for the case where
+        # the agent's Ruby/facter is broken or gone, and fact_on has no
+        # accept_all_exit_codes equivalent, so it would raise in exactly
+        # that scenario instead of running these diagnostics. Presence of
+        # /etc/yum.repos.d is also the precondition for the pc_repo.repo read
+        # a few lines below, so it doubles as that guard too.
+        is_rpm_family = on(agent, 'test -d /etc/yum.repos.d', accept_all_exit_codes: true).exit_code.zero?
+        if is_rpm_family
+          # Capture repo/package state before the test's teardown removes
+          # pc_repo and (attempts to) purge the package, which would
+          # otherwise destroy this evidence. RPM-family only: dnf/yum aren't
+          # meaningful on other platforms, and running them there just adds
+          # "command not found" noise on top of the real failure.
+          logger.error("=== repo/package diagnostics on #{agent} (installed version: #{installed_version.inspect}) ===")
+          # The dnf/yum provider's own reasoning about Package[puppet-agent]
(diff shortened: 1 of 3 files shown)


joshcooper:
@span786 I assume this is just for testing and can be closed?

joshcooper (approved):

joshcooper (approved):

joshcooper:
@span786 I merged the wrong PR initially, sorry about that. Still getting used to GH stacks

span786:
No worries! Opened a revert PR for the acceptance-test diagnostics added here (not needed now that the actual root cause is fixed in [puppet-private#220](https://github.com/puppetlabs/puppet-private/pull/220)) — the unrelated `install.pp` `versioncmp()` fix from this PR is left in place: #853

PR #654, opened 2023-05-10
(PA-5337) PE integration

mhashizume:
This commit adds mappings for Puppet Enterprise 2023.3 to puppet8.

PR #854, opened 2026-09-15
(PA-9340) Accept facter-cased "AlmaLinux" in install_shell.sh

imaqsood:
## Summary

`puppet_agent::install` currently fails on every AlmaLinux target with:

```
INFO: Downloading Puppet latest for AlmaLinux...
CRIT: Sorry AlmaLinux is not supported yet!
```

AlmaLinux *is* supported — this is a platform-string casing mismatch. Note the capital **L** in the error: that is the value the task received.

## Root cause

`tasks/install_shell.sh` maps AlmaLinux onto the `el` platform by comparing the platform string returned by the puppetlabs-facts `bash.sh` task against the literal `Almalinux`:

```sh
  # Handle AlmaLinux
  elif test "x$platform" = "xAlmalinux"; then
    platform="el"
```

puppetlabs-facts [PR #71](https://github.com/puppetlabs/puppetlabs-facts/pull/71) (commit [`2455367`](https://github.com/puppetlabs/puppetlabs-facts/commit/24553670), merged to `main` 2026-09-10, released as facts v1.8.0 on 2026-09-11) added an explicit mapping to `munge_name()` so `bash.sh` agrees with facter's `os.name`:

```sh
almalinux) echo "AlmaLinux" ;;
```

Before that, `munge_name()`'s generic fallback only uppercased the first character and produced `Almalinux`, which is what this task was written against.

`test`/`[` string comparison is case sensitive, and `install_shell.sh` does not set `shopt -s nocasematch`, so with facts >= 1.8.0 no branch matches, `$platform` is never mapped to `el`, and execution falls through to:

```sh
  *) critical "Sorry $platform is not supported yet!"
```

facts is the correct side here (facter reports `os.name => AlmaLinux`), so the fix belongs in this module. This change accepts both spellings so the task keeps working against older facts releases too.

`tasks/install_shell.sh` is the only place affected — the manifests, `metadata.json` and `spec/classes/puppet_agent_osfamily_redhat_spec.rb` already use `AlmaLinux`.

## Impact

Fleet-wide across the module repos, since each module's `.fixtures.yml` clones both `facts` and `puppet_agent` from GitHub `main`: all AlmaLinux acceptance jobs went red the day the facts commit landed. Confirmed identical failures in:

- [puppetlabs-haproxy #654](https://github.com/puppetlabs/puppetlabs-haproxy/actions/runs/34966616108/job/104372593934?pr=654) (AlmaLinux-8, puppetcore9-nightly)
- [puppetlabs-concat](https://github.com/puppetlabs/puppetlabs-concat/actions/runs/34942963093) (AlmaLinux-8, puppetcore8- and puppetcore9-nightly)
- [puppetlabs-stdlib](https://github.com/puppetlabs/puppetlabs-stdlib/actions/runs/34913235250) (nightly)

It also affects any Bolt/PE user running `puppet_agent::install` against an AlmaLinux node where facts >= 1.8.0 supplies the platform and facter is not present on the target.

## Verification

Ran the task in `almalinux:8` and `almalinux:9` containers with facts `main`'s `bash.sh` placed under `$PT__installdir/facts/tasks/` and `PT__noop=true`:

| case | before | after |
| --- | --- | --- |
| `almalinux:8` | `CRIT: Sorry AlmaLinux is not supported yet!`, exit 1 | exit 0, `platform=el`, `download_url=http://yum.puppet.com/puppet8-release-el-8.noarch.rpm` |
| `almalinux:9` | same `CRIT` | exit 0 |
| stub `bash.sh` returning the old `Almalinux` spelling | exit 0 | exit 0 (back-compat preserved) |

`bash -n tasks/install_shell.sh` is clean. The beaker task-acceptance suite was not run locally (needs VPN/internal Artifactory; its matrix is `ubuntu-20.04`/`rocky-8`, so it does not cover this path).

Jira: PA-9340


span786 (approved):

amitkarsale (approved):

joshcooper (changes requested):
The change in behavior in the facts module should be reverted. The change to upper case L should be done in a new major version of the facts module

